PR #48, opened 2025-08-26
Migrate to System.CommandLine 2.0.0-beta7

jongio:
Implement migration plan for System.CommandLine 2.0.0-beta5, updating references and adjusting option definitions accordingly.


core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -69,26 +70,28 @@ public virtual ValidationResult Validate(CommandResult commandResult, CommandRes
     {
         var result = new ValidationResult { IsValid = true };
 
+        // Determine any missing required options using beta5 APIs
```
jeffhandley: Nit: The "beta5" reference in the comment will immediately be a relic

```suggestion
        // Check for missing required options
```

core/Azure.Mcp.Core/src/Options/ParseResultExtensions.cs
```diff
@@ -1,12 +1,36 @@
...
+            if (child is OptionResult optionResult)
+            {
+                var option = optionResult.Option;
+                if (option is null)
+                {
+                    continue;
+                }
+
+                var aliases = option.Aliases;
+                if (aliases.Any(a => string.Equals(a, $"--{Models.Option.OptionDefinitions.RetryPolicy.DelayName}", StringComparison.OrdinalIgnoreCase)) ||
```
jeffhandley: 👍

docs/system-commandline-beta5-migration.md
```diff
@@ -0,0 +1,480 @@
...
+    }
+    finally
+    {
+      _logger.LogTrace("Finished running '{Command}'.", command.Name);
+    }
+  });
+}
+```
+
+### BaseCommand and Helpers
```
jeffhandley: This is where I've reviewed to. So far so good!
I will resume my review Wednesday.

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -42,8 +44,35 @@ protected virtual void HandleException(CommandContext context, Exception ex)
         context.Activity?.SetStatus(ActivityStatusCode.Error)?.AddTag(TagName.ErrorDetails, ex.Message);
 
         var response = context.Response;
+
+        // Map System.CommandLine "Option '--x' is required." exceptions to a 400 validation response
+        // with a standardized missing-options message so callers don't receive a 500.
+        if (ex is InvalidOperationException && ex.Message != null && ex.Message.IndexOf("is required", StringComparison.OrdinalIgnoreCase) >= 0)
```
jeffhandley: Checking the exception message for these string matches seems brittle. Ideally we'd get strucutured data on a custom exception type, but this might be all that can be done right now.

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -108,6 +183,38 @@ static void SetValidationError(CommandResponse? response, string errorMessage)
         }
     }
 
+    private static string NormalizeName(string? name) => (name ?? string.Empty).TrimStart('-', '/');
+
+    private static bool HasOptionResult(CommandResult commandResult, Option option)
```
jeffhandley: You might be able to use [SymbolResult.GetResult Method (System.CommandLine.Parsing)](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.symbolresult.getresult?view=net-9.0-pp#system-commandline-parsing-symbolresult-getresult%28system-commandline-option%29) here.

https://github.com/dotnet/dotnet/blob/30000d883e06c122311a66894579bc12329a09d4/src/command-line-api/src/System.CommandLine/Parsing/SymbolResult.cs#L79-L84

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
```diff
@@ -130,4 +161,21 @@ private static bool IsArrayOption(Option option)
             _ => element.GetRawText()
         };
     }
+
+    /// <summary>
+    /// Safe TryGetValue for ParseResult to avoid System.CommandLine throwing when options are missing.
+    /// </summary>
+    public static bool TryGetValue<T>(this ParseResult parseResult, Option<T> option, out T? value)
```
jeffhandley: Consider moving `TryGetValue<T>(this ParseResult parseResult, Option<T> option, out T? value)` and `TryGetValue<T>(this CommandResult commandResult, Option<T> option, out T? value)` from `CommandResultExtensions.cs` into the same file.

core/Azure.Mcp.Core/src/TestUtilities/ArgSplitter.cs
```diff
@@ -0,0 +1,64 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Azure.Mcp.TestUtilities;
+
+public static class ArgSplitter
+{
+    // Split command-line strings into arguments while respecting double quotes
+    public static string[] SplitArgs(string commandLine)
```
jeffhandley, line 11: Would [System.CommandLine.Parsing.CommandLineParser.SplitCommandLine(String)](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.commandlineparser.splitcommandline?view=net-9.0-pp) meet your needs here?

docs/system-commandline-beta5-migration.md
```diff
@@ -0,0 +1,480 @@
...
+3) Core Wiring
+- [x] `core/Azure.Mcp.Core` CommandFactory: replace handler wiring with `SetAction((ParseResult, CancellationToken) => Task<int>)`.
+- [x] Start telemetry/activity around action, tag relevant properties.
+- [x] Introduce centralized validation hook (e.g., `ICommandValidator` chain) executed before per-command validation.
+- [x] Call per-command `Validate(CommandResult, CommandResponse)` and return early on failure.
+- [x] Bind options once after validation via `BindOptions(ParseResult)` (create/standardize if needed).
+- [x] Execute command, write JSON response, and return exit code. Centralize exception → response mapping.
+- [x] `BaseCommand`: update to use `parseResult.GetValue(...)`; update `.IsRequired` → `.Required`; ensure `BindOptions(ParseResult)` is available and AOT-safe.
+- [ ] `SystemCommandLineExtensions`: update helpers to new APIs (`Options.Add`, `Subcommands.Add`, `GetValue`, action model) as needed.
+- [x] `SystemCommandLineExtensions`: update helpers to new APIs (`Options.Add`, `Subcommands.Add`, `GetValue`, action model) as needed. (Implemented via `core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs`.)
```
jeffhandley: These two todos are redundant; the second is checked off.

tests/TestUtilities/ArgSplitter.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Azure.Mcp.TestUtilities;
+
+public static class ArgSplitter
+{
+    // Split command-line strings into arguments while respecting double quotes
+    public static string[] SplitArgs(string commandLine)
```
jeffhandley: Another place to consider [System.CommandLine.Parsing.CommandLineParser.SplitCommandLine(String)](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.commandlineparser.splitcommandline?view=net-9.0-pp)

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
jeffhandley: And you might be able to use [System.CommandLine.Parsing.SymbolResult.GetResult](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.symbolresult.getresult?view=net-9.0-pp#system-commandline-parsing-symbolresult-getresult%28system-commandline-option%29) to achieve this logic. Note that method respects default values too.

tools/Azure.Mcp.Tools.Monitor/src/Commands/Table/TableListCommand.cs
```diff
@@ -31,28 +29,28 @@ List all tables in a Log Analytics workspace. Requires {WorkspaceOptionDefinitio
...
-        command.AddOption(_tableTypeOption);
+        command.Options.Add(_tableTypeOption);
         RequireResourceGroup();
     }
 
     protected override TableListOptions BindOptions(ParseResult parseResult)
     {
         var options = base.BindOptions(parseResult);
-        options.TableType = parseResult.GetValueForOption(_tableTypeOption) ?? MonitorOptionDefinitions.TableType.GetDefaultValue();
+        options.TableType = parseResult.GetValue(_tableTypeOption) ?? (string?)MonitorOptionDefinitions.TableType.GetDefaultValue();
```
jeffhandley: The `?? ...GetDefaultValue()` might be superfluous on these calls, as I think `.GetValue(option)` should respect the default value.

tools/Azure.Mcp.Tools.Monitor/tests/Azure.Mcp.Tools.Monitor.UnitTests/Log/WorkspaceLogQueryCommandTests.cs
```diff
@@ -156,16 +150,7 @@ public async Task ExecuteAsync_CallsServiceWithCorrectParameters()
...
-            "--subscription", _knownSubscription,
-            "--workspace", _knownWorkspace,
-            "--resource-group", _knownResourceGroup,
-            "--table", _knownTable,
-            "--query", _knownQuery,
-            "--hours", _knownHours,
-            "--limit", _knownLimit,
-            "--tenant", _knownTenant
-        ]);
+        var args = _commandDefinition.Parse(ArgSplitter.SplitArgs($"--subscription {_knownSubscription} --workspace {_knownWorkspace} --resource-group {_knownResourceGroup} --table {_knownTable} --query \"{_knownQuery}\" --hours {_knownHours} --limit {_knownLimit} --tenant {_knownTenant}"));
```
jeffhandley: For all these occurrences, you can probably use [System.CommandLine.Parsing.CommandLineParser.Parse(Command, String)](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.commandlineparser.parse?view=net-9.0-pp)

tools/Azure.Mcp.Tools.CloudArchitect/tests/Azure.Mcp.Tools.CloudArchitect.UnitTests/Design/DesignCommandTests.cs
```diff
@@ -208,6 +215,68 @@ public async Task ExecuteAsync_HandlesComplexEscapingScenarios()
         Assert.Equal(complexAnswer, answerValue);
     }
 
+    private static string[] SplitCommandLine(string commandLine)
```
jeffhandley: Did you find this `SplitCommandLine` utility was still needed within the tests?

jeffhandley (approved):

core/Azure.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
```diff
@@ -32,6 +32,26 @@ public sealed class CommandFactoryToolLoader(
 
     public const string RawMcpToolInputOptionName = "raw-mcp-tool-input";
 
+    private static string NormalizeName(string? name) => (name ?? string.Empty).TrimStart('-', '/');
```
alzimmermsft: Mind adding comments to why `NormalizeName` does what it does for anyone who looks at this in the future? Right now, it's a bit cryptic on what it's doing with `TrimStart`.

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand.cs
```diff
@@ -17,21 +18,41 @@ public abstract class SubscriptionCommand<
...
-        command.AddOption(_subscriptionOption);
+        command.Options.Add(_subscriptionOption);
 
-        command.AddValidator(result =>
+        // Command-level validation for presence: allow either --subscription or AZURE_SUBSCRIPTION_ID
+        // This mirrors the prior behavior that preferred the explicit option but fell back to env var.
+        command.Validators.Add(commandResult =>
         {
-            var subscriptionValue = result.GetValueForOption(_subscriptionOption);
+            static string Normalize(string? s) => (s ?? string.Empty).TrimStart('-', '/');
```
alzimmermsft: Should this be added as an internal helper method somewhere? This has been defined twice with the same logic.

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand.cs
```diff
@@ -17,21 +18,41 @@ public abstract class SubscriptionCommand<
...
+            // Look for an explicit option result among the command's children by comparing normalized
+            // option names and aliases rather than relying on reference equality. Consider an option
+            // present only if it has at least one non-whitespace token.
+            var optionResult = commandResult.Children.OfType<OptionResult>()
+                .FirstOrDefault(or =>
+                {
+                    if (or.Tokens == null || or.Tokens.Count == 0)
+                        return false;
+                    if (!string.IsNullOrWhiteSpace(or.Tokens[0].Value) &&
+                        (string.Equals(Normalize(or.Option.Name), Normalize(_subscriptionOption.Name), StringComparison.OrdinalIgnoreCase)
```
alzimmermsft: Do we need to normalize the subscription option name? Isn't that a constant that cannot be changed?

Or is the .NET runtime smart enough to realize that and amoratize it?

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -42,8 +43,35 @@ protected virtual void HandleException(CommandContext context, Exception ex)
...
+        // with a standardized missing-options message so callers don't receive a 500.
+        if (ex is InvalidOperationException && ex.Message != null && ex.Message.IndexOf("is required", StringComparison.OrdinalIgnoreCase) >= 0)
+        {
+            try
+            {
+                // Extract option tokens like --cluster from the exception message
+                var missing = System.Text.RegularExpressions.Regex.Matches(ex.Message, "--[A-Za-z0-9\\-]+")
+                    .Select(m => m.Value)
+                    .Where(s => !string.IsNullOrWhiteSpace(s))
+                    .ToList();
```
alzimmermsft: Any need to collect this into a list rather than a string joined by `, ` right away? Both cases you can check if there is a value.

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -70,25 +98,71 @@ public virtual ValidationResult Validate(CommandResult commandResult, CommandRes
         var result = new ValidationResult { IsValid = true };
 
         var missingOptions = commandResult.Command.Options
-            .Where(o => o.IsRequired && IsOptionValueMissing(commandResult.GetValueForOption(o)))
-            .Select(o => $"--{o.Name}")
+            .Where(o => o.Required && !o.HasDefaultValue)
+            .Where(o => !commandResult.HasOptionResult(o))
```
alzimmermsft: Could these `Where`s be merged?

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -70,25 +98,71 @@ public virtual ValidationResult Validate(CommandResult commandResult, CommandRes
...
+            .Select(o => $"--{NormalizeName(o.Name)}")
             .ToList();
 
-        if (missingOptions.Count > 0 || !string.IsNullOrEmpty(commandResult.ErrorMessage))
+        if (missingOptions.Count > 0)
```
alzimmermsft: Same question on the need to collect into a list before concatenation just to check if the list has values

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -108,17 +182,17 @@ static void SetValidationError(CommandResponse? response, string errorMessage)
         }
     }
 
-    private static bool IsOptionValueMissing(object? value)
-    {
-        return value == null || (value is string str && string.IsNullOrWhiteSpace(str));
-    }
+    private static string NormalizeName(string? name) => (name ?? string.Empty).TrimStart('-', '/');
```
alzimmermsft: Another instance of the same method with the same logic, let's make a shared helper method

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -70,25 +98,71 @@ public virtual ValidationResult Validate(CommandResult commandResult, CommandRes
...
+            SetValidationError(commandResponse, result.ErrorMessage);
+            return result;
         }
 
         // Check logical requirements (e.g., resource group requirement)
         if (result.IsValid && _requiresResourceGroup)
         {
-            var rg = commandResult.GetValueForOption(OptionDefinitions.Common.ResourceGroup);
-            if (string.IsNullOrWhiteSpace(rg))
+            var hasRg = commandResult.HasOptionResult(OptionDefinitions.Common.ResourceGroup);
```
alzimmermsft: Given `hasRg` is only used in the `if` clause, could we remove the local variable and just put this statement in the `if` clause?

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -127,8 +201,15 @@ protected void RequireResourceGroup()
         _requiresResourceGroup = true;
     }
 
-    protected string? GetResourceGroup(ParseResult parseResult) =>
-        _usesResourceGroup ? parseResult.GetValueForOption(OptionDefinitions.Common.ResourceGroup) : null;
+    protected string? GetResourceGroup(ParseResult parseResult)
+    {
+        if (!UsesResourceGroup)
```
alzimmermsft, line 161: This is just a nit, not by this PR, is there a better naming convention we could use for `UseResourceGroup` and `UsesResourceGroup`? The names are so similar but what they do is so extremely different and this seems like a good case where an agent and human alike would pitfall and accidentally use the wrong one causing a hard to find code bug.

For reference, when I initially read this, given `UseResourceGroup` comes first, I asked myself "why are we setting the use of resource group when checking for it being set", as I didn't catch the additional `s` for it actually calling `UsesResourceGroup`.

I know this sound bad but maybe `IsResourceGroupUsed`?

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
```diff
@@ -113,6 +127,18 @@ private static bool IsArrayOption(Option option)
         return CollectionTypeHelper.IsArrayType(option.ValueType);
     }
 
+    private static string NormalizeName(string n) => (n ?? string.Empty).TrimStart('-', '/');
```
alzimmermsft: Yet another instance of `NormalizeName`

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
```diff
@@ -30,8 +29,10 @@ public static ParseResult ParseFromDictionary(this Command command, IReadOnlyDic
 
         foreach (var (key, value) in arguments)
         {
+            // lookup by normalized name or any alias (case-insensitive)
             var option = command.Options.FirstOrDefault(o =>
-                o.Name.Equals(key, StringComparison.OrdinalIgnoreCase));
+                string.Equals(NormalizeName(o.Name), key, StringComparison.OrdinalIgnoreCase)
+                || o.Aliases.Any(a => string.Equals(NormalizeName(a), key, StringComparison.OrdinalIgnoreCase)));
```
alzimmermsft: This logic is also copied in a few spots, maybe we should also add a helper to get an option based on name or alias?

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
```diff
@@ -77,8 +78,21 @@ public static ParseResult ParseFromDictionary(this Command command, IReadOnlyDic
...
+        var option = command.Options.FirstOrDefault(o =>
+            string.Equals(NormalizeName(o.Name), "raw-mcp-tool-input", StringComparison.OrdinalIgnoreCase)
+            || o.Aliases.Any(a => string.Equals(NormalizeName(a), "raw-mcp-tool-input", StringComparison.OrdinalIgnoreCase))
+        );
+
+        option ??= command.Options.FirstOrDefault();
+
+        if (option == null)
+        {
+            return command.Parse(Array.Empty<string>());
```
alzimmermsft: I saw elsewhere in this PR you switched `Array.Empty` calls to `[]`, should that be done here too?

core/Azure.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -198,12 +185,21 @@ private void ConfigureCommandHandler(Command command, IBaseCommand implementatio
                 {
                     Console.WriteLine(JsonSerializer.Serialize(response, _srcGenWithOptions.CommandResponse));
                 }
+
+                var status = response.Status;
+                if (status < 200 || status >= 300)
```
alzimmermsft, line 190: I don't know if we can consider this change safe due to some corner cases. For example, in some older Azure services deleting a resource that doesn't exist may return a `404` rather than the recommended `204` (https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-return-codes)

core/Azure.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -198,12 +185,21 @@ private void ConfigureCommandHandler(Command command, IBaseCommand implementatio
...
+                }
+
+                return status;
             }
             catch (Exception ex)
             {
                 _logger.LogError("An exception occurred while executing '{Command}'. Exception: {Exception}",
                     command.Name, ex);
                 activity?.SetStatus(ActivityStatusCode.Error)?.AddTag(TagName.ErrorDetails, ex.Message);
+                return 1;
```
alzimmermsft, line 202: Any reason `return 1;` was added here?

core/Azure.Mcp.Core/src/Commands/IBaseCommand.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.CommandLine.Parsing;
```
alzimmermsft, line 4: Don't see any code changes in this file, why was a `using` added?

core/Azure.Mcp.Core/src/Extensions/CommandResultExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        // For zero-arity options (e.g., bool switches), identifier presence indicates explicit usage
+        var expectsValue = option.Arity.MaximumNumberOfValues > 0;
+        if (!expectsValue)
+        {
+            return result.IdentifierTokenCount > 0;
+        }
+
+        // For value-taking options, consider present only if there is at least one non-empty value token
+        var hasNonEmptyValue = result.Tokens is { Count: > 0 } && result.Tokens.Any(t => !string.IsNullOrWhiteSpace(t.Value));
```
alzimmermsft, line 27: Just wondering this, why use `result.Tokens is { Count: > 0}` over `result.Tokens.Count > 0`?

core/Azure.Mcp.Core/src/Options/ParseResultExtensions.cs
```diff
@@ -1,12 +1,46 @@
...
 public static class ParseResultExtensions
 {
-    public static bool HasAnyRetryOptions(this System.CommandLine.Parsing.ParseResult parseResult) =>
-        // Check if any retry-related options were specified on the command line
-        parseResult.CommandResult.Children
-            .Any(token => token.Symbol.Name.StartsWith("retry-", StringComparison.OrdinalIgnoreCase));
+    public static bool HasAnyRetryOptions(this System.CommandLine.ParseResult parseResult)
+    {
+        // Compare normalized names (trim leading '-' or '/') so we don't depend on alias formatting
+        static string Normalize(string s) => (s ?? string.Empty).TrimStart('-', '/');
```
alzimmermsft: Another instance of this

core/Azure.Mcp.Core/src/Options/ParseResultExtensions.cs
```diff
@@ -1,12 +1,46 @@
...
+
+        var retryNames = new[]
+        {
+            Models.Option.OptionDefinitions.RetryPolicy.DelayName,
+            Models.Option.OptionDefinitions.RetryPolicy.MaxDelayName,
+            Models.Option.OptionDefinitions.RetryPolicy.MaxRetriesName,
+            Models.Option.OptionDefinitions.RetryPolicy.ModeName,
+            Models.Option.OptionDefinitions.RetryPolicy.NetworkTimeoutName,
+        };
```
alzimmermsft: Does this need to be `new`ed up every time this method is called?

core/Azure.Mcp.Core/src/Options/ParseResultExtensions.cs
```diff
@@ -1,12 +1,46 @@
...
+                if (option is null)
+                {
+                    continue;
+                }
+
+                var name = Normalize(option.Name);
+                if (retryNames.Any(rn => string.Equals(rn, name, StringComparison.OrdinalIgnoreCase)))
+                    return true;
+
+                var aliases = option.Aliases ?? Array.Empty<string>();
```
alzimmermsft: Another code consistency question on `Array.Empty` vs `[]`

core/Azure.Mcp.Core/src/Options/ParseResultExtensions.cs
```diff
@@ -1,12 +1,46 @@
...
+            if (child is OptionResult optionResult)
+            {
+                var option = optionResult.Option;
+                if (option is null)
+                {
+                    continue;
+                }
+
+                var name = Normalize(option.Name);
+                if (retryNames.Any(rn => string.Equals(rn, name, StringComparison.OrdinalIgnoreCase)))
```
alzimmermsft: Code consistency question on creating local variables vs inlining `Normalize` into where it's being used.

servers/Azure.Mcp.Server/src/Azure.Mcp.Server.csproj
```diff
@@ -12,7 +12,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <AssemblyName>$(CliName)</AssemblyName>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
alzimmermsft: This seems like a substantial change, any reason behind it?

tools/Azure.Mcp.Tools.Aks/src/Options/AksOptionDefinitions.cs
```diff
@@ -8,10 +8,10 @@ public static class AksOptionDefinitions
     public const string ClusterName = "cluster";
 
     public static readonly Option<string> Cluster = new(
-        $"--{ClusterName}",
-        "AKS Cluster name."
+        $"--{ClusterName}"
     )
     {
-        IsRequired = true
+        Description = "AKS Cluster name.",
```
alzimmermsft, line 14: Not your PR, just general comment, I wonder if this description should be changed to

"Azure Kubernetes service (AKS) cluster name."

That way it can key in on both the full name and acronym.

tools/Azure.Mcp.Tools.AppConfig/src/Options/AppConfigOptionDefinitions.cs
```diff
@@ -13,70 +13,70 @@ public static class AppConfigOptionDefinitions
...
+        Required = false,
         AllowMultipleArgumentsPerToken = true
     };
 
     public static class KeyValueList
     {
         public static readonly Option<string> Key = new(
-            $"--{KeyName}",
-            "Specifies the key filter, if any, to be used when retrieving key-values. The filter can be an exact match, for example a filter of 'foo' would get all key-values with a key of 'foo', or the filter can include a '*' character at the end of the string for wildcard searches (e.g., 'App*'). If omitted all keys will be retrieved."
+            $"--{KeyName}"
```
alzimmermsft, line 67: Another observation not related to your change but given it's changing these files.

I wonder how the LLM handles that there are two `Key` options but in different classes. Same goes for `Label`.

This threw me off for a moment.

tools/Azure.Mcp.Tools.AppConfig/src/Options/AppConfigOptionDefinitions.cs
```diff
@@ -13,70 +13,70 @@ public static class AppConfigOptionDefinitions
     public const string TagsName = "tags";
 
     public static readonly Option<string> Account = new(
-        $"--{AccountName}",
-        "The name of the App Configuration store (e.g., my-appconfig)."
+        $"--{AccountName}"
     )
     {
-        IsRequired = true
+        Description = "The name of the App Configuration store (e.g., my-appconfig).",
```
alzimmermsft, line 19: Another observation, I wonder if the order of service name and concept matter for the LLM. E.x., does it work better with `The App Configuration store name (e.g., my-appconfig).` better.

tools/Azure.Mcp.Tools.VirtualDesktop/src/Options/VirtualDesktopOptionDefinitions.cs
```diff
@@ -11,25 +11,34 @@ public static class VirtualDesktopOptionDefinitions
...
 
     public static readonly Option<string> ResourceGroup = new(
-        $"--{ResourceGroupName}",
-        "The name of the Azure resource group. This is a logical container for Azure resources."
-    );
+        $"--{ResourceGroupName}"
+    )
+    {
+        Description = "The name of the Azure resource group. This is a logical container for Azure resources.",
+    };
```
alzimmermsft, line 43: Should this even exist? Isn't this a global defined Option?

tools/Azure.Mcp.Tools.VirtualDesktop/src/Options/VirtualDesktopOptionDefinitions.cs
```diff
@@ -11,25 +11,34 @@ public static class VirtualDesktopOptionDefinitions
...
 
     public static readonly Option<string> HostPoolResourceIdOption = new(
-        $"--{HostPoolResourceId}",
-        "The Azure resource ID of the host pool. When provided, this will be used instead of searching by name."
-    );
+        $"--{HostPoolResourceId}"
+    )
+    {
+        Description = "The Azure resource ID of the host pool. When provided, this will be used instead of searching by name."
```
alzimmermsft, line 25: I've noticed quite a number of options that start with `The Azure resource ID of X`, I wonder if a better pattern would be `The X Azure resource ID`. Ex, `The host pool Azure resource ID`.

This puts the service concept first, not sure if that will improve or hurt LLM correlation. Overall, I feel descriptions likely need a deeper dive so we can improve their confidence scores.

core/Azure.Mcp.Core/src/Commands/CommandFactory.cs
jongio: This is just for the ServiceStartCommand, not Azure services.

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
jongio: Will take up later.

core/Azure.Mcp.Core/src/Commands/IBaseCommand.cs
jongio: CommandResult is in new namespace

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand.cs
jongio: That normalization is because of how SCL returns them and how we store them in the option.Name value.

servers/Azure.Mcp.Server/src/Azure.Mcp.Server.csproj
jongio: Will turn back on before merge

core/Azure.Mcp.Core/src/Extensions/CommandResultExtensions.cs
jongio: Null-safety: Using result.Tokens is { Count: > 0 } is null-safe; the pattern fails for null so it won’t throw. A plain result.Tokens.Count > 0 would throw if Tokens is null.

core/Azure.Mcp.Core/src/Commands/CommandFactory.cs
jongio: The delegate passed to SetAction must return an int; on unexpected exceptions, returning 1 signals a generic failure to the host/shell.

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
```diff
@@ -42,8 +45,26 @@ protected virtual void HandleException(CommandContext context, Exception ex)
         context.Activity?.SetStatus(ActivityStatusCode.Error)?.AddTag(TagName.ErrorDetails, ex.Message);
 
         var response = context.Response;
+
+        // Handle structured validation errors first
+        if (ex is CommandValidationException cve)
```
jeffhandley, line 50: :+1: Nice change on the exception handling

core/Azure.Mcp.Core/src/Commands/CommandExtensions.cs
```diff
@@ -113,6 +127,16 @@ private static bool IsArrayOption(Option option)
         return CollectionTypeHelper.IsArrayType(option.ValueType);
     }
 
+    private static string GetOptionToken(Option option)
```
jeffhandley, line 130: `GetOptionToken` is another nice helper!

tools/Azure.Mcp.Tools.Storage/tests/Azure.Mcp.Tools.Storage.UnitTests/Blob/Container/ContainerCreateCommandTests.cs
```diff
@@ -93,11 +93,11 @@ public async Task ExecuteAsync_CreatesContainerWithoutPublicAccess()
...
             .Returns(expectedProperties);
 
-        var args = _parser.Parse([
-            "--account", _knownAccount,
+        var args = _commandDefinition.Parse([
+                "--account", _knownAccount,
             "--container", _knownContainer,
             "--subscription", _knownSubscription
-        ]);
+            ]);
```
jeffhandley: Nit: Some of the indentation became inconsistent here (and below as well).

jeffhandley (approved):
Bravo! :100:

core/Azure.Mcp.Core/src/Commands/BaseCommand.cs
joshfree, line 50: nit: CVE normally means common vulnerabilities & exploits 😁

core/Azure.Mcp.Core/src/Commands/GlobalCommand.cs
```diff
@@ -81,26 +80,52 @@ protected virtual TOptions BindOptions(ParseResult parseResult)
...
+                policy.HasMode = true;
+                policy.Mode = parseResult.GetValue(_retryModeOption);
+            }
+            if (parseResult.GetResult(_retryNetworkTimeoutOption) != null)
+            {
+                policy.HasNetworkTimeoutSeconds = true;
+                policy.NetworkTimeoutSeconds = parseResult.GetValue(_retryNetworkTimeoutOption);
+            }
+
+            // Only assign if at least one flag set (defensive)
```
joshfree, line 124: Should there be a log message for the else case or anything like that added for debugging / user awareness?

alzimmermsft (approved):

PR #65, opened 2025-08-27
Patch `create github release`

scbedd:
We are using `gh api`. The `gh api` does not support `**` pattern matching.

Therefore, we need to update the upload to more align with the actual artifact at rest.

image


vcolin7 (approved):
:shipit:

PR #315, opened 2025-09-04
Remove ContinueOnError from check-spelling step

weshaggard:
Turn spelling errors into warnings instead of failing the build.


weshaggard:
If we want spelling to fail for some files, we can add a separate check for those files to make them fail

PR #321, opened 2025-09-04
Make VSIX use Trimmed Binaries by Default with Opt-Out Support.

g2vinay:
Defaults VSIX to azmcp binaries.
Adds opt-out support to swtich back to self-contained binaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This repository contains core libraries, test frameworks, engineering systems, p
 [Azure MCP README]: https://github.com/microsoft/mcp/blob/main/servers/Azure.Mcp.Server/README.md
 [Azure MCP CHANGELOG]: https://github.com/microsoft/mcp/blob/main/servers/Azure.Mcp.Server/CHANGELOG.md
 [Azure MCP Source Code]: https://github.com/microsoft/mcp/blob/main/servers/Azure.Mcp.Server
-[Azure MCP Releases]: https://github.com/microsoft/mcp/releases?q=Azure
+[Azure MCP Releases]: https://github.com/microsoft/mcp/releases?q=Azure.Mcp.Server-0
 [Azure MCP Documentation]: https://learn.microsoft.com/azure/developer/azure-mcp-server/
 [Azure MCP Troubleshooting]: https://github.com/microsoft/mcp/blob/main/servers/Azure.Mcp.Server/TROUBLESHOOTING.md
 [Azure MCP Support]: https://github.com/microsoft/mcp/blob/main/servers/Azure.Mcp.Server/SUPPORT.md
--- a/docs/azmcp-commands.md
+++ b/docs/azmcp-commands.md
@@ -569,6 +569,9 @@ azmcp aks cluster get --subscription <subscription> \
 
 # List AKS clusters in a subscription
 azmcp aks cluster list --subscription <subscription>
+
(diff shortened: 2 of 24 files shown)


eng/scripts/Pack-Vsix.ps1
```diff
@@ -94,6 +94,36 @@ Processing VSIX packaging: $vsixBaseName
...
+            $nativeDirs = @(
+                Join-Path $serverDirectory.FullName ("$($serverName)-native"),
+                Join-Path $serverDirectory.Parent.FullName ("$($serverName)-native"),
+                Join-Path $ArtifactsPath ("$($serverName)-native")
+            )
+            $nativeCopied = $false
+            foreach ($nd in $nativeDirs) {
+                if (Test-Path $nd) {
+                    $nativePlatformDir = Join-Path $nd $platformName
+                    if (Test-Path $nativePlatformDir) {
```
alzimmermsft: Should be able to reduce indentations by eagerly joining the possible `$nd ` path and `$platformName` path as `Join-Path` doesn't care if the paths are real. And should just need one `Test-Path` call.

alzimmermsft (commented):
The Copolit comment around changing the `throw` to just a console log is very valid. Should determine whether execution should continue if no Azure MCP binary can be found.

jongio (approved):

eng/vscode/package.json
```diff
@@ -127,6 +127,11 @@
                     "type": "boolean",
                     "default": false,
                     "markdownDescription": "Start the Azure MCP server in read-only mode (operations that modify resources will be disabled)."
+                },
+                "azureMcp.useNonTrimmedBinaries": {
+                    "type": "boolean",
+                    "default": false,
+                    "markdownDescription": "Use non-trimmed Azure MCP server binaries from the `server` folder instead of the default trimmed `server/native` folder. Enable this to opt-out of trimmed binaries for compatibility or debugging purposes."
```
g2vinay: @anuchandy should we tweak this description in any way ?

eng/scripts/Pack-Vsix.ps1
g2vinay: This change is removed.

eng/vscode/package.json
anuchandy: I’m wondering if VSIX users even have visibility into the different folders that contain azmcp and its dependencies. If not, maybe we should adjust the wording slightly. For example:

“Opt out of the trimmed single-file Azure MCP Server and use the standard self-contained version (with the .NET runtime, entry point executable, and all DLLs).”

anuchandy (approved):
Looks good to me, thanks @g2vinay! Just one small comment on the message we show

eng/scripts/Pack-Vsix.ps1
```diff
@@ -23,6 +23,17 @@ if(!(Test-Path $ArtifactsPath)) {
     exit 1
 }
 
+$serverJsonFilePairs = Get-ChildItem -Path $ArtifactsPath -Filter "wrapper.json" -Recurse
```
g2vinay, line 26: @hallipr I have ported the Pack vsix changes off your PR, in case it doesn't merge tomorrow before release.

joshfree:
Closing 1-month old stale PR. @g2vinay please update / reopen if still needed

Issue #170, opened 2025-09-03
[NuGet] Publish azmcp as a dotnet global tool


alzimmermsft:
Taking in consideration this is a dotnet command line application, are you considering to publish as a global tool that we can install as long as the dotnet sdk is installed?

alzimmermsft:
Copied from https://github.com/Azure/azure-mcp/issues/422

PR #379, opened 2025-09-09
Add correct / required nuget metadata

chidozieononiwu:
## What does this PR do?
Updates various .nuspec metadata to meet nuget requirements

## GitHub issue number?
https://github.com/microsoft/mcp/issues/170


eng/scripts/Pack-Nuget.ps1
```diff
@@ -75,6 +75,7 @@ try {
             -replace '\$\(PackageDescription\)', $serverProjectProperties.Description `
             -replace '\$\(PackageId\)', $serverProjectProperties.PackageId `
             -replace '\$\(PackageVersion\)', $packageVersion `
+			-replace '\$\(ServerName\)', $serverProjectProperties.CliName `
```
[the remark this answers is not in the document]
weshaggard:
```suggestion
            -replace '\$\(ServerName\)', $serverProjectProperties.CliName `
```

eng/scripts/Pack-Nuget.ps1
```diff
@@ -112,6 +114,7 @@ try {
 				-replace "__Authors__", $wrapperPackageJson.author `
 				-replace "__Description__", ($serverProjectProperties.PackageDescription -replace '\$\(RuntimeIdentifier\)', $platformOSArch) `
 				-replace "__RepositoryUrl__", $RepoUrl `
+                -replace "__ProjectUrl__", "$RepoUrl/tree/main/servers/$serverName" `
```
weshaggard:
```suggestion
				-replace "__ProjectUrl__", "$RepoUrl/tree/main/servers/$serverName" `
```

eng/scripts/Pack-Nuget.ps1
```diff
@@ -84,6 +85,7 @@ try {
             -replace "__Authors__", $wrapperPackageJson.author `
             -replace "__Description__", $wrapperPackageJson.description `
             -replace "__Tags__", $serverProjectProperties.PackageTags `
+            -replace "__ProjectUrl__", "$RepoUrl/tree/main/servers/$serverName" `
```
joelverhagen, line 88: is `ServerName` = `CliName` like above? `azmcp` vs. `Azure.Mcp.Server` the latter of which is in the URL
https://github.com/microsoft/mcp/tree/main/servers/Azure.Mcp.Server

chidozieononiwu: It is the later in this case.

eng/scripts/Pack-Nuget.ps1
```diff
@@ -75,6 +75,7 @@ try {
             -replace '\$\(PackageDescription\)', $serverProjectProperties.Description `
```
hallipr, line 75: If you use string.Replace here, you don't have to escape anything

```
            .Replace('$(PackageDescription)', $serverProjectProperties.Description) `
```

Also, the right side of a `-replace` is still interpreted as a regex replacement pattern   `$1something$2`.  To be safe, you should call [Regex]::Replace() on each of the right hand side args unless you know they don't contain any regex replacement special characters

hallipr (approved):

PR #534, opened 2025-09-23
Update build scripts to use common metadata file

### LINKED ISSUE #53: Build artifacts should have package-generic metadata
The output from the Build stage is heavily NPM flavored.  It's currently just the unpacked npm modules complete with package.json.

To support all of the other package formats and their required metadata, Build-Module.ps1 (should be renamed to Build-Code.ps1) should output the product of `dotnet publish` along with a json manifest containing all of the metadata that would be needed by following Pack and Release jobs.

i.e.  Less like:
```json
{
  "name": "@azure/mcp-win32-x64",
  "version": "0.5.9",
  "description": "Azure MCP Server - Model Context Protocol implementation for Azure, for win32 on x64",
  "author": "Microsoft Corporation",
  "homepage": "https://github.com/Microsoft/mcp/blob/main/servers/Azure.Mcp.Server#readme",
  "license": "MIT",
  "keywords": [
    "azure",
    "mcp",
    "model-context-protocol"
  ],
  "bugs": {
    "url": "https://github.com/microsoft/mcp/issues"
  },
  "repository": {
    "url": "https://github.com/microsoft/mcp.git",
    "type": "git"
  },
  "engines": {
    "node": ">=20.0.0"
  },
  "main": "./index.js",
  "bin": {
    "azmcp-win32-x64": "./dist/azmcp.exe"
  },
  "os": [
    "win32"
  ],
  "cpu": [
    "x64"
  ]
}
```

More like:
```json
{
    "name": "Azure.Mcp.Server",
    "version": "1.2.3-beta.1",
    "dotnetOs": "win",
    "nodeOs": "win32",
    "cpu": "x64",
    "native": false,    
    "description": "Generic package agnostic description",
    "npmDescription": "Maybe a server likes to have a custom description in npm",
    "readmePath": "servers/Azure.Mcp.Server/README.md",
    "npmPackageName": "@azure/mcp",
    "dockerImageName": "azure/mcp",
    "vsixPackageName": "vscode-azure-mcp-extension"
} 
```

The json should contain all of the data that differs between servers and platform builds needed to package any of our target formats.   The packaging format would combine this data with the generic templates (npm's package.json, etc) from source to produce the packages.

hallipr:
## What does this PR do?
- Updates all build, test, package and release scripts to use a common build_info.json file
  - This file coordinates version numbers, package properties and platform details for all build artifacts
  - all other build and pack scripts should use build_info rather than pulling properties from project files
- Removes the static build matrix
  - build_info.json also includes dynamically generated build and tests matrices
  - rather than special casing steps in yaml, we should update the build matrix in the script and use simple boolean variables to conditionally run pipeline steps
- Updates the pack scripts to be server-generic.  All ids and server specific settings that go into a package should be pulled from build_info.json, which is built using properties from the servers' project files

Sample [build_info.json](https://github.com/user-attachments/files/22647952/build_info.json)

## GitHub issue number?
closes #53

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:
    - [ ] Updated command list in `/docs/azmcp-commands.md` and/or `/docs/fabric-commands.md`
    - [ ] Updated test prompts in `/docs/e2eTestPrompts.md`


eng/pipelines/templates/jobs/build.yml
```diff
@@ -97,30 +98,12 @@ jobs:
...
-        -Architecture '$(Architecture)'
-        -CleanBuild
-        -BuildNative
+  - template: /eng/common/pipelines/templates/steps/publish-1es-artifact.yml
+    parameters:
+      ArtifactPath: $(Build.ArtifactStagingDirectory)
+      ArtifactName: build_$(System.JobName)
 
-  # --- VS Code Extension Packaging Steps ---
+  # --- VS Code Extension Component Detection ---
```
weshaggard, line 109: We probably don't need to manually inject CG below as it should be ran in our build jobs already.

eng/pipelines/build-matrix.json
```diff
@@ -4,10 +4,10 @@
       "linux": {"OSVmImage": "env:LINUXVMIMAGE", "Pool": "env:LINUXPOOL"},
       "osx": {"OSVmImage": "env:MACVMIMAGE","Pool": "env:MACPOOL"},
       "win": {"OSVmImage": "env:WINDOWSVMIMAGE", "Pool": "env:WINDOWSPOOL"}
-   },
-   "Architecture": {
-      "x64": {"Architecture": "x64", "RunTests": "true" },
-      "arm64": {"Architecture": "arm64", "RunTests": "false"}
+    },
+    "Architecture": {
+      "x64": {"Architecture": "x64", "RunTests": "true", "BuildNative": "true" },
```
weshaggard: didn't we only want one build native?

eng/scripts/Build-Code.ps1
```diff
@@ -3,136 +3,124 @@
...
     [ValidateSet('x64','arm64')]
-    [string] $Architecture,
+    [string[]] $Architecture,
+
     [Parameter(ParameterSetName='AllPlatforms')]
     [switch] $AllPlatforms
 )
 
-$ErrorActionPreference = 'Stop'
+$ErrorActionPreference = 'Continue'
```
weshaggard: I think this is the default.

eng/scripts/New-BuildInfo.ps1
```diff
@@ -0,0 +1,405 @@
...
+
+function CheckVariable($name) {
+    $value = [Environment]::GetEnvironmentVariable($name)
+    if (-not $value) {
+        if ($isPipelineRun) {
+            Write-Error "Environment variable $name is not set."
+            exit 1
+        }
+        $substitute = "Missing-$name"
+        Write-Host "WARNING: Environment variable $name is not set. Using substitute value '$substitute'." -ForegroundColor Yellow
```
weshaggard: We should probably use the logging.ps1 helper functions to these show up as warning in devops as well.

eng/pipelines/templates/jobs/build.yml
hallipr: I think they're doing that so that CG issues with packages that are only used in sign/pack are detected in runs that don't do sign/pack

eng/pipelines/build-matrix.json
hallipr: This static matrix is being removed.  We generate the matrix in New-BuildInfo.ps1 now

servers/Template.Mcp.Server/src/Template.Mcp.Server.csproj
```diff
@@ -5,18 +5,29 @@
...
+    <VsixPackagePrefix></VsixPackagePrefix>
   </PropertyGroup>
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <AssemblyName>$(CliName)</AssemblyName>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <!-- Publishing settings -->
-    <IsAotCompatible>true</IsAotCompatible>
+    <IsAotCompatible>false</IsAotCompatible>
```
anuchandy: Was this intentional? I mean, does the template already contain AOT-incompatible code?

servers/Azure.Mcp.Server/azureicon.png
chidozieononiwu, line 1: Could there not be another Mcp sever requiring this icon?

servers/Template.Mcp.Server/src/Template.Mcp.Server.csproj
hallipr: I need a test server to see the automatic "native: true/false" in New-BuildInfo.ps1.   I want it to build out a server x platform matrix that only includes native legs if the server says it supports native.

servers/Azure.Mcp.Server/azureicon.png
hallipr: Each server should supply a path to its icon.  If the azure icon becomes common between 2 servers, we could move the icon to azure core's directory, or we could just put the same file in both server directories.

servers/Fabric.Mcp.Server/src/Fabric.Mcp.Server.csproj
```diff
@@ -5,17 +5,37 @@
...
+
+    <!-- npm settings -->
     <NpmPackageName>@microsoft/fabric-mcp</NpmPackageName>
     <NpmPackageKeywords>fabric,mcp,model-context-protocol</NpmPackageKeywords>
-    <PackageId>Microsoft.Fabric.Mcp</PackageId>
-    <ToolCommandName>$(CliName)</ToolCommandName>
-    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+
+    <!-- docker settings -->
+    <DockerImageName></DockerImageName>
```
scbedd, line 20: Intentionally empty?

eng/scripts/helpers/PathHelpers.ps1
```diff
@@ -0,0 +1,19 @@
+function Get-RepoRelativePath {
+    [CmdletBinding()]
+    param(
+        [parameter(Mandatory, ValueFromPipeline)]
+        [string] $Path,
+        [switch] $NormalizeSeparators
+    )
+
+    process {
+        $relativePath = Resolve-Path -LiteralPath $Path -Relative -RelativeBasePath $RepoRoot
```
scbedd: $RepoRoot is not defined in this script anywhere. It's probably running as part of a script which means that $RepoRoot is defined externally yeah?

eng/scripts/Build-Code.ps1
benbp: This should mean it overrides any outer error action preference that may be set to error I think. One question is if we should restore the outer preference after completing work.

eng/scripts/Build-Code.ps1
```diff
@@ -3,136 +3,124 @@
...
+    [switch] $Native,
     [string] $ServerName,
-    [Parameter(Mandatory=$true, ParameterSetName='SpecificPlatform')]
+
+    [Parameter(Mandatory, ParameterSetName='SpecificPlatform')]
+    [Alias('OS')]
     [ValidateSet('windows','linux','macOS')]
-    [string] $OperatingSystem,
-    [Parameter(Mandatory=$true, ParameterSetName='SpecificPlatform')]
+    [string[]] $OperatingSystem,
```
benbp, line 20:
```suggestion
    [string[]] $OperatingSystems,
```

maybe, since it's an array

eng/scripts/Build-Code.ps1
```diff
@@ -3,136 +3,124 @@
...
     [switch] $DebugBuild,
     [switch] $CleanBuild,
-    [switch] $BuildNative,
+    [switch] $Native,
     [string] $ServerName,
-    [Parameter(Mandatory=$true, ParameterSetName='SpecificPlatform')]
+
+    [Parameter(Mandatory, ParameterSetName='SpecificPlatform')]
+    [Alias('OS')]
     [ValidateSet('windows','linux','macOS')]
```
benbp: I know it's the official spelling, but can we just use a lowercase `macos` for the accepted value?

servers/Azure.Mcp.Server/src/Azure.Mcp.Server.csproj
```diff
@@ -1,18 +1,31 @@
...
-    <PackageIcon>azureicon.png</PackageIcon>
+
+    <!-- dnx/nuget settings -->
+    <DnxPackageId>Azure.Mcp</DnxPackageId>
+    <DnxToolCommandName>$(CliName)</DnxToolCommandName>
+    <DnxPackageTags>azure,mcp</DnxPackageTags>
+
+    <!-- vsix settings -->
+    <VsixPackagePrefix>vscode-azure-mcp</VsixPackagePrefix>
+    <VsixDescription>Provides Model Context Protocol (MCP) integration and tooling for Azure in Visual Studio Code.</VsixDescription>
```
benbp, line 25:
```suggestion
    <VsixDescription>Provides Model Context Protocol (MCP) integration and tooling for Azure in Visual Studio Code</VsixDescription>
```

eng/scripts/New-BuildInfo.ps1
```diff
@@ -0,0 +1,428 @@
...
+    [string] $OutputPath,
+    [string] $ServerName,
+    [switch] $IncludeNative,
+    [switch] $TestPipeline
+)
+
+. "$PSScriptRoot/../common/scripts/common.ps1"
+. "$PSScriptRoot/helpers/PathHelpers.ps1"
+$RepoRoot = $RepoRoot.Path.Replace('\', '/')
+$isPipelineRun = $env:TF_BUILD -eq 'true'
```
benbp: I would make this a `[switch]` parameter `-CI` with the default being set by the `TF_BUILD` check, that way it's easier to test CI mode locally when development requires it.

eng/scripts/Prepare-Docker.ps1
```diff
@@ -0,0 +1,85 @@
...
+$ignoreMissingArtifacts = $env:TF_BUILD -ne 'true'
+
+if(!$ArtifactsPath) {
+    $ArtifactsPath = "$RepoRoot/.work/build"
+}
+
+if(!$BuildInfoPath) {
+    $BuildInfoPath = "$RepoRoot/.work/build_info.json"
+    if(!(Test-Path $BuildInfoPath)) {
+        & "$PSScriptRoot/New-BuildInfo.ps1" -PublishTarget none -BuildId 12345 -OutputPath $BuildInfoPath
```
benbp: `-BuildId 12345` is supposed to be a dummy value? Or is it a missed placeholder

benbp:
I think it's worth having a markdown doc somewhere explaining at a high level how this all works, how one might test build info generation if needed, and how the build info gets dumped to an artifact/logged for debugging.

benbp:
Some nits and other minor suggestions otherwise this is looking good to me. I worry a bit about complexity growth in the scripts and yaml if we get more and more requests for functionality.

benbp:
Also we should support pipeline tagging in the same way we do for unified pull request builds in the language repos, so that users can filter by server name in the tags dropdown for the pipeline.

eng/scripts/Build-Code.ps1
hallipr: This would be a script scoped variable, no?  So setting it here shouldn't affect any calling script unless it's dot-included

hallipr: i.e. Variable assignment in powershell is scoped, scope variables are accessible by lower scopes, assignments happen at the lower scope unless modified.

SomeScript.ps1
``` powershell
function SomeFunction() {
  Write-Host "Func -> Local: $SomeVariable"
  Write-Host "Func -> Script: $Script:SomeVariable"
  Write-Host "Func -> Global: $Global:SomeVariable"

  Write-Host "`nAssign local variable at function scope"
  $SomeVariable = 'Bar'

  Write-Host "Func -> Local: $SomeVariable"
  Write-Host "Func -> Script: $Script:SomeVariable"
  Write-Host "Func -> Global: $Global:SomeVariable"

  Write-Host "`nAssign script variable at function scope"
  $Script:SomeVariable = 'Baz'

  Write-Host "Func -> Local: $SomeVariable"
  Write-Host "Func -> Script: $Script:SomeVariable"
  Write-Host "Func -> Global: $Global:SomeVariable"
}

Write-Host "Script -> Local: $SomeVariable"
Write-Host "Script -> Script: $Script:SomeVariable"
Write-Host "Script -> Global: $Global:SomeVariable"

Write-Host "`nAssign local variable at script scope"
$SomeVariable = 'Foo'

Write-Host "Script -> Local: $SomeVariable"
Write-Host "Script -> Script: $Script:SomeVariable"
Write-Host "Script -> Global: $Global:SomeVariable"

SomeFunction

Write-Host "Script -> Local: $SomeVariable"
Write-Host "Script -> Script: $Script:SomeVariable"
Write-Host "Script -> Global: $Global:SomeVariable"
```

```
> $SomeValue = 'GlobalValue'
> .\SomeScript.ps1

Script -> Local: GlobalValue
Script -> Script:
Script -> Global: GlobalValue

Assign local variable at script scope
Script -> Local: Foo
Script -> Script: Foo
Script -> Global: GlobalValue
Func -> Local: Foo
Func -> Script: Foo
Func -> Global: GlobalValue

Assign local variable at function scope
Func -> Local: Bar
Func -> Script: Foo
Func -> Global: GlobalValue

Assign script variable at function scope
Func -> Local: Bar
Func -> Script: Baz
Func -> Global: GlobalValue
Script -> Local: Baz
Script -> Script: Baz
Script -> Global: GlobalValue
```

servers/Fabric.Mcp.Server/src/Fabric.Mcp.Server.csproj
hallipr: It is.  We need to sync with the fabric team on their image name.  Their releases ~~will~~ would fail until we get that.

eng/scripts/Build-Code.ps1
hallipr: And, with powershell's parameter prefix matching, callers using "-OperatingSystem" will still work

eng/scripts/Build-Code.ps1
hallipr: We could, but it's expressed in paths and matched to legs in the build matrix

hallipr: nm... We only use the parameter as a filter for platforms in build_info.json and powershell will do a case insensitive match there

servers/Azure.Mcp.Server/src/Azure.Mcp.Server.csproj
hallipr: I copied this from the current [package.json](https://github.com/microsoft/mcp/blob/main/eng/vscode/package.json#L4):
```
 "description": "Provides Model Context Protocol (MCP) integration and tooling for Azure in Visual Studio Code.",
```

MS tends to put periods at the end of descriptions.  Others don't:
image
image

image

We're not 100% consistent:
image

eng/scripts/New-BuildInfo.ps1
hallipr: I'm pretty wary of defaulting a switch parameter:
https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoiddefaultvalueswitchparameter?view=ps-modules

What about a dual condition:

``` powershell
[switch] $CI

...

$isPipelineRun = $CI -or $env:TF_BUILD -eq 'true'
```

eng/scripts/Prepare-Docker.ps1
hallipr: Dummy value

hallipr: But, this should have been removed from the script.   I want all the build / pack scripts to fail if there's no build_info file, but with an error message that says how to get one

hallipr:
> Also we should support pipeline tagging in the same way we do for unified pull request builds in the language repos, so that users can filter by server name in the tags dropdown for the pipeline.

I'll add a followup PR

servers/Fabric.Mcp.Server/src/Fabric.Mcp.Server.csproj
weshaggard: Instead of having the release fail can we instead opt-out fabric from the docker release until a name is provided?

eng/pipelines/templates/jobs/docker/release-docker.yml
```diff
@@ -78,6 +37,36 @@ jobs:
...
+            exit 1
+        }
+
+        $imageName = $serverInfo.dockerImageName
+        Write-Host "Setting variable DockerImageName to $imageName"
+        Write-Host "##vso[task.setvariable variable=DockerImageName;isOutput=true]$imageName"
+
+        $version = $serverInfo.version
+        Write-Host "Setting variable DockerImageVersion to $version"
+        Write-Host "##vso[task.setvariable variable=DockerImageVersion;isOutput=true]$version"
```
weshaggard, line 60: nit: Do these need to be output variables now?

eng/scripts/Pack-Npm.ps1
```diff
@@ -4,135 +4,248 @@
...
 
+$ErrorActionPreference = "Stop"
 . "$PSScriptRoot/../common/scripts/common.ps1"
 $RepoRoot = $RepoRoot.Path.Replace('\', '/')
 
 $wrapperSourcePath = "$RepoRoot/eng/npm/wrapper"
 $platformSourcePath = "$RepoRoot/eng/npm/platform"
 
+# When running locally, ignore missing artifacts instead of failing
+$ignoreMissingArtifacts = $env:TF_BUILD -ne 'true'
```
weshaggard, line 20: nit: I think we have shared function for this in common.

eng/pipelines/templates/jobs/initialize.yml
```diff
@@ -23,30 +27,22 @@ jobs:
...
-      Write-Host "Setting version suffix to '$suffix'"
-      Write-Host "##vso[task.setvariable variable=VersionSuffix;isOutput=true]$suffix"
-    name: GetVersion
-    displayName: "Get version suffix"
-    env:
-      DYNAMIC: ${{ parameters.DynamicPrereleaseVersion }}
+  - task: Powershell@2
+    displayName: "Create build info file"
+    name: CreateBuildInfo
+    condition: and(succeeded(), ne(variables['NoPackagesChanged'],'true'))
```
weshaggard: Do we not need an empty build_info if NoPackagesChanged is set?

eng/pipelines/templates/jobs/sign-and-pack.yml
```diff
@@ -117,4 +125,23 @@ jobs:
...
+      - PackNpm
+      - PackNuget
+    SparseCheckout: false
+    JobTemplatePath: /eng/pipelines/templates/jobs/verify-packages.yml
+    AdditionalParameters:
+      TestTimeoutInMinutes: 10
+      ServerName: ${{ parameters.ServerName }}
+    MatrixConfigs:
+      - Name: build_matrix
+        Path: eng/pipelines/build-matrix.json
```
weshaggard: This matrix no longer exists. You will need to try to share the matrix of the build.

eng/scripts/Build-Code.ps1
```diff
@@ -3,136 +3,124 @@
...
-                'osx' { $node_os = 'darwin'; $extension = '' }
-                default { $node_os = $os; $extension = '' }
-            }
+    New-Item -Path $serverOutputDirectory -ItemType Directory -Force | Out-Null
 
-            $outputDir = "$serverOutputDirectory/$os-$arch"
-            Write-Host "Building version $version, $os-$arch in $outputDir" -ForegroundColor Green
+    foreach ($os in $OperatingSystem) {
+        foreach ($arch in $Architecture) {
+            $platform = $server.platforms
```
weshaggard: You might need to check that `$platform` is not empty or contains more than one entry.

eng/scripts/Build-Code.ps1
```diff
@@ -156,45 +144,22 @@ function BuildServer($serverName) {
...
 try {
-    $serverNames = @(if($ServerName) {
-        $ServerName
-    } else {
-        Get-ChildItem -Path "$RepoRoot/servers" -Directory | Select-Object -ExpandProperty Name
-    })
+    if ($CleanBuild) {
+        # Clean up any previous build artifacts.
+        Write-Host "Removing existing bin and obj folders"
+        Remove-Item * -Recurse -Include 'obj', 'bin' -Force -ProgressAction SilentlyContinue
```
weshaggard, line 169: Why clean this way? I would call "dotnet clean" or "dotnet rebuild"

eng/scripts/Build-Docker.ps1
```diff
@@ -3,43 +3,125 @@
...
+                "--build-arg PUBLISH_DIR=$(quote $relativePublishDirectory)"
+                "--build-arg CLI_NAME=$(quote $server.cliName + $platform.extension)"
+                "--file $(quote $dockerFile)"
+                "--tag $(quote $tag)"
+                "--no-cache"
+                "--progress plain"
+                "."
+            )
+
+            Invoke-LoggedCommand "docker build $($dockerArgs -join ' ')"
```
weshaggard, line 117: We should handle the error code on these calls.

eng/pipelines/templates/jobs/docker/release-docker.yml
hallipr: They do.  We need to pass them to the injected ACR step using variables

eng/scripts/New-BuildInfo.ps1
```diff
@@ -0,0 +1,429 @@
...
+    $parentDirectory = Split-Path $OutputPath -Parent
+    New-Item -Path $parentDirectory -ItemType Directory -Force | Out-Null
+
+    $buildInfo | ConvertTo-Json -Depth 5 | Out-File -FilePath $OutputPath -Encoding utf8 -Force
+
+    if ($isPipelineRun) {
+        Write-Host "##vso[task.setvariable variable=WindowsBuildMatrix;isOutput=true]$($buildMatrices.windows | ConvertTo-Json -Compress)"
+        Write-Host "##vso[task.setvariable variable=LinuxBuildMatrix;isOutput=true]$($buildMatrices.linux | ConvertTo-Json -Compress)"
+        Write-Host "##vso[task.setvariable variable=MacOsBuildMatrix;isOutput=true]$($buildMatrices.macos | ConvertTo-Json -Compress)"
+        Write-Host "##vso[task.setvariable variable=UnitTestMatrix;isOutput=true]$($unitTestMatrix | ConvertTo-Json -Compress)"
```
weshaggard: It doesn't appear we use the UnitTestMatrix anywhere yet as we just run on all the build legs.

Also, it appears as though the test matrices all assume one OS which might work for now but will likely need tweaking.

eng/scripts/Pack-Npm.ps1
hallipr: I wasn't able to find one that matches the purpose.  There's `Test-SupportsDevOpsLogging` in [`logging.ps1`](https://github.com/Microsoft/mcp/blob/main/eng/common/scripts/logging.ps1).

The other scripts in eng/common seem to reimplement it locally.

eng/scripts/Build-Code.ps1
hallipr: During a conversation on intermediate artifact reuse, @agocke mentioned that not all build tasks are great about only reusing intermediates for the same build parameters, i.e.  if you call msbuild again with different parameters, you should clean first as you can't rely on the tasks invalidating the appropriate cached intermediates.  He also said that it's best to just nuke the bin and obj folders rather than calling `dotnet clean`

hallipr: `Remove-Item * -Recurse -Include 'obj', 'bin'` is also really fast

eng/scripts/Build-Docker.ps1
hallipr: `Invoke-LoggedCommand` should error if the exit code isn't in $AllowedExitCodes (defaults to `@(0)`).  It's why I added the parameter.

hallipr: Aah... nm...  without $errorAction == stop, it continues after write-error.  I added $LASTEXITCODE handling to both calls

eng/pipelines/templates/jobs/vsix/release-vsix.yml
```diff
@@ -49,5 +49,5 @@ jobs:
                 $baseName = Get-ChildItem *.vsix | Select-Object -First 1 | ForEach-Object { $_.BaseName }
                 Write-Host "VSIX baseName: $baseName"
                 npm install -g @vscode/vsce
-                echo "vsce publish --azure-credential --packagePath ""$($baseName).vsix"" --manifestPath ""$($baseName).manifest"" --signaturePath ""$($baseName).signature.p7s"""
-                vsce publish --azure-credential --packagePath "$($baseName).vsix" --manifestPath "$($baseName).manifest" --signaturePath "$($baseName).signature.p7s"
+                echo "vsce publish --azure-credential --packagePath ""$baseName.vsix"" --manifestPath ""$baseName.manifest"" --signaturePath ""$baseName.signature.p7s"""
+                vsce publish --azure-credential --packagePath "$baseName.vsix" --manifestPath "$baseName.manifest" --signaturePath "$baseName.signature.p7s"
```
scbedd, line 53:
```suggestion
                vsce publish --azure-credential --packagePath "$($baseName.vsix)" --manifestPath "$($baseName.manifest)" --signaturePath "$($baseName.signature.p7s)"
```

These won't resolve in a pshell script as is right?

eng/pipelines/templates/jobs/vsix/release-vsix.yml
```diff
@@ -49,5 +49,5 @@ jobs:
                 $baseName = Get-ChildItem *.vsix | Select-Object -First 1 | ForEach-Object { $_.BaseName }
                 Write-Host "VSIX baseName: $baseName"
                 npm install -g @vscode/vsce
-                echo "vsce publish --azure-credential --packagePath ""$($baseName).vsix"" --manifestPath ""$($baseName).manifest"" --signaturePath ""$($baseName).signature.p7s"""
-                vsce publish --azure-credential --packagePath "$($baseName).vsix" --manifestPath "$($baseName).manifest" --signaturePath "$($baseName).signature.p7s"
+                echo "vsce publish --azure-credential --packagePath ""$baseName.vsix"" --manifestPath ""$baseName.manifest"" --signaturePath ""$baseName.signature.p7s"""
```
scbedd:
```suggestion
                echo "vsce publish --azure-credential --packagePath `"$($baseName.vsix)`" --manifestPath `"$($baseName.manifest)`" --signaturePath `"$($baseName.signature.p7s)`""
```

weshaggard (approved):
Changes look good but I don't want the release to fail the first time Fabric attempts to use it so please work with them to get naming figured out or opt them out of those stages if we don't have a value yet.

eng/pipelines/templates/jobs/vsix/release-vsix.yml
scbedd: Image

eng/pipelines/templates/jobs/sign-and-pack.yml
```diff
@@ -117,4 +125,19 @@ jobs:
 - template: /eng/pipelines/templates/jobs/vsix/pack-and-sign-vsix.yml
   parameters:
     DependsOn: Sign
-    ServerName: ${{ parameters.ServerName }}
+
+- template: /eng/pipelines/templates/jobs/docker/stage-docker.yml
+  parameters:
+    DependsOn: Sign
+
+- ${{ each os in split('linux|windows|macOS', '|') }}:
```
scbedd, line 133: Putting a pin in this. I believe that Dozie is handling the different name of the OS in the build matrix _in_ `verify-packages.yml`.

Will verify and come back to this.

eng/scripts/Build-Code.ps1
```diff
@@ -3,136 +3,136 @@
...
 
-if($AllPlatforms -and $BuildNative) {
-    Write-Warning "Native Builds do not support Cross OS builds. Only building for the current OS."
+if(!$BuildInfoPath) {
+    $BuildInfoPath = "$RepoRoot/.work/build_info.json"
+}
+
+if (!(Test-Path $BuildInfoPath)) {
+    LogError "Build info file $BuildInfoPath does not exist. Run eng/scripts/New-BuildInfo.ps1 to create it."
+    $exitCode = 1
```
scbedd, line 46: ~~We should probably immediately bail here right?~~

eng/scripts/Build-Code.ps1
```diff
@@ -156,45 +156,22 @@ function BuildServer($serverName) {
 
             Invoke-LoggedMsBuildCommand $command -GroupOutput
 
-            $package = [ordered]@{
```
scbedd, line 159: I am missing it in the large context of this PR. Where do we generate the package.json? I think you added it to the build of the csproj right?

eng/scripts/Build-Local.ps1
```diff
@@ -4,13 +4,13 @@
 [CmdletBinding(DefaultParameterSetName='none')]
 param(
     [string] $ServerName,
-    [switch] $Trimmed,
+    [switch] $NoTrimmed,
```
scbedd, line 7: Why change from `Trimmed` to `NoTrimmed`?

eng/scripts/Compress-ForSigning.ps1
```diff
@@ -21,12 +23,25 @@ if(!$ArtifactPrefix) {
...
+    $exitCode = 1
+}
+
+if(!(Test-Path -Path $BuildInfoPath -PathType Leaf)) {
+    LogError "Build info file '$BuildInfoPath' does not exist or is not a file."
+    $exitCode = 1
+}
+
+if ($exitCode -ne 0) {
+    exit $exitCode
```
scbedd, line 44: It's not worth it in this PR, but later on we should parse these common arguments (for BuildInfoJson at least) in a common verification function. LATER. We should merge this PR now 👍

scbedd (changes requested):
Overall definitely looks solid to me. I have some concerns around usage of member variables for which I left comments. After those resolve, will swap to `approve`. 

If this turns out solid livetests etc we should get this in ASAP.

eng/pipelines/templates/jobs/sign-and-pack.yml
hallipr: He's expecting a matrix file and the generate-matrix job.  We're creating all of the matrices at the start of pipeline instead of using a static matrix file now.

eng/scripts/Build-Code.ps1
hallipr: package.json as the primary metadata carrier has been replaced with build_info.json, which is constructed in New-Buil.dInfo.ps1, which is invoked in Initialize.xml.  Build, test and pack scripts should all work from the metadata in build_info.ps1.

For package.json (and other publish files):
- npm's package.json is constructed in Pack-Npm.ps1
- vsix's package.json is static and each server will probably have their own, but Pack-VSIX.ps1 will be responsible for locating it and updating it with the current build info.
- nuget's .nuspec file and several other nuget specific xml files are constructed in Pack-Nuget.ps1
- Dockerfile is still in the root of the repo and will need to be updated by Prepare-Docker.ps1 to accommodate the new server executable names

eng/scripts/Build-Local.ps1
hallipr: The CI output will be self-contained, trimmed and single-file, so I'm moving the defaults for Build-Local to match.

I could be easily swayed to the opposite argument that:
- Build-Local should be as fast as possible
- A local dev will always have the runtime installed, so self-contained is unnecessary and only slows packaging
- Trimming is incredibly slow and is largely unnecessary local except in the case of AOT warning checks
- Single File slows initial run time and should probably not be the local default

I'll follow with another PR to make Build-Local more local dev friendly

servers/Fabric.Mcp.Server/src/Fabric.Mcp.Server.csproj
hallipr: Docker is already opt-in and Fabric is explicitly opting out:

[servers/Fabric.Mcp.Server/build.yml](https://github.com/Microsoft/mcp/blob/main/servers/Fabric.Mcp.Server/build.yml)
```yaml
extends:
  template: /eng/pipelines/templates/common.yml
  parameters:
    ServerName: Fabric.Mcp.Server
    PublishTarget: ${{ parameters.PublishTarget }}
    RunLiveTests: true
    IncludeDockerRelease : false
    IncludeNative: false
```

I'll follow with a PR to make all of the packaging formats opt-in:
```yaml
    ReleaseDocker : false
    ReleaseNpm : true
    ReleaseNuget: true
    ReleaseVsix: false
```

PR #558, opened 2025-09-24
Remove setting value of `Activity.Current` in telemetry service by removing async StartActivity.

conniey:
## What does this PR do?

Currently we are explicitly setting `Activity.Current` due to the async flow of `ITelemetryService`.  This changes the `StartActivity` to synchronous.

References: https://github.com/dotnet/runtime/issues/30915#issuecomment-533757730

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:
    - [ ] Updated command list in `/docs/azmcp-commands.md` and/or `/docs/fabric-commands.md`
    - [ ] Updated test prompts in `/docs/e2eTestPrompts.md`

### DIFF
--- a/core/Azure.Mcp.Core/src/Services/Telemetry/ITelemetryService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Telemetry/ITelemetryService.cs
@@ -29,5 +29,5 @@ public interface ITelemetryService : IDisposable
     /// Performs any initialization operations before telemetry service is ready.
     /// </summary>
     /// <returns>A task that completes when initialization is complete.</returns>
-    ValueTask InitializeAsync();
+    Task InitializeAsync();
 }
--- a/core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
@@ -115,7 +115,7 @@ internal class TelemetryService : ITelemetryService
     /// <summary>
     /// <inheritdoc/>
     /// </summary>
-    public async ValueTask InitializeAsync()
+    public async Task InitializeAsync()
     {
         if (!_isEnabled)
         {
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/CommandFactoryHelpers.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/CommandFactoryHelpers.cs
@@ -174,6 +174,6 @@ internal class CommandFactoryHelpers
         {
         }
 
-        public ValueTask InitializeAsync() => ValueTask.CompletedTask;
+        public Task InitializeAsync() => Task.CompletedTask;
     }
 }
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Telemetry/TelemetryServiceTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Telemetry/TelemetryServiceTests.cs
@@ -226,7 +226,7 @@ public class TelemetryServiceTests
         // Act & Assert
         using var service = new TelemetryService(informationProvider, mockOptions, _mockServiceOptions, _logger);
 
-        await Assert.ThrowsAsync<ArgumentNullException>(() => service.InitializeAsync().AsTask());
+        await Assert.ThrowsAsync<ArgumentNullException>(() => service.InitializeAsync());
 
         Assert.Throws<InvalidOperationException>(() => service.StartActivity("an-activity-id", implementation));
     }


core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
```diff
@@ -31,20 +33,23 @@ public TelemetryService(IMachineInformationProvider informationProvider, IOption
...
         if (!_isEnabled)
         {
             return null;
         }
 
-        await _isInitialized.Task;
+        if (!_isInitialized)
+        {
+            throw new InvalidOperationException(
+                "Telemetry service has not been initialized. Use InitializeAsync() before any other operations.");
```
vukelich: Suggestion: use `$"Telemetry service has not been initialized. Use {nameof(InitializeAsync)}() before any other operations."` Since `nameof` references a symbol, in this case a method, a change or removal of the method name will cause a compilation error. This keeps the error message in-line with the code. #Closed

core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
```diff
@@ -70,21 +75,27 @@ public void Dispose()
     {
     }
 
-    private async Task InitializeTagList()
+    public async ValueTask InitializeAsync()
     {
+        var wasInitialized = Interlocked.CompareExchange(ref _isInitialized, true, false);
```
vukelich: Since this method is asynchronous and is setting required state, we should use a blocking technique to prevent race conditions where:

Caller 1: starts the initialization then starts working on the steps in this method
Caller 2: calls this but exits early seeing `wasInitialized` as already `true`. Then calls one of the `StartActivity` methods (1) before Caller 1's invocation adds entries into `_tagsList` causing the new Activity to miss tags or (2) during Caller 1's addition of entries into `_tagsList` which could cause enumerating of the tags in Caller 2 to throw `System.InvalidOperationException` with a message of `"Collection was modified; enumeration operation may not execute"`.

Async blocking is very easy with a `SemaphoreSlim`. #Closed

core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
```diff
@@ -70,21 +75,27 @@ public void Dispose()
...
 
             _tagsList.Add(new(TagName.MacAddressHash, macAddressHash));
             _tagsList.Add(new(TagName.DevDeviceId, deviceId));
-
-            _isInitialized.SetResult();
         }
         catch (Exception ex)
         {
-            _isInitialized.SetException(ex);
+            _logger.LogError(ex, "Error occurred initializing telemetry service.");
```
vukelich: If initialization fails, what do we want to happen? Should the process end? Should this be retried? Do we just leave the information missing for the rest of the lifetime of the application?

I'd recommend putting your design decision in comments in the code as well as in `ITelemetryService.InitializeAsync`'s doc comment. Behavioral expectations are great to document in the code. #Closed

vukelich: As of now in `Program.cs`, the process will stop during `Main`. This is probably the behavior we want to keep things simple, but it's still worth documenting.

core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
conniey: Thanks! I can't believe I overlooked that. 

[edit]: I can believe it. Lol. Never mind.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -108,6 +110,8 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
         try
         {
             using var host = CreateHost(options);
+
+            await InitializeServicesAsync(host.Services);
```
vukelich, line 129: Why is `InitializeServicesAsync` called in `ServiceStartCommand.cs` and in `Program.cs`? In `Program.cs`, we assign a value to `ServiceStartCommand.InitializeServicesAsync` as a reference to `Program.InitializeServicesAsync`, we directly call `Program.InitializeServicesAsync` within `Program.Main`, and `ServiceStartCommand.ExecuteAsync` will call `ServiceStartCommand.InitializeServicesAsync`.

Are these covering different possible code paths?

core/Azure.Mcp.Core/src/Services/Telemetry/ITelemetryService.cs
```diff
@@ -8,7 +8,26 @@ namespace Azure.Mcp.Core.Services.Telemetry;
 
 public interface ITelemetryService : IDisposable
 {
-    ValueTask<Activity?> StartActivity(string activityName);
+    /// <summary>
+    /// Creates and starts a new telemetry activity.
+    /// </summary>
+    /// <param name="activityName">Name of the activity.</param>
+    /// <returns>An Activity object or null if there are no active listeners or telemetry is disabled.</returns>
+    /// <exception cref="IllegalOperationException">If the service is not in an operational state or <see cref="InitializeAsync"/> was not invoked.</exception>
```
vukelich: Should be `InvalidOperationException`. There is no `IllegalOperationException` in .NET. #Closed

core/Azure.Mcp.Core/src/Services/Telemetry/ITelemetryService.cs
```diff
@@ -8,7 +8,26 @@ namespace Azure.Mcp.Core.Services.Telemetry;
...
+    Activity? StartActivity(string activityName);
 
-    ValueTask<Activity?> StartActivity(string activityName, Implementation? clientInfo);
+    /// <summary>
+    /// Creates and starts a new telemetry activity.
+    /// </summary>
+    /// <param name="activityName">Name of the activity.</param>
+    /// <param name="clientInfo">MCP client information to add to the activity.</param>
+    /// <returns>An Activity object or null if there are no active listeners or telemetry is disabled.</returns>
+    /// <exception cref="IllegalOperationException">If the service is not in an operational state or <see cref="InitializeAsync"/> was not invoked.</exception>
```
vukelich: Should be `InvalidOperationException`. There is no `IllegalOperationException` in .NET. #Closed

core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
```diff
@@ -31,20 +37,41 @@ public TelemetryService(IMachineInformationProvider informationProvider, IOption
...
         }
 
-        await _isInitialized.Task;
+        if (!_isInitialized)
+        {
+            throw new InvalidOperationException(
+                $"Telemetry service has not been initialized. Use {nameof(InitializeAsync)}() before any other operations.");
+        }
+
+        if (!_initializationCompleted.Wait(_operationTimeout))
```
vukelich: I strongly recommend not synchronously waiting on this. We want to design this type such that we require initialization to be completed before any invocation to `StartActivity`. Lean into the idea that the system is in a misused and invalid state if `InitializeAsync` did not complete successfully. #Closed

servers/Azure.Mcp.Server/CHANGELOG.md
```diff
@@ -182,6 +182,7 @@ The Azure MCP Server updates automatically by default whenever a new release com
   - `azmcp_storage_share_file_list`
   - `azmcp_storage_table_list`
 - Updated the `OpenWorld` and `Destructive` hints for all tools. [[#510](https://github.com/microsoft/mcp/pull/510)]
+- Fix flow of `Activity.Current` in telemetry service by changing `ITelemetryService`'s activity calls to synchronous. [[#558](https://github.com/microsoft/mcp/pull/558)]
```
alzimmermsft: Need to move this into the right CHANGELOG entry

core/Azure.Mcp.Core/src/Services/Telemetry/TelemetryService.cs
```diff
@@ -70,8 +97,18 @@ public void Dispose()
     {
     }
 
-    private async Task InitializeTagList()
+    /// <summary>
+    /// <inheritdoc/>
+    /// </summary>
+    public async ValueTask InitializeAsync()
```
vukelich: My suggestion for using a `SemaphoreSlim` would look like the following:

```csharp
private readonly SemaphoreSlim _initalizeLock = new(1);

/// <summary>
/// Task created on the first invocation of <see cref="InitializeAsync"/>.
/// This is saved so that repeated invocations will see the same exception
/// as the first invocation.
/// </summary>
private Task? _initalizationTask = null;

public async Task InitializeAsync(CancellationToken cancellationToken)
{
    // Quick check if initialization already happened. Avoids
    // trying to get the lock.
    if (_initalizationTask == null)
    {
        // Get async lock for starting initialization
        await _initalizeLock.WaitAsync();
        
        try
        {
            // Check after acquiring lock to ensure we honor work
            // started while we were waiting.
            if (_initalizationTask == null)
            {
                _initalizationTask = InnerInitializeAsync();
            }
        }
        finally
        {
            _initalizeLock.Release();
        }
    }

    // Await the response of the initialization work regardless of if
    // we or another invocation created the Task representing it. All
    // awaiting on this will give the same result to ensure idempotency.
    await _initalizationTask;
    
    async Task InnerInitializeAsync()
    {
        try
        {            
            var macAddressHash = await _informationProvider.GetMacAddressHash();
            var deviceId = await _informationProvider.GetOrCreateDeviceId();

            _tagsList.Add(new(TagName.MacAddressHash, macAddressHash));
            _tagsList.Add(new(TagName.DevDeviceId, deviceId));

            _initializationSuccessful = true;
        }
        catch (Exception ex)
        {
            _logger.LogError(ex, "Error occurred initializing telemetry service.");
            throw;
        }
        finally
        {
            _isInitialized = true;
        }
    }
}
``` #Closed

File: core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs (hunk 4 above)
Comment (conniey) [reply core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs:129]: The original design has this as a CLI with some commands. Then they decided that we should have a CLI command that starts the MCP server with access to the same set of commands. So it covers both those paths

Comment (vukelich) [reply core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs:129]: I still don't follow. `Program.Main` will run before any commands are parsed.

Comment (vukelich) [reply core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs:129]: Let's chat offline. I don't want my confusion to block PR submission.

File: core/Azure.Mcp.Core/src/Services/Telemetry/ITelemetryService.cs [hunk 10]
```diff
@@ -8,7 +8,26 @@ namespace Azure.Mcp.Core.Services.Telemetry;
...
+    /// <param name="clientInfo">MCP client information to add to the activity.</param>
+    /// <returns>An Activity object or null if there are no active listeners or telemetry is disabled.</returns>
+    /// <exception cref="InvalidOperationException">If the service is not in an operational state or <see cref="InitializeAsync"/> was not invoked.</exception>
+    Activity? StartActivity(string activityName, Implementation? clientInfo);
+
+    /// <summary>
+    /// Performs any initialization operations before telemetry service is ready.
+    /// </summary>
+    /// <returns>A task that completes when initialization is complete.</returns>
+    ValueTask InitializeAsync();
```
vukelich: My sample for rewriting `TelemetryService.InitializeAsync` intentionally changed the signature to return `Task` rather than `ValueTask`. The former can be awaited on any number of times, but `ValueTask` should only be awaited on once, ever. That's where my rule of thumb knowledge stops, so if you know `ValueTask` is the right thing here, keep it. If not, I'd recommend using a regular `Task`. We expect this method to be called once or a couple times at most, so I don't see a need for `ValueTask` optimization. Let me know the value you see.

conniey: Mm, my understanding of ValueTask is that if the result is if you were expecting to await the instance of that class, the result is usually synchronously available. (I figured since InstantiateAsync is usually performed once and then its done that it would be easier to just return a ValueTask.)

PR #576, opened 2025-09-25
Update to ModelContextProtocol 0.4.0-preview.1

stephentoub:
Updates to use ModelContextProtocol 0.4.0-preview.1


core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -58,16 +58,16 @@ public async Task<IMcpClient> CreateClientAsync(McpClientOptions clientOptions)
     /// </summary>
     /// <param name="clientOptions">Options to configure the client behavior.</param>
     /// <returns>A configured MCP client using SSE transport.</returns>
-    private async Task<IMcpClient> CreateSseClientAsync(McpClientOptions clientOptions)
+    private async Task<McpClient> CreateHttpClientAsync(McpClientOptions clientOptions)
```
alzimmermsft, line 61: Could you update the docs to stop referencing SSE as this no longer uses that transport

stephentoub: It's the same transport, the type is just renamed. The transport has long been the streamable HTTP transport, which is backwards compatible with the now defunct "SSE transport".

Which docs need updating?

alzimmermsft: Ah, okay, didn't realize that. Was more thinking the docs on the private method reference SSE, and seeing SSE -> HTTP made me think the behavior changed here.

joshfree:
image

If you follow the instructions here https://aka.ms/azmcp/intake then live tests will auto-run for you

stephentoub:
@jongio, could we get this merged as soon as you're ready for it to be? I'm having to repeatedly merge and fix conflicts as other PRs get merged. Thanks.

joshfree:
/azp run mcp - pullrequest - live

Running live tests on behalf of stoub

Issue #624, opened 2025-09-30
Tool Description: Enhance tool description and Reduce Ambiguity for better LLM Selection


anannya03:
Enhance the tool description of the following to increase the confidence score to >0.4 and/or reduce the ambiguity from the major tools(cosmos, messaging, storage).

- [x] Resource Health (done- https://github.com/microsoft/mcp/pull/588)
- [x] Foundry (done- https://github.com/microsoft/mcp/pull/599)
- [x] Subscription (done- https://github.com/microsoft/mcp/pull/559)
- [x] Best Practices (closed because no change needed as per the comments on the PR- https://github.com/microsoft/mcp/pull/603)
- [x] Event grid (done- https://github.com/microsoft/mcp/pull/552)
- [x] CloudArchitect (done- https://github.com/microsoft/mcp/pull/556)
- [x] App Lens (done- https://github.com/microsoft/mcp/pull/587)
- [x] Cosmos (done - https://github.com/microsoft/mcp/pull/625)
- [x] Deploy (done - https://github.com/microsoft/mcp/pull/640)
- [x] Monitor (in progress - https://github.com/microsoft/mcp/pull/630)
- [x] Messaging (done - https://github.com/microsoft/mcp/pull/642)
- [x] Storage (done- https://github.com/microsoft/mcp/pull/650)

PR #650, opened 2025-10-01
Enhance tool description to reduce ambiguity in Storage tools

anannya03:
## What does this PR do?
Enhance tool description to reduce ambiguity in Storage tools

**Results for the Storage tools:**
[storage-results.md](https://github.com/user-attachments/files/22634545/storage-results.md)

## GitHub issue number?
https://github.com/microsoft/mcp/issues/624


tools/Azure.Mcp.Tools.Storage/src/Commands/Blob/Container/ContainerCreateCommand.cs
```diff
@@ -18,7 +18,7 @@ public sealed class ContainerCreateCommand(ILogger<ContainerCreateCommand> logge
 
     public override string Description =>
         """
-        Creates an Azure Storage container, returning the last modified time, the ETag of the created container, and more.
+        Create/provision a new Azure Storage blob container in a storage account. Required: --account <account>, --container <container>, --subscription <subscription>. Optional: --public-access-level, --tenant <tenant>. Returns: container name, lastModified, eTag, leaseStatus, publicAccessLevel, hasImmutabilityPolicy, hasLegalHold. Creates a logical container for organizing blobs within a storage account.
```
JasonYeMSFT, line 21: Do you need to repeat the information for the input? The input schema should already have description for each argument and info on which ones are required.

anannya03: Yes, we're adding the parameter details on purpose. Even though the confidence score of the storage tools was already good, they were ambiguous (the confidence score difference between the first tool and second tool in our results.md was <0.3). At times this can lead to the LLM getting confused and pick the wrong tool. We've seen that adding the parameter details helps make the description more unambiguous and guides the LLM to choose the correct tool.

Issue #757, opened 2025-10-09
Improvement: Flatten azmcp_redis_cache_accesspolicy_list and azmcp_redis_cluster_database_list into azmcp_redis_list


sharedferret:
In order to streamline the data returned for Redis resources, the data from `azmcp_redis_cache_accesspolicy_list` and `azmcp_redis_cluster_database_list` should be returned as part of the Redis cache resource details returned from `azmcp_redis_list`.

Issue #756, opened 2025-10-09
Improvement: Combine azmcp_redis_cache_list and azmcp_redis_cluster_list into a single command


sharedferret:
Currently, there are two separate commands for listing Redis cache resources in a user subscription - one for Azure Cache for Redis resources, and another for Azure Managed Redis and Azure Redis Enterprise clusters. To improve usability, these two commands should be combined into a single `azmcp_redis_list` command that returns all of the customer's Redis resources at the same time.

PR #760, opened 2025-10-10
Add `mcp_redis_list` to return details on Redis cache resources

sharedferret:
## What does this PR do?
This PR replaces the `mcp redis cache list` and `mcp redis cluster list` commands with a unified `mcp redis list` command to return data on all Redis resources in the subscription. Previously, the `mcp redis cache list` returned data on Azure Cache for Redis and Azure Redis Enterprise resources, and `mcp redis cluster list` returned data on Azure Managed Redis resources. This change updates the list command to return AMR, Enterprise, and ACR resources in a single command.

This change additionally flattens the responses from `mcp redis cache accesspolicy list` and `azmcp redis cluster database list` into `mcp redis list`, so that the main list command returns all details for the Redis resources in a user's subscription as a single command.

ToolDescriptionEvaluator scores for the unified command are as follows:
Test 288: 0.810504 (1st)
Test 289: 0.685128 (1st)
Test 290: 0.781228 (1st)
Test 291: 0.572767 (1st)
Test 292: 0.478070 (1st)

## GitHub issue number?
https://github.com/microsoft/mcp/issues/756
https://github.com/microsoft/mcp/issues/757

## Pre-merge Checklist
- [x] Required for All PRs
- [x] For MCP tool changes:
    - [x] **One tool per PR**: This PR adds or modifies only one MCP tool for faster review cycles
    - [x] Updated `servers/Azure.Mcp.Server/README.md` and/or `servers/Fabric.Mcp.Server/README.md` documentation
    - [x] Updated command list in `/servers/Azure.Mcp.Server/docs/azmcp-commands.md` and/or `/docs/fabric-commands.md`
    - [x] For new or modified tool descriptions, ran [`ToolDescriptionEvaluator`](https://github.com/microsoft/mcp/blob/main/eng/tools/ToolDescriptionEvaluator/Quickstart.md) and obtained a score of `0.4` or more and a top 3 ranking for all related test prompts
    - [x] For new tools associated with Azure services or publicly available tools/APIs/products, add URL to documentation in the PR description
- [x] Extra steps for **Azure MCP Server** tool changes:
    - [x] Updated test prompts in `/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md`


sharedferret:
@microsoft-github-policy-service agree company="Microsoft"

tools/Azure.Mcp.Tools.Redis/src/Commands/ResourceListCommand.cs
```diff
@@ -49,23 +49,23 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
         try
         {
             var redisService = context.GetService<IRedisService>() ?? throw new InvalidOperationException("Redis service is not available.");
-            var caches = await redisService.ListCachesAsync(
+            var caches = await redisService.ListResourcesAsync(
```
philon-msft:
```suggestion
            var resources = await redisService.ListResourcesAsync(
```

tools/Azure.Mcp.Tools.Redis/src/Commands/ResourceListCommand.cs
```diff
@@ -49,23 +49,23 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
...
                 options.Subscription!,
                 options.Tenant,
                 options.AuthMethod,
                 options.RetryPolicy);
 
-            context.Response.Results = ResponseResult.Create(new(caches ?? []), RedisJsonContext.Default.CacheListCommandResult);
+            context.Response.Results = ResponseResult.Create(new(caches ?? []), RedisJsonContext.Default.ResourceListCommandResult);
```
philon-msft:
```suggestion
            context.Response.Results = ResponseResult.Create(new(resources ?? []), RedisJsonContext.Default.ResourceListCommandResult);
```

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,109 +30,38 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
     }
 
-    public async Task<IEnumerable<AccessPolicyAssignment>> ListAccessPolicyAssignmentsAsync(
-        string cacheName,
-        string resourceGroupName,
-        string subscription,
-        string? tenant = null,
-        AuthMethod? authMethod = null,
-        RetryPolicyOptions? retryPolicy = null)
+    private async Task<List<Resource>> ListAcrResourcesAsync(SubscriptionResource subscriptionResource)
```
philon-msft:
```suggestion
    private async Task<IEnumerable<Resource>> ListAcrResourcesAsync(SubscriptionResource subscriptionResource)
```

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -151,97 +80,89 @@ public async Task<IEnumerable<AccessPolicyAssignment>> ListAccessPolicyAssignmen
...
     }
 
-    public async Task<IEnumerable<Database>> ListDatabasesAsync(
-        string clusterName,
-        string resourceGroupName,
-        string subscription,
-        string? tenant = null,
-        AuthMethod? authMethod = null,
-        RetryPolicyOptions? retryPolicy = null)
+    private async Task<List<Resource>> ListAmrResourcesAsync(SubscriptionResource subscriptionResource)
```
philon-msft:
```suggestion
    private async Task<IEnumerable<Resource>> ListAmrResourcesAsync(SubscriptionResource subscriptionResource)
```

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,109 +30,38 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
-                        PreferredDataArchiveAuthMethod = cache.RedisConfiguration.PreferredDataArchiveAuthMethod,
-                        PreferredDataPersistenceAuthMethod = cache.RedisConfiguration.PreferredDataPersistenceAuthMethod,
-                        ZonalConfiguration = cache.RedisConfiguration.ZonalConfiguration,
-                        StorageSubscriptionId = cache.RedisConfiguration.StorageSubscriptionId,
-                        IsEntraIDAuthEnabled = string.IsNullOrWhiteSpace(cache.RedisConfiguration.IsAadEnabled) ? null : StringComparer.OrdinalIgnoreCase.Equals(cache.RedisConfiguration.IsAadEnabled, "True"),
-                    }
-                });
-            }
+            var resources = new List<Resource>();
+            resources.AddRange(await acrResourcesTask);
```
philon-msft: The `WhenAll()` and these 2 lines that await resources should each be wrapped in a try/catch that ignores exceptions, so if (for example) ACR fails, we still get AMR resources

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,109 +30,38 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
-            var accessPolicyAssignmentCollection = cacheResponse.Value.GetRedisCacheAccessPolicyAssignments();
+            if (string.IsNullOrWhiteSpace(acrResource?.Id.ToString())
+                || string.IsNullOrWhiteSpace(acrResource.Data.Name))
+            {
+                continue;
+            }
+
+            var resource = acrResource.Data;
+
+            var accessPolicyAssignmentCollection = acrResource.GetRedisCacheAccessPolicyAssignments();
```
philon-msft: These calls to get details about individual resources should all be wrapped in a try/catch that ignores exceptions so (for example) if we fail to get access policy assignments for one resource, we keep looping and trying to get the assignments for other resources

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -151,97 +80,89 @@ public async Task<IEnumerable<AccessPolicyAssignment>> ListAccessPolicyAssignmen
...
-            var databaseCollection = clusterResponse.Value.GetRedisEnterpriseDatabases();
+            if (string.IsNullOrWhiteSpace(amrResource?.Id.ToString())
+                || string.IsNullOrWhiteSpace(amrResource.Data.Name))
+            {
+                continue;
+            }
+
+            var resource = amrResource.Data;
+
+            var databaseCollection = amrResource.GetRedisEnterpriseDatabases();
```
philon-msft: Similar to retrieving access policy assignments - be sure to wrap this in a try/catch

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,257 +30,226 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
-                });
+            try
+            {
+                acrResourcesTask = ListAcrResourcesAsync(subscriptionResource);
+                amrResourcesTask = ListAmrResourcesAsync(subscriptionResource);
+                await Task.WhenAll(acrResourcesTask, amrResourcesTask);
             }
+            catch (Exception)
+            {
+                acrResourcesTask = Task.FromResult(Enumerable.Empty<Resource>());
```
philon-msft: We don't want to abandon both tasks - the idea is to carry on and return resources from the one that didn't throw. I think you want something like this: 

```cs
try
{
    var resourcesTasks = new List<Task<IEnumerable<Resource>>>();
    resourcesTasks.Add(ListAmrResourcesAsync(subscriptionResource));
    resourcesTasks.Add(ListAcrResourcesAsync(subscriptionResource));
    await Task.WhenAll(resourcesTasks);
}
catch { }
finally
{
    var resources = new List<Resource>();
    foreach (var resourcesTask in resourcesTasks)
    {
        if (resourcesTask.Status == TaskStatus.RanToCompletion)
        {
            resources.AddRange(await acrResourcesTask);
        }
    }
    return resources;
}
```

joshfree:
@sharedferret can you get this merged very early Monday AM?  If not, let's hold off on changes until after the October stabilization period.

sharedferret:
I should be able to merge it tonight or early tomorrow.On Oct 12, 2025, at 11:20, Josh Free ***@***.***> wrote:﻿joshfree left a comment (microsoft/mcp#760)
@sharedferret can you get this merged very early Monday AM?  If not, let's hold off on changes until after the October stabilization period.

servers/Azure.Mcp.Server/CHANGELOG.md
```diff
@@ -24,7 +24,9 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Breaking Changes
 
-- Unified required parameter validation: null or empty values now always throw `ArgumentException` with an improved message listing all invalid parameters. Previously this would throw either `ArgumentNullException` or `ArgumentException` only for the first invalid value. [[#718](https://github.com/microsoft/mcp/pull/718)]
+- Unified required parameter validation: null or empty values now always throw `ArgumentException` with an improved message listing all invalid parameters. Previously this would throw either `ArgumentNullException` or `ArgumentException` for only the first invalid value. [[#718](https://github.com/microsoft/mcp/pull/718)]
+- Replaced `azmcp_redis_cache_list` and `azmcp_redis_cluster_list` with a unified `azmcp_redis_list` command that lists all Redis resources in a subscription. [[#756](https://github.com/microsoft/mcp/issues/756)]
+- Flattened `azmcp_redis_cache_accesspolicy_list` and `azmcp_redis_cluster_database_list` into `azmcp_redis_list`. [[#757](https://github.com/microsoft/mcp/issues/757)]
```
alzimmermsft, line 29: This could be merged into on CHANGELOG entry as all four commands are being merged into the same one.

tools/Azure.Mcp.Tools.Redis/tests/Azure.Mcp.Tools.Redis.UnitTests/ResourceListCommandTests.cs
```diff
@@ -0,0 +1,276 @@
...
+using Azure.Mcp.Tools.Redis.Commands;
+using Azure.Mcp.Tools.Redis.Models.CacheForRedis;
+using Azure.Mcp.Tools.Redis.Models.ManagedRedis;
+using Azure.Mcp.Tools.Redis.Services;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using NSubstitute.ExceptionExtensions;
+using Xunit;
+using CacheModel = Azure.Mcp.Tools.Redis.Models.Resource;
```
alzimmermsft, line 18: If we're doing this in testing, does this indicate that `Resource` is a bad name and should be replaced with something better.

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,257 +30,229 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
-                    IsAccessKeyAuthenticationDisabled = cache.IsAccessKeyAuthenticationDisabled,
-                    LinkedServers = cache.LinkedServers.Any() ?
-                        [.. cache.LinkedServers.Select(server => server.Id.ToString())]
-                        : null,
-                    MinimumTlsVersion = cache.MinimumTlsVersion.ToString(),
-                    PrivateEndpointConnections = cache.PrivateEndpointConnections.Any() ?
-                        [.. cache.PrivateEndpointConnections.Select(connection => connection.Id.ToString())]
-                        : null,
-                    Identity = cache.Identity is null ? null : new ManagedIdentityInfo
+                    if (resourceTask.Status == TaskStatus.RanToCompletion)
```
alzimmermsft, line 49: Not a big fan of ignoring failed tasks as this could mask failures in data retrieval, leading for consumers to believe they worked on all resources when they actually didn't.

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,257 +30,229 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
-                caches.Add(new()
+            try
+            {
+                resourcesTasks.Add(ListAcrResourcesAsync(subscriptionResource));
+                resourcesTasks.Add(ListAmrResourcesAsync(subscriptionResource));
+                await Task.WhenAll(resourcesTasks);
+            }
+            catch (Exception)
+            { }
+            finally
```
alzimmermsft, line 45: This doesn't feel right to me where the consumption of the awaited tasks happens in a `finally` block.

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,257 +30,229 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
+                    var accessPolicyAssignment = accessPolicyAssignmentResource.Data;
+                    accessPolicyAssignments.Add(new()
+                    {
+                        AccessPolicyName = accessPolicyAssignment.AccessPolicyName,
+                        IdentityName = accessPolicyAssignment.ObjectIdAlias,
+                        ProvisioningState = accessPolicyAssignment.ProvisioningState?.ToString(),
+                    });
                 }
+            }
+            catch (Exception) { }
```
alzimmermsft, line 98: We're swallowing a lot of potential errors throughout this tool, is that what we really want to be doing?

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
```diff
@@ -29,257 +30,229 @@ public async Task<IEnumerable<Cache>> ListCachesAsync(
...
     }
 
-    public async Task<IEnumerable<Database>> ListDatabasesAsync(
-        string clusterName,
-        string resourceGroupName,
-        string subscription,
-        string? tenant = null,
-        AuthMethod? authMethod = null,
-        RetryPolicyOptions? retryPolicy = null)
+    private async Task<IEnumerable<Resource>> ListAmrResourcesAsync(SubscriptionResource subscriptionResource)
```
alzimmermsft, line 168: For my own knowledge, this is actually the acronym `Amr` and not a typo on `Arm`, right?

sharedferret: Yes, Amr in this case is referring to Azure Managed Redis.

tools/Azure.Mcp.Tools.Redis/tests/Azure.Mcp.Tools.Redis.UnitTests/ResourceListCommandTests.cs
sharedferret: I'll update this reference. The use of `Resource` as a name for Redis is to reduce confusion that could exist between Caches, Clusters, and Resources, which have all been used in the past to refer to different types of Azure Redis resources.

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
sharedferret: I had a conversation with @philon-msft about this last week, since I had a similar concern. The reasoning behind silently dropping most exceptions that occur after fetching the subscription is so that the command is able to return as much information as it's able to in the case of a downstream outage. For this command, it's better to return a partial result instead of no result, especially since this command is now making four different service calls to fetch data.

We could also potentially collect any exceptions we see during the execution of the command and return them alongside the data, but I'm not sure how useful that would end up being.

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
philon-msft: This should be a very rare failure mode, but the idea is that we don't want to block customers from working with their AMR Redis resources just because the ACR Redis control plane is down (and vice versa). Is there a good pattern we can follow to report a partial failure like that while keeping people unblocked?

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
philon-msft: Definitely a little unusual, but that was the best way I could think of to salvage the succeeded tasks while ignoring failed ones. Certainly open to suggestions though

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
philon-msft: Yes, we're making a best effort to return as much data as we can and gracefully omit what we can't retrieve. Is there a pattern should follow to be more clear? Like an optional `RetrievalError` field in the models so the LLM sees retrieval errors instead of missing data? I wasn't super confident that would be interpreted correctly by the LLM, unless there's a standardized approach for MCP

tools/Azure.Mcp.Tools.Redis/src/Services/RedisService.cs
sharedferret: For now, I'll change this so that we handle (or swallow) errors in the tasks themselves, so that we aren't relying on a `finally` to compile results.

philon-msft: We could eliminate the `finally` and just have the `foreach` after the `try/catch` if that looks better?

PR #801, opened 2025-10-14
Add Azure MCP Bug Bash report issue template

g2vinay:
This template allows users to report bugs found during the Azure MCP Bug Bash, including sections for bug description, expected behavior, actual behavior, reproduction steps, and environment details.

### DIFF
--- a/.github/workflows/auto-milestone-bugbash.yml
+++ b/.github/workflows/auto-milestone-bugbash.yml
@@ -1,0 +1,50 @@
+name: "Auto-assign milestone for Bug Bash template issues"
+
+on:
+  issues:
+    types: [opened]
+
+permissions:
+  issues: write
+  contents: read
+
+env:
+  TARGET_MILESTONE: "2025-10"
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-milestone-bugbash.yml
```diff
@@ -0,0 +1,50 @@
+name: "Auto-assign milestone for Bug Bash template issues"
```
joshfree, line 1: nice, but can you just set the milestone in the template above?

g2vinay: not sure, I don't think milestone property is officially supported. 
We can try it.

g2vinay: nope won't work, milestone property isn't official supported, per GH.
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

PR #835, opened 2025-10-16
Remove Azure SDK package source from nuget.config

weshaggard:
Removed Azure SDK package source from configuration.

Testing to see if this feed is needed for this repo.


joshfree:
I think we used this during the AOT work, but it's probably not needed any longer

PR #904, opened 2025-10-21
Fix Foundry Serialization issue.

### LINKED ISSUE #621: AI Foundry agent connection error and evaluation tool not invoked for relevant prompts
**Description:**
When testing the prompts `Query an agent in my AI foundry project` and `Evaluate the full query and response I got from my agent for task_adherence`, different errors occurred. 
- For the `Query an agent in my AI foundry project`  prompt, connecting to the agent resulted in an internal error related to backend configuration or serialization (missing JsonTypeInfo metadata). 
Image

- For the `Evaluate the full query and response I got from my agent for task_adherence` prompt, the expected tool `azmcp_foundry_agents_evaluate` was not triggered at all. 
Image
Image

**Steps to Reproduce:**
1. Start mcp server in namespaces mode.
2. Open GitHub Copilot in VS Code and [switch to Agent mode](https://code.visualstudio.com/docs/copilot/chat/chat-agent-mode).
3. Click Refresh on the tools list, and select the `Built-in`, `Extension: Github Copilot for Azure`, `MCP Server: Microsoft Docs` and `MCP Server: Azure MCP`.
4. Enter test prompts `Query an agent in my AI foundry project` and `Evaluate the full query and response I got from my agent for task_adherence`

**Environment:**

- OS: Windows.
- Vscode Version: 1.104.2
- GitHub Copilot Chat Version: 0.31.3
- Large Language Model: GPT-4.1
- Virtual Tool: Off
- Azure MCP Server: 0.8.2

**Expected Behavior:**
Both prompts should trigger the appropriate tools and return the expected results without errors.

@joshfree for notification

g2vinay:
Fixes #621 foundry serialization issue.

### DIFF
--- a/tools/Azure.Mcp.Tools.Foundry/src/Commands/FoundryJsonContext.cs
+++ b/tools/Azure.Mcp.Tools.Foundry/src/Commands/FoundryJsonContext.cs
@@ -59,6 +59,5 @@ namespace Azure.Mcp.Tools.Foundry.Commands;
 [JsonSerializable(typeof(CognitiveServicesAccountDeploymentData))]
 [JsonSerializable(typeof(FoundryService.ToolDefinitionAIFunction))]
 [JsonSerializable(typeof(List<FoundryService.ToolDefinitionAIFunction>))]
-[JsonSerializable(typeof(Dictionary<string, object>))]
 [JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase, DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingDefault)]
 internal sealed partial class FoundryJsonContext : JsonSerializerContext;


tools/Azure.Mcp.Tools.Foundry/src/Commands/FoundryJsonContext.cs
```diff
@@ -56,5 +56,18 @@ namespace Azure.Mcp.Tools.Foundry.Commands;
 [JsonSerializable(typeof(AgentsQueryAndEvaluateResult))]
 [JsonSerializable(typeof(AgentsEvaluateResult))]
 [JsonSerializable(typeof(CognitiveServicesAccountDeploymentData))]
+[JsonSerializable(typeof(Dictionary<string, object>))]
+[JsonSerializable(typeof(Dictionary<string, string>))]
```
anuchandy: my understanding is that using `[JsonSerializable(typeof(...<object>))]` (i.e. object)  is not ideal for trimming, Each value is object, so at runtime the serializer needs to discover the actual type of every value. That adds overhead and impact many source-gen optimizations.

I remember replacing object with the more trim-safe `JsonElement` in some services. Does that approach works  here? 

Tagging @agocke if he has some thoughts.

anuchandy: I believe the reason for the serialization error is that the code is trying to obtain type information for our local type `ToolDefinitionAIFunction` from the JSON context of an external type (`Microsoft.Extensions.AI.AIJsonUtilities.DefaultOptions.GetTypeInfo`), which has no knowledge of it. Ideally, the fix would be to define our own JSON context for `ToolDefinitionAIFunction`, like we do for other models.

PR #910, opened 2025-10-21
Add remote HTTP server mode with OAuth authentication and downstream token acquisition.

vukelich:
## What does this PR do?

Add remote HTTP server mode with OAuth authentication and downstream token acquisition.

- Add RunAsRemoteHttpService and OutgoingAuthStrategy options to enable MCP server to run as a remote HTTP service
- Introduce ITokenProvider abstraction for dependency injection of downstream authentication tokens and credentials
- Add support for On-Behalf-Of (OBO) token flow and hosting environment identity modes for outgoing authentication
- Implement OAuth Protected Resource Metadata endpoint at /.well-known/oauth-protected-resource with WWW-Authenticate challenge
- Add Visual Studio launch profile for debugging remote MCP server with Microsoft.Identity.Web configuration
- Modernize HTTP host creation using WebApplicationBuilder with authentication and authorization middleware

This is for our goals of allowing Microsoft customers self-hosting remote MCP servers.

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
    - [X] **Read [contribution guidelines](https://github.com/microsoft/mcp/blob/main/CONTRIBUTING.md)**
    - [X] PR title clearly describes the change
    - [X] Commit history is clean with descriptive messages ([cleanup guide](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md))
- [ ] For MCP tool changes:

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -68,6 +68,9 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="9.0.9" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="9.0.9" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="9.0.9" />
+    <PackageVersion Include="Microsoft.Identity.Abstractions" Version="9.5.0" />
+    <PackageVersion Include="Microsoft.Identity.Web" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.Identity.Web.Azure" Version="4.0.1" />
     <PackageVersion Include="Microsoft.HybridRow" Version="1.1.0-preview3" />
     <PackageVersion Include="Microsoft.Azure.Cosmos.Aot" Version="0.1.4-preview.2" />
     <PackageVersion Include="Microsoft.Azure.Mcp.AzTypes.Internal.Compact" Version="0.2.802" />
@@ -78,7 +81,7 @@
     <PackageVersion Include="YamlDotNet" Version="16.3.0" />
     <PackageVersion Include="System.CommandLine" Version="2.0.0-rc.1.25451.107" />
     <PackageVersion Include="System.Formats.Asn1" Version="9.0.9" />
-    <PackageVersion Include="System.IdentityModel.Tokens.Jwt" Version="8.11.0" />
+    <PackageVersion Include="System.IdentityModel.Tokens.Jwt" Version="8.14.0" />
     <PackageVersion Include="System.Linq.AsyncEnumerable" Version="10.0.0-rc.1.25451.107" />
     <PackageVersion Include="System.Net.ServerSentEvents" Version="10.0.0-rc.1.25451.107" />
     <PackageVersion Include="System.Numerics.Tensors" Version="9.0.0" />
--- a/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
+++ b/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
@@ -64,14 +64,7 @@ public static class AzureMcpServiceCollectionExtensions
 
         // Register tool loader strategies
         services.AddSingleton<CommandFactoryToolLoader>();
-        services.AddSingleton(sp =>
-        {
-            return new RegistryToolLoader(
-                sp.GetRequiredService<RegistryDiscoveryStrategy>(),
-                sp.GetRequiredService<IOptions<ToolLoaderOptions>>(),
-                sp.GetRequiredService<ILogger<RegistryToolLoader>>()
-            );
-        });
+        services.AddSingleton<RegistryToolLoader>();
 
         services.AddSingleton<SingleProxyToolLoader>();
         services.AddSingleton<CompositeToolLoader>();
@@ -216,7 +209,7 @@ public static class AzureMcpServiceCollectionExtensions
 
         var mcpServerBuilder = services.AddMcpServer();
 
-        if (serviceStartOptions.EnableInsecureTransports)
+        if (serviceStartOptions.EnableInsecureTransports || serviceStartOptions.RunAsRemoteHttpService)
         {
             mcpServerBuilder.WithHttpTransport();
         }
--- a/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
+++ b/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
@@ -3,15 +3,21 @@
 
 using System.CommandLine.Parsing;
 using System.Net;
+using Azure.Mcp.Core.Areas.Server.Models;
 using Azure.Mcp.Core.Areas.Server.Options;
 using Azure.Mcp.Core.Commands;
 using Azure.Mcp.Core.Helpers;
+using Azure.Mcp.Core.Services.Azure.Authentication;
 using Azure.Mcp.Core.Services.Telemetry;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Identity.Web;
 using OpenTelemetry.Logs;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
@@ -69,6 +75,8 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
         command.Options.Add(ServiceOptionDefinitions.Debug);
         command.Options.Add(ServiceOptionDefinitions.EnableInsecureTransports);
         command.Options.Add(ServiceOptionDefinitions.InsecureDisableElicitation);
+        command.Options.Add(ServiceOptionDefinitions.RunAsRemoteHttpService);
+        command.Options.Add(ServiceOptionDefinitions.OutgoingAuthStrategy);
         command.Validators.Add(commandResult =>
         {
             ValidateMode(commandResult.GetValueOrDefault(ServiceOptionDefinitions.Mode), commandResult);
@@ -77,6 +85,10 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
             ValidateNamespaceAndToolMutualExclusion(
                 commandResult.GetValueOrDefault<string[]?>(ServiceOptionDefinitions.Namespace.Name),
                 commandResult.GetValueOrDefault<string[]?>(ServiceOptionDefinitions.Tool.Name),
+                commandResult);
+            ValidateOutgoingAuthStrategy(
+                commandResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.RunAsRemoteHttpService.Name),
+                commandResult.GetValueOrDefault<OutgoingAuthStrategy>(ServiceOptionDefinitions.OutgoingAuthStrategy.Name),
                 commandResult);
         });
     }
@@ -97,6 +109,17 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
             mode = ModeTypes.All;
         }
 
+        var runAsRemoteHttpService = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.RunAsRemoteHttpService.Name);
+        var outgoingAuthStrategy = parseResult.GetValueOrDefault<OutgoingAuthStrategy>(ServiceOptionDefinitions.OutgoingAuthStrategy.Name);
+
+        // Apply safe defaults for outgoing authentication strategy based on hosting mode
+        if (outgoingAuthStrategy == OutgoingAuthStrategy.NotSet)
+        {
+            outgoingAuthStrategy = runAsRemoteHttpService
+                ? OutgoingAuthStrategy.UseOnBehalfOf
+                : OutgoingAuthStrategy.UseHostingEnvironmentIdentity;
+        }
+
         var options = new ServiceStartOptions
         {
             Transport = parseResult.GetValueOrDefault<string>(ServiceOptionDefinitions.Transport.Name) ?? TransportTypes.StdIo,
@@ -106,7 +129,9 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
             ReadOnly = parseResult.GetValueOrDefault<bool?>(ServiceOptionDefinitions.ReadOnly.Name),
             Debug = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.Debug.Name),
             EnableInsecureTransports = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.EnableInsecureTransports.Name),
-            InsecureDisableElicitation = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.InsecureDisableElicitation.Name)
+            InsecureDisableElicitation = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.InsecureDisableElicitation.Name),
+            RunAsRemoteHttpService = runAsRemoteHttpService,
+            OutgoingAuthStrategy = outgoingAuthStrategy
         };
         return options;
     }
@@ -246,6 +271,26 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
     }
 
     /// <summary>
+    /// Validates that the outgoing authentication strategy is compatible with the hosting mode.
+    /// </summary>
+    /// <param name="runAsRemoteHttpService">Whether the server is running as a remote HTTP service.</param>
+    /// <param name="outgoingAuthStrategy">The outgoing authentication strategy.</param>
+    /// <param name="commandResult">Command result to update on failure.</param>
+    private static void ValidateOutgoingAuthStrategy(
+        bool runAsRemoteHttpService,
+        OutgoingAuthStrategy outgoingAuthStrategy,
+        CommandResult commandResult)
+    {
+        // UseOnBehalfOf is only valid when running as a remote HTTP service
+        if (!runAsRemoteHttpService && outgoingAuthStrategy == OutgoingAuthStrategy.UseOnBehalfOf)
+        {
+            commandResult.AddError($"The {OutgoingAuthStrategy.UseOnBehalfOf} outgoing authentication" +
+                $" strategy is only valid when running as a remote HTTP service " +
+                $"(--{ServiceOptionDefinitions.RunAsRemoteHttpServiceName}).");
+        }
+    }
+
+    /// <summary>
     /// Provides custom error messages for specific exception types to improve user experience.
     /// </summary>
     /// <param name="ex">The exception to format an error message for.</param>
@@ -258,6 +303,8 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
             "Invalid mode option specified. Use --mode single, namespace, or all for the supported modes.",
         ArgumentException argEx when argEx.Message.Contains("--namespace and --tool options cannot be used together") =>
             "Configuration error: The --namespace and --tool options are mutually exclusive. Use either one or the other to filter available tools.",
+        ArgumentException argEx when argEx.Message.Contains($"{OutgoingAuthStrategy.UseOnBehalfOf} outgoing authentication strategy") =>
+            $"Configuration error: The {OutgoingAuthStrategy.UseOnBehalfOf} authentication strategy requires --{ServiceOptionDefinitions.RunAsRemoteHttpServiceName} to be enabled.",
         InvalidOperationException invOpEx when invOpEx.Message.Contains("Using --enable-insecure-transport") =>
             "Insecure transport configuration error. Ensure proper authentication configured with Managed Identity or Workload Identity.",
         _ => base.GetErrorMessage(ex)
@@ -270,7 +317,7 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
     /// <returns>An IHost instance configured for the MCP server.</returns>
     private IHost CreateHost(ServiceStartOptions serverOptions)
     {
-        if (serverOptions.EnableInsecureTransports)
+        if (serverOptions.EnableInsecureTransports || serverOptions.RunAsRemoteHttpService)
         {
             return CreateHttpHost(serverOptions);
         }
@@ -315,6 +362,9 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
             })
             .ConfigureServices(services =>
             {
+                // Configure the outgoing authentication strategy.
+                services.AddSingleIdentityTokenCredentialProvider();
+
                 ConfigureServices(services);
                 ConfigureMcpServer(services, serverOptions);
             })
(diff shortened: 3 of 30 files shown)


core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -326,48 +373,163 @@ private IHost CreateStdioHost(ServiceStartOptions serverOptions)
...
+                                $"Bearer realm=\"{request.Host}\", resource_metadata=\"{resourceMetadataUrl}\"";
+                        }
+                        return Task.CompletedTask;
+                    }
+                };
+            });
+
+            services.AddAuthorizationBuilder()
+                .SetFallbackPolicy(null)
+                .AddPolicy("McpAccess", policy =>
```
anuchandy, line 426: Note - for Postgres and Speech, we’re currently using the `McpToolExecutor` policy with the role value `Mcp.Tool.Executor`

```C#
builder.Services.AddAuthorization(options =>
{
    options.AddPolicy("McpToolExecutor", p => p.RequireRole("Mcp.Tool.Executor"));

    ...
});
```

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -95,6 +109,17 @@ protected override ServiceStartOptions BindOptions(ParseResult parseResult)
             mode = ModeTypes.All;
         }
 
+        var runAsRemoteHttpService = parseResult.GetValueOrDefault<bool>(ServiceOptionDefinitions.RunAsRemoteHttpService.Name);
+        var outgoingAuthStrategy = parseResult.GetValueOrDefault<OutgoingAuthenticationTypes>(ServiceOptionDefinitions.OutgoingAuthStrategy.Name);
+
+        // Apply safe defaults for outgoing authentication strategy based on hosting mode
+        if (outgoingAuthStrategy == OutgoingAuthenticationTypes.NotSet)
```
anuchandy: Should we align option name and its enum name - `OutgoingAuthStrategy` option and `OutgoingAuthStratigies` enum

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -326,48 +373,163 @@ private IHost CreateStdioHost(ServiceStartOptions serverOptions)
...
-                logging.AddConsole();
-            })
-            .ConfigureWebHostDefaults(webBuilder =>
+        WebApplicationBuilder builder = WebApplication.CreateBuilder();
+
+        // Configure URLs
+        string? url = GetSafeAspNetCoreUrl(serverOptions);
+        if (url != null)
+        {
+            builder.WebHost.UseUrls(url);
```
anuchandy: Just wondering if it might be a bit more readable if we write it as:

```C#
if (serverOptions.EnableInsecureTransports)
{
    // For 'EnableInsecureTransports' validate the url, when running as a remote 
    // HTTP service ('RunAsRemoteHttpService'), let the typical IConfiguration
    // binding handle the ASPNETCORE_URLS setup.
    //
    string url = GetSafeAspNetCoreUrl();
    builder.WebHost.UseUrls(url);
}
```

Basically, `if InsecureTransports` `then SafeAspNetCoreUrl` go hand in hand, so grouping them together makes it  a bit more readable, I feel.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -326,48 +373,163 @@ private IHost CreateStdioHost(ServiceStartOptions serverOptions)
...
+            mcpEndpointBuilder.RequireAuthorization("McpAccess");
+        }
+
+        // Map non-MCP endpoints as needed.
+        if (serverOptions.RunAsRemoteHttpService)
+        {
+            // Health checks are anonymous (no authentication required)
+            app.MapHealthChecks("/health")
+                .AllowAnonymous();
+        }
```
anuchandy, line 561: nit: wondering if it makes sense to merge these two if statements.

```C#
if (serverOptions.RunAsRemoteHttpService)
{
   // All MCP endpoints require MCP.All scope or role
   app.MapMcp().RequireAuthorization("McpAccess"); 
   // Health checks are anonymous (no authentication required)
   app.MapHealthChecks("/health").AllowAnonymous();

}
```

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -382,8 +544,15 @@ private static void ConfigureMcpServer(IServiceCollection services, ServiceStart
...
     /// <returns>A validated URL string for ASP.NET Core binding.</returns>
-    private static string GetSafeAspNetCoreUrl()
+    private static string? GetSafeAspNetCoreUrl(ServiceStartOptions options)
     {
+        if (options.RunAsRemoteHttpService)
+        {
+            // When running as a remote HTTP service, let the typical IConfiguration
+            // binding handle the ASPNETCORE_URLS value without additional validation.
+            return null;
+        }
```
anuchandy, line 585: If we make this change ([link](https://github.com/microsoft/mcp/pull/910/files#r2450058269)), we can keep this method as is and return string instead of string?.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/AuthenticationServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,120 @@
...
+    private static IServiceCollection AddHttpOnBehalfOfTokenCredentialProvider(
+        this IServiceCollection services,
+        IConfiguration configuration)
+    {
+        // Dependencies - directly in constructor.
+        services.AddHttpContextAccessor();
+
+        // Dependencies - indirectly required to get MicrosoftIdentityTokenCredential.
+        services.AddMicrosoftIdentityWebAppAuthentication(configuration)
+            .EnableTokenAcquisitionToCallDownstreamApi()
```
anuchandy: The discussion with Identity team on non-AOT nature of this API https://github.com/AzureAD/microsoft-identity-web/issues/3576

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+    public Task<TokenCredential> GetTokenAsync(string? tenantId, CancellationToken cancellationToken)
+    {
+        if (_httpContextAccessor.HttpContext is not HttpContext httpContext)
+        {
+            throw new InvalidOperationException("There is no ongoing HTTP request.");
+        }
+
+        if (tenantId is not null)
+        {
+            if (httpContext.User.FindFirst("tid")?.Value is string tidClaim
```
anuchandy, line 41: Wondering - are we guaranteed to only receive v2.0 tokens here, if not should we consider fall back to the legacy claim URI?

```C#
var principal = httpContext.User;
principal.FindFirst("tid")?.Value
    ?? principal.FindFirst("http://schemas.microsoft.com/identity/claims/tenantid")?.Value;
```

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+        _logger = logger;
+    }
+
+    /// <inheritdoc/>
+    public Task<TokenCredential> GetTokenAsync(string? tenantId, CancellationToken cancellationToken)
+    {
+        if (_httpContextAccessor.HttpContext is not HttpContext httpContext)
+        {
+            throw new InvalidOperationException("There is no ongoing HTTP request.");
+        }
```
anuchandy, line 31: Should we add a preemptive check to be more defensive?

```C#
if (httpContext.User.Identity?.IsAuthenticated != true)
{
     throw new InvalidOperationException("User is not authenticated.");
}
```

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
```diff
@@ -40,42 +52,54 @@ public async Task<List<TenantResource>> GetTenants()
         return results;
     }
 
-    public bool IsTenantId(string tenant)
+    /// <inheritdoc/>
+    public bool IsTenantId(string tenantId)
     {
-        return Guid.TryParse(tenant, out _);
+        return Guid.TryParse(tenantId, out _);
```
anuchandy, line 58: Thanks for cleaning up the variable name - definitely easier to read now.

anuchandy (commented):
thanks Steven!

tools/Azure.Mcp.Tools.Marketplace/tests/Azure.Mcp.Tools.Marketplace.LiveTests/ProductListCommandTests.cs
```diff
@@ -23,7 +25,8 @@ public ProductListCommandTests(ITestOutputHelper output) : base(output)
     {
         var memoryCache = new MemoryCache(Microsoft.Extensions.Options.Options.Create(new MemoryCacheOptions()));
         var cacheService = new CacheService(memoryCache);
-        var tenantService = new TenantService(cacheService);
+        var tokenProvider = new SingleIdentityTokenCredentialProvider(NullLoggerFactory.Instance);
+        var tenantService = new TenantService(tokenProvider, cacheService);
```
anuchandy, line 29: Note: I noticed that the `MarketplaceService` (as well as `PostgresService`, `MySqlService` and `MonitorHealthModelService`) uses class member–based caching instead of the `ICacheService`. For reference, in [this exploration commit](https://github.com/microsoft/mcp/pull/786/commits/e5957de21a0354be00de72ce37d1de0c127f3522) and https://github.com/microsoft/mcp/pull/786/commits/19d22c7aa280d011d070e610821b84844c8df977, I removed those member variables in favor of ICacheService so the OBO flow remains safe (with a no-op implementation of ICacheService).

Directory.Packages.props
```diff
@@ -68,6 +68,9 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="9.0.9" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="9.0.9" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="9.0.9" />
+    <PackageVersion Include="Microsoft.Identity.Abstractions" Version="9.5.0" />
+    <PackageVersion Include="Microsoft.Identity.Web" Version="3.14.0" />
```
anuchandy: nit - Identity team suggested (in the AOT internal channel) today that we should use 4.0.0.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -64,14 +64,7 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
...
         // Register tool loader strategies
         services.AddSingleton<CommandFactoryToolLoader>();
-        services.AddSingleton(sp =>
-        {
-            return new RegistryToolLoader(
-                sp.GetRequiredService<RegistryDiscoveryStrategy>(),
-                sp.GetRequiredService<IOptions<ToolLoaderOptions>>(),
-                sp.GetRequiredService<ILogger<RegistryToolLoader>>()
-            );
-        });
```
anuchandy, line 74: Just curious - is there a specific reason for changing (or deleting) this? Could it affect any other flow?

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Must we use `McpToolExecutor` and `Mcp.Tool.Executor`?

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
vukelich: The previous way of a delegate to create the instance is useful if you need to dynamically pick and implementation type when exposing an interface type through DI (we expose the concrete type rather than an interface), if you're treating the delegate as a factory for transient or scoped lifetime instances (we are using singleton), or if your implementation has multiple constructors and/or extra setup steps after instantiation (neither relevant here). 

The DI container's resolution will effectively do the same as what was done before. By removing the explicit call to the constructor, we're letting this set of the code be agnostic to changes to the type's constructor.

tools/Azure.Mcp.Tools.Marketplace/tests/Azure.Mcp.Tools.Marketplace.LiveTests/ProductListCommandTests.cs
vukelich: Thank you for mentioning that. Let's knock out caching, directly or via `ICacheService`, in a follow up PR. The feature branch is there for us to iterate.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
vukelich: Thanks! I was tempted to go even further and add signature overloads that took `Guid` instead of `string` when the method wants just a GUID string value. But I decided that can be a smaller, focused PR if needed.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
vukelich: Yes! We can guarantee that by making setting that in the app registration's manifest. If you haven't already done that in the scripts you've written, please add them. We don't have a legacy product, so I think it's good to avoid introducing such back-compat.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: I think I'm going to update `GetSafeAspNetCoreUrl` to take in the builder so that it can encapsulate both possible paths then return `void`.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: I think I'm going to update `GetSafeAspNetCoreUrl` to take in the builder so that it can encapsulate both possible paths then return `void`.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: I was considering that, but I also wanted to make it very clear for the uninitiated the pieces that are "for MCP goodness" vs "for regular web API goodness" at this part in the middleware setup. I was using conditional blocks similar to how some folks use `#region`.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+                    tidClaim);
+
+                throw new InvalidOperationException(
+                    $"The requested token tenant '{tenantId}' does not match the tenant of the authenticated user '{tidClaim}'.");
+            }
+        }
+
+        // MicrosoftIdentityTokenCredential is registered as scoped, so we
+        // can get it from the request services to ensure we get the right instance.
+        MicrosoftIdentityTokenCredential tokenCredential = httpContext
```
anuchandy, line 56: Curious - any reason for us to use `MicrosoftIdentityTokenCredential` instead of `ITokenAcquisition`?  `MicrosoftIdentityTokenCredential` adds a new dependency (`Microsoft.Identity.Web.Azure`), so just wondering if it provides enough value over `ITokenAcquisition.GetAccessTokenForUserAsync(scopes)`. While exploring, I used ITokenAcquisition (example: [JwtOboCredentialProvider.cs#L160](https://github.com/microsoft/mcp/blob/d9190a39066e35be0c5144bdf6f61b4d56716b69/core/Azure.Mcp.Core/src/Services/Azure/Authentication/JwtOboCredentialProvider.cs#L160)) and wrapped it in a TokenCredential, so curious o learn.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
anuchandy: No, we just need to agree on what goes in and make sure the Postgres demo code uses the same role. We should also document the same.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
anuchandy: I see - thanks for the explaining, that makes sense.

tools/Azure.Mcp.Tools.Marketplace/tests/Azure.Mcp.Tools.Marketplace.LiveTests/ProductListCommandTests.cs
anuchandy: Sounds good!

joshfree:
adding `Do Not Merge` to exclude it from the October/1.0 query for now

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Agreed. I knew Microsoft services, like MS Graph, have some established patterns, so I asked ChatGPT to come up with some thoughts: https://chatgpt.com/share/68f91558-b140-8013-a26e-5ef666821c56

TL;DR: 
1. Dot notation good as something like `Noun.Verb` or `Noun.All`.
2. Delegated permissions and app roles (aka app permissions) should have similar naming. The app roles typically have a similar set as the delegated permissions but with `.All` suffixed. 

I like the idea a middle ground of `Mcp.Tools.ReadWrite` for now. `Mcp` as a pretty meaningful root for all MCP goodness, `Tools` as child representing tool calling, and `Verb` suffix representing running tools that can read or read/write. Another child of `Mcp` at some point might be `Agent` if MCP introduces an agentic concept so we'd have `Mcp.Agent.Verb`. This incorporates our joint `Mcp` prefixing, your `Tool` hierarchy substring, and MS Graph's pattern and my usage of a `Verb` suffix. 

How's that sound to you?

anuchandy: Thanks for digging into this further and sharing the discussion link! This sounds good to me - I like the `Mcp.Tool.{Verb}` pattern. Agree that using verbs like Read or ReadWrite makes the intent much clearer (than a generic "execute") and, the pattern looks extensible, and keep things consistent, and easy to reason about.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
anuchandy: Got it, thanks for simplifying this!

core/Azure.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+            return Task.FromResult(_credential);
+        }
+
+        if (!_tenantSpecificCredentials.TryGetValue(tenantId, out TokenCredential? tenantCredential))
+        {
+            tenantCredential = new CustomChainedCredential(
+                tenantId,
+                _loggerFactory.CreateLogger<CustomChainedCredential>()
+            );
+            _tenantSpecificCredentials[tenantId] = tenantCredential;
```
anuchandy: If I understand correctly, when the strategy is `UseHostingEnvironmentIdentity` and the server runs with `--run-as-remote-http-service` or `--enable-insecure-transports`, multiple HTTP requests may execute concurrently and access the singleton `SingleIdentityTokenCredentialProvider`, which maintains the `_tenantSpecificCredentials` dictionary. Should we make this dictionary thread-safe to avoid potential races (e.g., that could corrupt its internal state during resize or rehash operations)?

```C#
private readonly ConcurrentDictionary<string, TokenCredential> _tenantCredentials = new();

...

var tenantCredential = _tenantCredentials.GetOrAdd(
   tenantId,
   id => new CustomChainedCredential(id, _loggerFactory.CreateLogger<CustomChainedCredential>())
);

return Task.FromResult(tenantCredential);
```

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
```diff
@@ -1,15 +1,101 @@
...
+    /// </remarks>
+    /// <exception cref="OperationCanceledException">
+    /// Thrown when the operation has been cancelled.
+    /// </exception>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when a credential cannot be provided.
+    /// </exception>
+    Task<TokenCredential> GetTokenAsync(
+        string? tenantId,
+        CancellationToken cancellationToken);
```
anuchandy, line 100: I really like the idea of treating ITenantService as the provider of credentials.  (so `IAzureTokenCredentialProvider` injected only in one place)

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
```diff
@@ -40,42 +52,54 @@ public async Task<List<TenantResource>> GetTenants()
...
             throw new InvalidOperationException($"Tenant with ID {tenantId} has a null DisplayName");
 
-        return tenant.Data.DisplayName;
+        return tenantName;
+    }
+
+    /// <inheritdoc/>
+    public async Task<TokenCredential> GetTokenAsync(string? tenantId, CancellationToken cancellationToken)
+    {
+        return await _credentialProvider.GetTokenAsync(tenantId, cancellationToken);
```
anuchandy: Nit: one minor thing I got confused about - the method name `GetTokenAsync` initially made me think it returns an `Azure.Core.AccessToken`, but it actually returns a `Azure.Core.TokenCredential`. Since `Azure.Core.TokenCredential` itself has a `GetTokenAsync` method, should we consider renaming this to `ITenantService::GetTokenCredentialAsync` and `IAzureTokenCredentialProvider::GetTokenCredentialAsync` for better clarity?

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
```diff
@@ -1,20 +1,32 @@
...
+using Azure.Mcp.Core.Services.Azure.Authentication;
 using Azure.Mcp.Core.Services.Caching;
 using Azure.ResourceManager;
 using Azure.ResourceManager.Resources;
 
 namespace Azure.Mcp.Core.Services.Azure.Tenant;
 
-public class TenantService(ICacheService cacheService)
-    : BaseAzureService, ITenantService
+public class TenantService: BaseAzureService, ITenantService
```
anuchandy, line 12: With `TenantService` now responsible for providing credentials, I’m wondering if there’s still a need for it to inherit from `BaseAzureService`?, if not it will let us avoid exposing the protected `TenantService` property in `BaseAzureService`, which is a workaround to break the circular dependency, thinking loud

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// This can be overridden using <see cref="Authentication.AuthenticationServiceCollectionExtensions.AddAzureTokenCredentialProvider"/>
+    /// based on parsed command line arguments and environment variables.
+    /// </description>
+    /// </item>
+    /// </list>
+    /// </remarks>
+    public static IServiceCollection AddAzureTenantService(this IServiceCollection services)
+    {
+        // !!! HACK !!!
+        // Program.cs for the CLI servers have their own DI containers vs ServiceStartCommand.
```
anuchandy, line 42: Completely out of scope for this work, but for future GA releases, we might want to revisit whether we still need to support the fully functional CLI mode (beyond `server start` and `--version`). I recall GHCP4z used to rely on the CLI mode at one point, but that’s no longer the case. Feel free to resolve the comment - I just wanted to highlight the additional overhead of maintaining a mode (CLI) whose value for MCP server users is unclear.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Leaving as-is. Resolving.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/AuthenticationServiceCollectionExtensions.cs
vukelich: As mentioned over a call, I can remove outgoing auth dependencies easily. Incoming auth dependencies are unavoidable. We'll leave what I have as-is for now, then decide to change it if we find issues or if we can't get all AOT compatibility in place.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
vukelich: `MicrosoftIdentityTokenCredential` is the bridge between `Microsoft.Identity.Web` and the Azure SDK for `TokenCredential` when we want to use the "downstream API" features of `Microsoft.Identity.Web`. [See the official documentation here.](https://github.com/AzureAD/microsoft-identity-web/blob/master/README-Azure.md)

If we don't want to use `Microsoft.Identity.Web` libraries for outgoing auth, then I'd recommend we instead use the Azure SDK's `OnBehalfOfCredential` and supply it with all the arguments needed manually. To get the client assertion needed, we can just grab it from the incoming request's `Authorization` header's Bearer token value. There's also some minor trickery needed to find the appropriate client ID and tenant ID. [This was in my earlier proof-of-concept here.](https://github.com/AzureAD/microsoft-identity-web/blob/master/README-Azure.md) That avoids any needs on `Microsoft.Identity.Web` libraries for downstream calls and keeps us wholly in ASP.NET Core and Azure SDK land. The `Microsoft.Identity.Web` libraries abstract some of that for us, and `Microsoft.Identity.Web.Azure` makes the last step similarly trivial.

As discussed in a call today, I'm indifferent on using `Microsoft.Identity.Web` libraries for outbound auth while I see it as required to use the libraries needed for setting up incoming auth.

vukelich: Resolving this without any code changes. Please reopen if you disagree.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
vukelich: Good catch. We can just use a simple `lock` block.

vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
vukelich: Thanks. I agree for the current state of the repo. 

I'd like to remove a lot of what's in `BaseAzureService`, such as its `static` methods and its dependency on `ITenantService`, so that derived types that are much simpler can take more targeted and direct dependencies on what they actually need, like being able to directly depend on `IAzureTokenCredentialProvider` if there is an outgoing Azure auth need. A lot of difficulty we're going to have in testing this are the multiple dependency injection containers (in Azure and Fabric's `Program.cs` vs Core's `ServiceStartCommand.cs`), and each of those DI's has its own copies of `ITenantService` and much of everything else.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
vukelich: Sure, I can do that.

vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
vukelich: Totally agreed. However, that will touch a lot of stuff that I'd rather do in a follow up PR. I want to have such a dramatic change separate from the functionality added in this PR.

vukelich: Resolving as won't fix. Please reopen if you disagree.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
vukelich: I agree, but I'm very biased with my only focus being on `server start` as a scenario. @joshfree mentioning you because this focusing on the azmcp.exe being exclusively an MCP server would be very nice IMO but you have a clearer product vision.

vukelich: Marking as resolved for my organization of PR feedback. Josh, feel free to comment here or just emoji.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
vukelich: Marking as resolved. Thanks for the callout.

Directory.Packages.props
vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
vukelich: Updated locally. Pending push.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Renaming enum to `OutgoingAuthStrategy`. Updated locally. Pending push.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Pushed. Please take a look. Marking as resolved.

Directory.Packages.props
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/HttpOnBehalfOfTokenCredentialProvider.cs
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Pushed. Please take a look. Marking as resolved.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
vukelich: Pushed. Please take a look. Marking as resolved.

anuchandy (approved):

PR #912, opened 2025-10-22
tool description and prompt updates for app service add database command

### LINKED ISSUE #860: Some prompts in `Azure App Service` cannot stably trigger corresponding tools
**Description:**
When entering prompts such as `Add a MySQL database to app service <app_name>`, `Configure a SQL Server database for app service <app_name>`, `Add a PostgreSQL database to app service <app_name>`, `Add a CosmosDB database to app service <app_name>`, `Add database <database_name> on server <database_server> to app service <app_name>`, or `Set connection string for database <database_name> in app service <app_name>`, the `azmcp_appservice_database_add` tool is not triggered **reliably**. **In five tests, there were three failures where the tool did not respond as expected.**
Image
Image

**Steps to Reproduce:**
1. Start mcp server in `namespaces` mode.
2. Open GitHub Copilot in VS Code and [switch to Agent mode](https://code.visualstudio.com/docs/copilot/chat/chat-agent-mode).
3. Click Refresh on the tools list, and select the `Built-in`, `Extension: Github Copilot for Azure`, `MCP Server: Microsoft Docs` and `MCP Server: Azure MCP`.
4. Enter test prompt `Add a MySQL database to app service <app_name>`.

**Environment:**

- OS: Windows.
- Vscode Version: 1.105.0
- GitHub Copilot Chat Version: 0.32.1
- Large Language Model: GPT-4.1
- Virtual Tool: Off
- Azure MCP Server: 0.9.0 and 0.9.3

**Expected Behavior:**
The `azmcp_appservice_database_add` tool can be triggered reliably.

@joshfree for notification.

KarishmaGhiya:
This PR updates the tool description and the e2e test prompts for `app service add database command`
Fixes #860

### DIFF
--- a/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
+++ b/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
@@ -113,8 +113,8 @@ This file contains prompts used for end-to-end testing to ensure each tool is in
 | azmcp_appservice_database_add | Add database connection <connection_string> for database <database_name> on server <database_server> to app service <app_name> in resource group <resource_group> |
 | azmcp_appservice_database_add | Add database connection string for <database_name> to app service <app_name> using connection string <connection_string> in resource group <resource_group> |
 | azmcp_appservice_database_add | Connect database <database_name> to my app service <app_name> using connection string <connection_string> in resource group <resource_group> |
-| azmcp_appservice_database_add | Set up database <database_name> for app service <app_name> with connection string <connection_string> under resource group <resource_group_name> |
-| azmcp_appservice_database_add | Configure database <database_name> for app service <app_name> with the connection string <connection_string> in resource group <resource_group_name> |
+| azmcp_appservice_database_add | Set up database <database_name> for app service <app_name> with connection string <connection_string> under resource group <resource_group> |
+| azmcp_appservice_database_add | Configure database <database_name> for app service <app_name> with the connection string <connection_string> in resource group <resource_group> |
 
 ## Azure Application Insights
 


servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
```diff
@@ -105,15 +105,16 @@ This file contains prompts used for end-to-end testing to ensure each tool is in
...
-| azmcp_appservice_database_add | Configure a SQL Server database for app service <app_name> |
-| azmcp_appservice_database_add | Add a MySQL database to app service <app_name> |
-| azmcp_appservice_database_add | Add a PostgreSQL database to app service <app_name> |
-| azmcp_appservice_database_add | Add a CosmosDB database to app service <app_name> |
-| azmcp_appservice_database_add | Add database <database_name> on server <database_server> to app service <app_name> |
-| azmcp_appservice_database_add | Set connection string for database <database_name> in app service <app_name> |
-| azmcp_appservice_database_add | Configure tenant <tenant> for database <database_name> in app service <app_name> |
-| azmcp_appservice_database_add | Add database <database_name> with retry policy to app service <app_name> |
+| azmcp_appservice_database_add | Add database connection <connection_string> to my app service <app_name> for database <database_name> in resource group <resource_group> in resource group <resource_group_name> under subscription <subscription> in tenant <tenant_id> |
```
srnagar: If the user doesn't provide the resource group, subscription and tenant in the prompt, shouldn't the agent fetch those (using other tools or from chat context) before calling this tool? The connection string and the database name should be the required details. We shouldn't require users to specify the subscription and tenant id in every prompt.

tools/Azure.Mcp.Tools.AppService/src/Commands/Database/DatabaseAddCommand.cs
```diff
@@ -21,8 +21,8 @@ public sealed class DatabaseAddCommand(ILogger<DatabaseAddCommand> logger) : Bas
 
     public override string Description =>
         """
-        Add a database connection to an App Service. This command configures database connection
-        settings for the specified App Service, allowing it to connect to a database server.
+        Add a database connection for an App Service using connection string for an existing database. This command configures database connection
+        settings for the specified App Service, allowing it to connect to a database server. You must specify the App Service name, database name, database type, server, connection string, resource group name, tenant and subscription.
```
srnagar: We need resource group, subscription and tenant information for almost all the tools. Do we need the user to specify it in every single prompt?

servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
KarishmaGhiya: Ideally yes. But the number of times the tool call failed during testing was bcz of missing subscription/ resource group name even when it called the mcp tool correctly. It can detect subscription from context but not resource group name.

tools/Azure.Mcp.Tools.AppService/src/Commands/Database/DatabaseAddCommand.cs
KarishmaGhiya: No. We don't need the users to specify these details in every prompt for other tools. But for this tool, I have noticed that without these details it's difficult for the LLM agent to figure out which database and app service, the behavior is unpredictable. 

In the sense, sometimes it will create it's own script at one point to query all subscriptions to find the correct resource group name etc. That is unnecessary. So might as well have the user provide it in a context either before or during the prompt.

KarishmaGhiya: Checked again, LLM can detect subscription from context but not resource group name. Tenant is not needed. Will update the description.

PR #964, opened 2025-10-27
Support Recordable LiveTests

### LINKED ISSUE #541: Create `test-proxy` client, integrate in `CommandTestBase`
The client should be based on the `tsp`.

Add a debug-only section of code that adds a DelegatingHandler for `TEST_PROXY_URL`. This will be set by the `stdiotransport` that `CommandTestBase` uses to spin up the mcp server for test use.

[Coding here](https://github.com/Microsoft/mcp/blob/main/core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs)

scbedd:
Resolves #541

Before merge, need to:

- [x] Finish recordings for `Keyvault` tests
- [x] Verify playback for `Keyvault` tests
  - [x]  Need to support sanitizer registration/some updates to the base sanitization configs before this can merge. I was hoping I could get away with v0 not supporting it but no dice.
- [x] Refactor Azure.Mcp.Core modifications to only be present during debug compilation
- [x] Cannot use non-nuget feed, so convert to download standalone proxy executable
- [x] Confirm playback mismatch properly reports to the user
  - This works excellently now: image
- [x] Clean up remaining todos
- [x] Ensure `Keyvault` livetests are unaffected in CI

After merge:
- [ ] Swap the default in CI to recorded, not live
- [ ] Add new copilot instructions for general support and usage of the test-proxy
- [ ] Use new copilot instructions to migrate existing tools to recordings
- [ ] Write transition document that `copilot` can use to transition a LiveTest to a recorded one

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/TestProxy.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/TestProxy.cs
@@ -262,7 +262,7 @@ public sealed class TestProxy(bool debug = false) : IDisposable
         {
             while (!ct.IsCancellationRequested && !reader.EndOfStream)
             {
-                var line = await reader.ReadLineAsync().ConfigureAwait(false);
+                var line = await reader.ReadLineAsync(ct).ConfigureAwait(false);
                 if (line == null)
                     break;
                 lock (sink)


core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -50,14 +50,24 @@ public override ValueTask<AccessToken> GetTokenAsync(TokenRequestContext request
...
     private static bool ShouldUseOnlyBrokerCredential()
     {
         return EnvironmentHelpers.GetEnvironmentVariableAsBool(OnlyUseBrokerCredentialEnvVarName);
     }
 
+    private static bool IsPlaybackMode() => string.Equals(Environment.GetEnvironmentVariable(PlaybackModeEnvVarName), "true", StringComparison.OrdinalIgnoreCase);
+
     private static TokenCredential CreateCredential(string? tenantId, ILogger<CustomChainedCredential>? logger = null)
     {
+        // Short-circuit for playback to avoid any real auth & interactive prompts.
```
hallipr, line 107: AZURE_TOKEN_CREDENTIALS already supports specific credential targeting.

Is there a functional difference between 
`AZURE_TOKEN_CREDENTIALS = 'PlaybackTokenCredential'` and `AZURE_MCP_PLAYBACK_MODE = 'true'`?

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
```diff
@@ -53,7 +53,26 @@ public HttpClient CreateClient(Uri? baseAddress, Action<HttpClient> configureCli
...
+            Console.WriteLine("Inserting RecordingRedirectHandler for test proxy.");
+            // RecordingRedirectHandler should be the last delegating handler before the transport
+            pipeline = new RecordingRedirectHandler(proxyUri)
+            {
+                InnerHandler = pipeline
+            };
+        }
+        var client = new HttpClient(pipeline);
+#endif
+#if !DEBUG
```
hallipr:
```suggestion
#else
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/Helpers/BinaryContentHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.ClientModel;
+using System.Text.Json;
+
+namespace Azure.Mcp.Tests.Client.Helpers;
+
+/// <summary>
+/// This entire class is a placeholder. Azure.Mcp.Core takes a dependency on Azure.Core, so what I should do is generate the test-proxy client
+/// taking an Azure.Core dependency to generate, versus System.ClientModel (which is what it is rn)
```
hallipr: We should consider that Microsoft.Fabric.Mcp.Core exists too.  Will it take an Azure.Core dependency?

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
scbedd: I haven't folded your upstream changes in. I'll definitely swap to that by EoD today. I'm getting through re-recording the keyvault tests.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
```diff
@@ -16,27 +18,57 @@ public abstract class CommandTestsBase(ITestOutputHelper output) : IAsyncLifetim
...
+                // ResourceBaseName used for vault name input to tools (maps to sanitized recording host).
+                ResourceBaseName = "Sanitized",
+                SubscriptionName = "Sanitized",
+                TenantName = "Sanitized"
+            };
+
+            // Signal playback mode to child process & any credential logic.
+            Environment.SetEnvironmentVariable("AZURE_MCP_PLAYBACK_MODE", "true");
+            Environment.SetEnvironmentVariable("AZURE_TENANT_ID", Settings.TenantId);
+            Environment.SetEnvironmentVariable("AZURE_SUBSCRIPTION_ID", Settings.SubscriptionId);
```
hallipr: Rather than set environment variables for the current process, can these be moved into the initialization of the McpClient client?

hallipr: For local credential selection (clients on the test side) we should just require that devs use a custom credential that takes testSettings as a parameter.  That credential would just choose between either AzurePowerShellCredential or PlaybackTokenCredential

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
```diff
@@ -46,15 +78,28 @@ public virtual async ValueTask InitializeAsync()
         string[] defaultArgs = enableDebug
             ? ["server", "start", "--mode", "all", "--debug"]
             : ["server", "start", "--mode", "all"];
-        var arguments = _customArguments ?? defaultArgs;
+        var arguments = CustomArguments ?? defaultArgs;
+
+        var dictionaryEvents = new Dictionary<string, string?> {
```
hallipr: This is an oddly named variable

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        await StopRecordOrPlayback();
+        await base.DisposeAsync();
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        base.Dispose(disposing);
+
+        // On test failure, append proxy stderr for diagnostics.
```
hallipr, line 183: Can this be simplified by adding it to the end of your DisposeAsync override?

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/Helpers/BinaryContentHelper.cs
scbedd: Honestly I should update this summary as it doesn't really reflect reality anymore.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
scbedd: I need to fold your changes in, then we'll see.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/TestProxy.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            ZipFile.ExtractToDirectory(downloadPath, proxyDir);
+        }
+
+        _cachedExecutable = _findExecutableInDirectory(proxyDir);
+
+        return _cachedExecutable;
+    }
+
+    private bool _checkProxyVersion(string proxyDirectory, string version)
```
hallipr: These method names are using an odd naming scheme.  Our `.editorconfig` suggests private and internal fields will be lowercase with an underscore and static fields get an additional "s" prefix, but only fields get special treatment.  All other members use PascalCase, no matter the static or access modifiers.

```suggestion
    private bool CheckProxyVersion(string proxyDirectory, string version)
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/TestProxy.cs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    /// <summary>
+    /// Snapshots the current stderr output from the testproxy. This is a destructive read; the internal buffer is cleared after the call.
+    ///
+    /// This means that if multiple tests fail in sequence, each test will only see the stderr output generated since the last call to SnapshotStdErr(), which means
+    /// we won't be seeing errors from previous test failures. This is intentional to ensure that each test only gets the relevant stderr output.
+    /// </summary>
+    /// <returns></returns>
+    public string? SnapshotStdErr()
```
hallipr, line 318: [nit] Order methods from most to least public:
- public
- internal
- protected internal
- protected
- private protected
- private

We used to be able to enforce this through [StyleCop rule SA1202](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md)

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -55,9 +55,22 @@ private static bool ShouldUseOnlyBrokerCredential()
     {
         return EnvironmentHelpers.GetEnvironmentVariableAsBool(OnlyUseBrokerCredentialEnvVarName);
     }
+#if DEBUG
```
weshaggard: I assume, given all these core changes are under DEBUG only then we will not be able to run recordings or playback in any release builds. I suspect that will be fine for our current implementation but something we need to see about enforcing somehow in order to not break things.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/Microsoft.ClientModel.TestFramework.sln
```diff
@@ -0,0 +1,48 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
hallipr: The project this .sln references is missing, so we probably don't need this file either

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Azure.Mcp.Tests.csproj
```diff
@@ -1,6 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <IsAotCompatible>false</IsAotCompatible>
+    <!-- Suppress NuGet source mapping and downgrade warnings -->
```
hallipr: What warning were you getting here?

eng/emitter-package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "main": "dist/src/index.js",
+  "dependencies": {
+    "@typespec/http-client-csharp": "latest"
```
weshaggard: Why is this needed? Should we be careful about using latest?

eng/emitter-package.json
```diff
@@ -0,0 +1,23 @@
+{
```
hallipr: I see the value of including the framework for regenerating the proxy client in the repo.  Adding some powershell to eng/scripts that uses typespec to regen the client would be nice.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/Internal/Argument.cs
```diff
@@ -0,0 +1,74 @@
...
+
+#nullable disable
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Azure.Mcp.Tests.Generated.Internal
+{
+    internal static partial class Argument
```
weshaggard, line 14: What are the generated classes for? Are they intended to be mocks of some sort?

tools/Azure.Mcp.Tools.KeyVault/src/Services/KeyVaultService.cs
```diff
@@ -289,8 +291,40 @@ public async Task<KeyVaultCertificateWithPolicy> ImportCertificate(
...
+
+    private static Uri BuildVaultUri(string vaultName) => new($"https://{vaultName}.vault.azure.net");
+
+    // Create clients with injected HttpClient, this will enable record/playback during testing.
+    private KeyClient CreateKeyClient(string vaultName, Azure.Core.TokenCredential credential, RetryPolicyOptions? retry)
+    {
+        var httpClient = _httpClientService.CreateClient(BuildVaultUri(vaultName));
+        var options = new KeyClientOptions();
+        options = ConfigureRetryPolicy(AddDefaultPolicies(options), retry);
+        options.Transport = new Azure.Core.Pipeline.HttpClientTransport(httpClient);
```
hallipr: These look like they chain.  Does this work?

```suggestion
        var transport = new Azure.Core.Pipeline.HttpClientTransport(httpClient);
        var options = new KeyClientOptions { Transport = transport }
            .AddDefaultPolicies()
            .ConfigureRetryPolicy(retry);
```

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/RecordedKeyVaultCommandTests.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
```
hallipr: I don't think this is needed now

eng/emitter-package.json
scbedd: You're right. I'll rip it out and just leave a note about regenerating the client in a README at the root of the `generated` folder.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/Internal/Argument.cs
scbedd: They're the client to interact with the `test-proxy`. So I don't need to manually construct any of the requests that would be necessary to interact with the test-proxy.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/TestProxy.cs
scbedd: K. I'll go to this. It's a preference owned by the repo so if you're asking for the PascalCase then I'll go that way on all these functions.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/Microsoft.ClientModel.TestFramework.sln
scbedd: This definitely shouldn't be added here. Will clean it out.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Azure.Mcp.Tests.csproj
scbedd: No longer necessary! Will rip it out. Left behind from when I was suppressing the multiple-feed errors.

eng/emitter-package.json
scbedd: Wes left another comment. I'll leave behind a README at the root of the `Generated` directory that talks about the process to regenerate. I'll do that instead of keeping this around.

tools/Azure.Mcp.Tools.KeyVault/src/Services/KeyVaultService.cs
scbedd: I'll try it 👍

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/RecordedKeyVaultCommandTests.cs
scbedd: You're 100% right. Will clean it up.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
scbedd: I will swap to honor `AZURE_TOKEN_CREDENTIALS = 'PlaybackTokenCredential'` before fixing up the keyvault recordings.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
scbedd: Yes. I've cut over to using AZURE_TOKEN_CREDENTIALS and if it's PlaybackTokenCredential then run in playback mode.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
scbedd: Correct. I'll make certain it's noted in docs etc.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
scbedd: Yeah. Fair.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/RECORDED_TESTS.md
```diff
@@ -0,0 +1,22 @@
+# Updating `LiveTests` for `Record/Playback`
+
+This document will eventually contain a guide to enable Record/Playback for an existing `LiveTest` package. Until then it will contain a local todo for remaining feature work that needs to be finished.
```
scbedd: Remove.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: Maybe/probably Need to understand a bit more about what the base.Dispose is doing.

eng/emitter-package.json
scbedd: Removed.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: cleaned up.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/README.md
vcolin7, line 1: I don't think we ever auto-generate code in this repo.

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
```diff
@@ -6,13 +6,23 @@
...
+using Azure.Mcp.Tests.Generated.Models;
 using Azure.Security.KeyVault.Keys;
 using Xunit;
 
 namespace Azure.Mcp.Tools.KeyVault.LiveTests;
 
-public class KeyVaultCommandTests(ITestOutputHelper output) : CommandTestsBase(output)
+public class KeyVaultCommandTests(ITestOutputHelper output, TestProxyFixture fixture) : RecordedCommandTestsBase(output, fixture)
 {
+    public override List<BodyRegexSanitizer> BodyRegexSanitizers => new List<BodyRegexSanitizer>() {
```
vcolin7: I don't know if we should proactively sanitize/redact the contents of a certificate response to avoid triggering security alerts. [Here's how](https://github.com/Azure/azure-sdk-for-java/blob/4f7f2e3ebf614dbff43372a086fd5fa0c1095c16/sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java#L106) we did it for Java back in the day when recordings used to live in the same repo as our code. How does it work now that we have a separate assets repo?

vcolin7: Also, authentication for KV is challenge-based, meaning that the very first call we make to the service will be met with a `401` response with an authentication challenge the KV library will use to get a token from Entra. Do we know if that request/response for the token also be recorded?

scbedd: Yep. This shows up in the recording side as an initial `401` on the first request, and the response has a `www-authenticate` header with the bearer token id. (We sanitize this out  for playback)

scbedd: Once I get `playback` working here, I'll share the recordings with you before I push anything and get your `signoff`.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Generated/README.md
scbedd: This is for folks who need to regenerate this later. EG me/wes/patrick/whoever is regenerating the test-proxy client.

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
scbedd: Shared recordings offline for review.

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -82,9 +82,22 @@ private static bool ShouldUseOnlyBrokerCredential()
...
+    private const string PlaybackModeEnvVarName = "AZURE_TOKEN_CREDENTIALS";
+    private static bool IsPlaybackMode() => string.Equals(Environment.GetEnvironmentVariable(PlaybackModeEnvVarName), "PlaybackTokenCredential", StringComparison.OrdinalIgnoreCase);
+#endif
 
     private static TokenCredential CreateCredential(string? tenantId, ILogger<CustomChainedCredential>? logger = null)
     {
+#if DEBUG
+        // Short-circuit for playback to avoid any real auth & interactive prompts.
+        if (IsPlaybackMode())
```
hallipr: It looks like this is the only reference to IsPlaybackMode.  If so, I'd move line 121 up to here and reuse its env variable read:

```
        // Check if AZURE_TOKEN_CREDENTIALS is explicitly set
        string? tokenCredentials = Environment.GetEnvironmentVariable(TokenCredentialsEnvVarName);
        bool hasExplicitCredentialSetting = !string.IsNullOrEmpty(tokenCredentials);
```

hallipr: like:

```cs
// Check if AZURE_TOKEN_CREDENTIALS is explicitly set
string? tokenCredentials = Environment.GetEnvironmentVariable(TokenCredentialsEnvVarName);
bool hasExplicitCredentialSetting = !string.IsNullOrEmpty(tokenCredentials);

#if DEBUG
bool isPlaybackMode = string.Equals(tokenCredentials , "PlaybackTokenCredential", StringComparison.OrdinalIgnoreCase);
if (isPlaybackMode)
```

scbedd: 👍 I'll adjust thanks for feedback.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
```diff
@@ -16,27 +18,57 @@ public abstract class CommandTestsBase(ITestOutputHelper output) : IAsyncLifetim
     protected const string TenantNameReason = "Service principals cannot use TenantName for lookup";
 
     protected McpClient Client { get; private set; } = default!;
-    protected LiveTestSettings Settings { get; private set; } = default!;
+    public LiveTestSettings Settings { get; set; } = default!;
```
hallipr: I wouldn't expect anything other than a derived class would need access to the test class's Setting property

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
```diff
@@ -16,27 +18,57 @@ public abstract class CommandTestsBase(ITestOutputHelper output) : IAsyncLifetim
...
     protected StringBuilder FailureOutput { get; } = new();
     protected ITestOutputHelper Output { get; } = output;
 
-    private string[]? _customArguments;
+    public string[]? CustomArguments;
+    public TestMode TestMode = TestMode.Live;
```
hallipr, line 26: Wouldn't this be accessible through `Settings.TestMode`?

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
```diff
@@ -183,4 +228,5 @@ protected virtual async ValueTask DisposeAsyncCore()
     {
         await Client.DisposeAsync().ConfigureAwait(false);
     }
+
```
hallipr: unnecessary line break

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            await Proxy!.AdminClient.AddSanitizersAsync(sanitizers);
+        }
+    }
+
+    public override async ValueTask DisposeAsync()
+    {
+        await StopRecordOrPlayback();
+
+        // On test failure, append proxy stderr for diagnostics.
```
hallipr, line 183: nice

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -0,0 +1,316 @@
...
+            { "x-recording-file", pathToRecording },
+        };
+
+        if (!string.IsNullOrWhiteSpace(assetsPath))
+        {
+            recordOptions["x-recording-assets-file"] = assetsPath;
+        }
+        var bodyContent = BinaryContentHelper.FromObject(recordOptions);
+
+        if (TestMode is TestMode.Playback)
```
hallipr: All of these pattern matches seem odd.   Enum equality usually works without `is` pattern syntax:

```
if (TestMode == TestMode.Playback)
```

hallipr: I guess it works, but `is` is a fairly late addition that's usually used for more complex things like
```
if (user is { Registered: true, Admin: false })
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
scbedd: I'm giving it a sane default. I don't really mind if it is immediately overriden.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
scbedd: dotnet fmt didn't pick it up >.>. Removed.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -0,0 +1,316 @@
...
+
+                foreach (var key in playbackResult.Value.Keys)
+                {
+                    Output.WriteLine($"[Playback] Variable from recording: {key} = {playbackResult.Value[key]}");
+                    TestVariables[key] = playbackResult.Value[key];
+                }
+            }
+            catch (Exception e)
+            {
+                if (Proxy is not null)
```
hallipr: Proxy can't be null here per line 205:

```
 if (Proxy is null)
        {
            throw new InvalidOperationException("Test proxy is not initialized.");
        }
```

core/Azure.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
scbedd: Next commit will have htis.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Helpers/TestEnvironment.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        Live,
+        Record,
+        Playback
+    }
+
+    public static class TestEnvironment
+    {
+        public static bool IsRunningInCi =>
+            string.Equals(Environment.GetEnvironmentVariable("CI"), "true", StringComparison.OrdinalIgnoreCase) || !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("TF_BUILD"));
```
hallipr, line 23: I wonder if devs who know that `$env:TF_BUILD = 'true'` emulates CI mode also know that `$env:TF_BUILD = 'false'` also emulates CI mode

hallipr: or, do we have a consistent pattern for anything other than `TB_BUILD == 'true'`?

scbedd: /shrug.

This isn't even immediately necessary in this PR, but it's also not worth removing as I will definitely use it later.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: I converted them to null comparisons.

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/Azure.Mcp.Tools.KeyVault.LiveTests.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <IsTestProject>true</IsTestProject>
     <OutputType>Exe</OutputType>
+    <NoWarn>$(NoWarn);NU1507;NU1605</NoWarn>
```
hallipr: Were these existing warnings?

scbedd: Ah damnit. Totally unnecessary.

vcolin7:
I haven't had a chance to look at the PR as a whole to understand all the recording and sanitization logic, but I noticed a couple of things in a test recording for `Azure.Mcp.Tools.KeyVault.LiveTests.KeyVaultCommandTests.Should_import_certificate` shared by @scbedd separately:
1. The URL used for the import request (POST) has a double slash before the cert name `//certificateimport4684753782047765310`. Could this be an issue with the test itself? Does the recording work anyway in playback?
    _**Reply from Scott:** Yeah, I believe this is an existing test issue. I don't believe that I'm causing that, just revealing it. I will disable literally all sanitization and confirm this._
2. The request body includes the whole contents of the certificate being imported and its password. This isn't inherently a problem because it's just test data we have already committed to the repo, but I was wondering if there was a way to add custom sanitizers like we do for our SDKs.
    _**Reply from Scott:** Yes, that was one item I put a pin in for myself._

About list operations:

1. I also noticed that the URL for these had a slash at the end followed by a question mark and query parameters (e.g. `https://Sanitized.vault.azure.net/certificates/?api-version=7.5&includePending=false`). I don't have an issue with that as long as it doesn't break playback, but thought you should know 
    _**Reply from Scott:** I don't think I'm adding this._
2. A funny thing about redaction is that the certificate/key/secret names won't be redacted when they're a part of the request URL, but they will in the response body if they include the word `certificate`. For example, one of the certs created by a previous test is listed as `https://Sanitized.vault.azure.net/certificates/Sanitizedimport4611623653797873243` instead of it being something `likecertificateimport4684753782047765310`.
    _**Reply from Scott:** Yeah, you're seeing previous runs._

scbedd:
> I haven't had a chance to look at the PR as a whole to understand all the recording and sanitization logic, but I noticed a couple of things in a test recording for `Azure.Mcp.Tools.KeyVault.LiveTests.KeyVaultCommandTests.Should_import_certificate` shared by @scbedd separately:
> 
> 1. The URL used for the import request (POST) has a double slash before the cert name //certificateimport4684753782047765310. Could this be an issue with the test itself? Does the recording work anyway in playback?
>    _**Reply from Scott:** Yeah, I believe this is an existing test issue. I don't believe that I'm causing that, just revealing it. I will disable literally all sanitization and confirm this._
> 2. The request body includes the whole contents of the certificate being imported and its password. This isn't inherently a problem because it's just test data we have already committed to the repo, but I was wondering if there was a way to add custom sanitizers like we do for our SDKs.
>    _**Reply from Scott:** Yes, that was one item I put a pin in for myself._
> 
> About list operations:
> 
> 1. I also noticed that the URL for these had a slash at the end followed by a question mark and query parameters (e.g. https://Sanitized.vault.azure.net/certificates/?api-version=7.5&includePending=false). I don't have an issue with that as long as it doesn't break playback, but thought you should know
>    _**Reply from Scott:** I don't think I'm adding this._
> 2. A funny thing about redaction is that the certificate/key/secret names won't be redacted when they're a part of the request URL, but they will in the response body if they include the word certificate. For example, one of the certs created by a previous test is listed as https://Sanitized.vault.azure.net/certificates/Sanitizedimport4611623653797873243 instead of  it being something likecertificateimport4684753782047765310
>    _**Reply from Scott:** Yeah, you're seeing previous runs._

Going to do a round of re-recording this morning and see what I can do for the test certificate sanitizer.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/CommandTestsBase.cs
hallipr: I meant the property in general is just a duplicated value. 

Settings sets / owns TestMode.  Then we immediately pull it out to a property beside Settings:
Settings = Deserialise<TestSettings>(json)
TestMode = Settings.TestMode

Then, for all uses, Settings.TestMode is in context,  but we look at TestMode instead.

If you're just adding a convenience accessor, I'd opt for an arrow property:
```suggestion
    public TestMode TestMode => Settings.TestMode;
```

scbedd:
~~Working through new warnings inflicted by my merge from `main`.~~ k g2g.

PR #1080, opened 2025-11-06
Migrate `Keyvault` tool to recorded tests

scbedd:
This PR migrates `keyvault` livetests to recorded tests, adds and assets.json.

Day-to-day development is not yet affected. I need to add a couple more PRs to add stages that will run recorded tests in `CI` before it will mean anything to devs.

### DIFF
--- a/tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
+++ b/tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
@@ -21,7 +21,7 @@ public class KeyVaultCommandTests(ITestOutputHelper output, TestProxyFixture fix
     private readonly KeyVaultTestCertificateAssets _importCertificateAssets = KeyVaultTestCertificates.Load();
 
     public override List<BodyRegexSanitizer> BodyRegexSanitizers => new List<BodyRegexSanitizer>() {
-        // should clear out `kid` hostnames of actual vault names
+        // Sanitizes all hostnames in URLs to remove actual vault names (not limited to `kid` fields)
         new BodyRegexSanitizer(new BodyRegexSanitizerBody() {
           Regex = "(?<=http://|https://)(?<host>[^/?\\.]+)",
           GroupForReplace = "host",


tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
```diff
@@ -1,18 +1,59 @@
...
+                {
+                    Value = _importCertificateAssets.CsrBase64
+                })
+            };
+        }
+    }
+
+    // Disable `$..id` sanitizer as it interferes with Key Vault tests
+    // This will LIKELY become a global disable exclusion in the future.
+    public override List<string> DisabledDefaultSanitizers => new List<string>() { "AZSDK3430", };
```
scbedd: Going to move this to the central definition. Sanitizing `$..id` which is what `AZSDK3430` is just breaks most of the tests by default.

scbedd:
After running recorded tests, we're seeing a new failure on `MacOS` only. `windows` and `linux` both passing recorded tests. Exploring for solution.

tools/Azure.Mcp.Tools.KeyVault/tests/Azure.Mcp.Tools.KeyVault.LiveTests/KeyVaultCommandTests.cs
```diff
@@ -1,18 +1,55 @@
...
+
+    public override List<BodyRegexSanitizer> BodyRegexSanitizers => new List<BodyRegexSanitizer>() {
+        // should clear out `kid` hostnames of actual vault names
+        new BodyRegexSanitizer(new BodyRegexSanitizerBody() {
+          Regex = "(?<=http://|https://)(?<host>[^/?\\.]+)",
+          GroupForReplace = "host",
+        })
+    };
+
+    public override List<BodyKeySanitizer> BodyKeySanitizers
```
vcolin7, line 31: I'm a bit unfamiliar with how this class is used in the `RecordedCommandTestsBase` class. Is this supposed to sanitize the contents of the body in both requests and responses? If so, I noticed the recordings are not sanitizing `value`, `cer`, nor `csr`.

scbedd: This is intended. These certificate values are actually checked by the code, so there needs to be a value there. These values will _never_ change from the statics defined in fake-pfx

Issue #540, opened 2025-09-23
Enable Record/Playback pathway for


scbedd:
Last week, I got started on looking into record/playback options. Originally, we intended on honoring HTTP_PROXY and simply recording all traffic through the `test-proxy`. However, we have found a gigantic implementation gap there, due to the fact that to record HTTPS traffic through a proxy, the proxy itself needs to MITM the HTTPS traffic. Ala `Fiddler`, `mitmproxy`, or `dev-proxy`. With all of those options, we face a TON of pain as far as transferring other concepts that the `test-proxy` already handles for us, like `sanitization`, `matching`, and `assets storage`. There was a much longer discussion, but [we judged](https://teams.microsoft.com/l/meetingrecap?driveId=b%21K_ngyKPmj0Wv1DlJn55lXJFqSRrLp8hHnUS42MEueSHoJJZxa6mYSblsBkq8w1L1&driveItemId=01LJCU2RCASQ47ASZV6ZEZ2ODVBMZ763ET&sitePath=https%3A%2F%2Fmicrosoft-my.sharepoint.com%2F%3Av%3A%2Fp%2Fscbedd%2FEUCUOfBLNfZJnTh1CzP_bJMBKFWRIEDDgwzOcGaFLG7Sdg&fileUrl=https%3A%2F%2Fmicrosoft-my.sharepoint.com%2Fpersonal%2Fscbedd_microsoft_com%2FDocuments%2FRecordings%2FDiscuss%2520MCP%2520Proxy%2520Options-20250919_123122-Meeting%2520Recording.mp4%3Fweb%3D1&iCalUid=040000008200E00074C5B7101A82E008000000009C4F952FF628DC010000000000000000100000003CE22E69BDBEFC45B266D67CB912002B&threadId=19%3Ameeting_NmU5NzU0YjQtNTUxNS00OTE0LTgwYmMtMTE1MGUxNmZjMGFi%40thread.v2&organizerId=12e4dcc7-6a8f-4206-88d6-bc961596e8c9&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&callId=fff66b2d-b9fb-44a6-bfe3-6aaf34952fec&threadType=meeting&meetingType=Scheduled&subType=RecapSharingLink_RecapCore) that the biggest bang for buck is to enable URL rewrite at the core `HttpClient` level via DelegateHandlers. 

[...]

PR #1117, opened 2025-11-10
Proxy Integration Test Framework + Initial Functionality Tests + Session/Individual Test Matcher Support

scbedd:
Related to #540

@hallipr , I chose to place these tests within the `LiveTests` csproj. This is intentional, because due to what they're testing, I consider them `Recorded` tests. So they should be run against the livetests csproj with no livetestsettings present. 

For now they will run as part of the live test suite.

todo: 

- [x] Accept feedback from reviewers
- [x] Fix failing proxy tests
- [x] ~~I am at a loss to explain how my changes have broken the ServerCommandTests.~~ broken by other main pr
- [x] Cleanup recording jsons properly

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.LiveTests/RecordingFramework/RecordedCommandTestsBaseTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.LiveTests/RecordingFramework/RecordedCommandTestsBaseTests.cs
@@ -1,15 +1,18 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Reflection;
 using System.Text.Json;
 using Azure.Mcp.Tests.Client;
+using Azure.Mcp.Tests.Client.Attributes;
 using Azure.Mcp.Tests.Client.Helpers;
 using Azure.Mcp.Tests.Generated.Models;
 using Azure.Mcp.Tests.Helpers;
 using Microsoft.Extensions.FileSystemGlobbing;
 using NSubstitute;
 using Xunit;
+using Xunit.v3;
 
 namespace Azure.Mcp.Core.LiveTests.RecordingFramework;
 
@@ -39,9 +42,15 @@ public sealed class RecordedCommandTestsBaseTest : IAsyncLifetime
         Assert.Equal("sampleValue", variablesElement.GetProperty("sampleKey").GetString());
     }
 
+    [CustomMatcher(IgnoreQueryOrdering = true, CompareBodies = true)]
     [Fact]
     public async Task PerTestMatcherAttributeAppliesWhenPresent()
     {
+        var activeMatcher = GetActiveMatcher();
+        Assert.NotNull(activeMatcher);
+        Assert.True(activeMatcher!.CompareBodies);
+        Assert.True(activeMatcher.IgnoreQueryOrdering);
+
         DefaultHarness = new RecordedCommandTestHarness(CollectedOutput, Fixture)
         {
             DesiredMode = TestMode.Record,
@@ -64,6 +73,36 @@ public sealed class RecordedCommandTestsBaseTest : IAsyncLifetime
         await playbackHarness.DisposeAsync();
 
         CollectedOutput.Received().WriteLine(Arg.Is<string>(s => s.Contains($"Applying custom matcher to recordingId \"{recordingId}\"")));
+    }
+
+    [Fact]
+    public void CustomMatcherAttributeClearsAfterExecution()
+    {
+        var attribute = new CustomMatcherAttribute(compareBody: true, ignoreQueryordering: true);
+        var xunitTest = Substitute.For<IXunitTest>();
+        var methodInfo = typeof(RecordedCommandTestsBaseTest).GetMethod(nameof(CustomMatcherAttributeClearsAfterExecution))
+            ?? throw new InvalidOperationException("Unable to locate test method for CustomMatcherAttribute verification.");
+
+        attribute.Before(methodInfo, xunitTest);
+        try
+        {
+            var active = GetActiveMatcher();
+            Assert.Same(attribute, active);
+            Assert.True(active!.CompareBodies);
+            Assert.True(active.IgnoreQueryOrdering);
+        }
+        finally
+        {
+            attribute.After(methodInfo, xunitTest);
+        }
+
+        Assert.Null(GetActiveMatcher());
+    }
+
+    private static CustomMatcherAttribute? GetActiveMatcher()
+    {
+        var method = typeof(CustomMatcherAttribute).GetMethod("GetActive", BindingFlags.NonPublic | BindingFlags.Static);
+        return (CustomMatcherAttribute?)method?.Invoke(null, null);
     }
 
     [Fact]
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
@@ -3,7 +3,6 @@
 
 using System.ClientModel;
 using System.ClientModel.Primitives;
-using System.Reflection;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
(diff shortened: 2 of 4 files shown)


core/Azure.Mcp.Core/tests/Azure.Mcp.Core.LiveTests/SessionRecords/Azure.Mcp.Core.LiveTests.RecordedCommandTestsBaseLifecycleTests.ApplyAttributeMatcherSettings_UsesCustomMatcherWhenAttributePresent.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Entries": [],
+  "Variables": {
```
scbedd: will clean these up.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -84,17 +98,22 @@ public virtual void RegisterVariable(string name, string value)
     }
 
     // used to resolve a recording "path" given an invoking test
-    private static readonly RecordingPathResolver _pathResolver = new();
+    protected static readonly RecordingPathResolver _pathResolver = new();
```
hallipr: The naming here would change when it becomes protected:

```
    protected static readonly RecordingPathResolver PathResolver = new();
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -84,17 +98,22 @@ public virtual void RegisterVariable(string name, string value)
...
 
     protected virtual bool IsAsync => false;
 
     // todo: use this when we have versioned tests to run this against.
     protected virtual string? VersionQualifier => null;
 
+    protected override async ValueTask LoadSettingsAsync()
```
hallipr, line 108: Any reason to override with just a call to base?

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -107,6 +126,77 @@ public override async ValueTask InitializeAsync()
...
+        {
+            return;
+        }
+
+        // Use reflection to find the test method on the derived class
+        var testMethod = GetType().GetMethod(methodName);
+        if (testMethod == null)
+        {
+            return;
+        }
```
hallipr: This seems fairly indirect.  The docs show TestContext.TestMethod having a `Method` property with type `MethodInfo`. is that not the case?
https://api.xunit.net/v3/3.2.0/Xunit.v3.IXunitTestMethod.html

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -107,6 +126,77 @@ public override async ValueTask InitializeAsync()
...
+        var matcherSb = new StringBuilder();
+        matcherSb.Append($"CompareBodies={matcher.CompareBodies}, IgnoreQueryOrdering={matcher.IgnoreQueryOrdering}");
+        if (!string.IsNullOrEmpty(matcher.IgnoredHeaders))
+        {
+            matcherSb.Append($", IgnoredHeaders={matcher.IgnoredHeaders}");
+        }
+        if (!string.IsNullOrEmpty(matcher.ExcludedHeaders))
+        {
+            matcherSb.Append($", ExcludedHeaders={matcher.ExcludedHeaders}.");
+        }
```
hallipr, line 172: Would this be a useful ToString override for CustomDefaultMatcher?

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.LiveTests/RecordingFramework/RecordedCommandTestsBaseTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    [Fact]
+    public async Task ProxyRecordProducesRecording()
+    {
+        await DefaultHarness!.InitializeAsync();
+
+        Assert.NotNull(Fixture.Proxy);
+        Assert.False(string.IsNullOrWhiteSpace(Fixture.Proxy!.BaseUri));
+
+        DefaultHarness!.RegisterVariable("sampleKey", "sampleValue");
+        await DefaultHarness!.DisposeAsync();
```
hallipr, line 36: And the default harness needs to be disposed before we assert/read the file?

hallipr: just thinking that c# has `await using` and IDisposable consumers don't usually invoke Dispose/DisposeAsync directly

```
await using (DefaultHarness) {
}
```

Or just `await using DefaultHarness;` if we don't care that DisposeAsync is called at the end of the method

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
```diff
@@ -19,6 +23,12 @@ public abstract class RecordedCommandTestsBase(ITestOutputHelper output, TestPro
 {
     protected TestProxy? Proxy { get; private set; } = fixture.Proxy;
 
+    private string RecordingId { get; set; } = string.Empty;
```
hallipr: If `RecordingId` were protected get, your test harness would have access to it without reflection

hallipr:
```suggestion
    protected string RecordingId { get; private set; } = string.Empty;
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.LiveTests/RecordingFramework/RecordedCommandTestsBaseTests.cs
scbedd: > And the default harness needs to be disposed before we assert/read the file?

Disposing the default harness is effectively telling the test to stop, which writes the test file. Then we examine that test file to be certain that content that should be in it IS in it.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: Yes, but CustomDefaultMatcher is a generated class. I could do it with an extension class but there is no real reason to do that. When I have more than one call site I'm good with that 👍

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: Yeah. Because if I call base.LoadSettingsAsync() like I _did_ ,there is no way to disable the default behavior.

If I make it an overridden function, then override _that_ in recordedCommandTestHarness, everything works the way I expect!

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: Change incoming.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: It's not populated properly in my experience, no. I'll try again?

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: Good call.

core/Azure.Mcp.Core/tests/Azure.Mcp.Tests/Client/RecordedCommandTestsBase.cs
scbedd: image

And TestMethod itself doesn't have a `GetCustomAttribute` to get the matcher.

hallipr (approved):

PR #1212, opened 2025-11-18
Stabilize the behavior and reliability of ExternalProcessService

anuchandy:
## What does this PR do?

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
+++ b/core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
@@ -36,6 +36,21 @@ public class ExternalProcessService(ILogger<ExternalProcessService> logger) : IE
     /// <returns>
     /// A <see cref="ProcessResult"/> containing the exit code, stdout, stderr, and the full command.
     /// </returns>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// Thrown when <paramref name="operationTimeoutSeconds"/> is less than or equal to zero.
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="fileName"/> is null or empty.
+    /// </exception>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when the process fails to start.
+    /// </exception>
+    /// <exception cref="TimeoutException">
+    /// Thrown when the operation (process execution and stream draining) exceeds <paramref name="operationTimeoutSeconds"/>.
+    /// </exception>
+    /// <exception cref="OperationCanceledException">
+    /// Thrown when <paramref name="cancellationToken"/> is triggered before the operation completes.
+    /// </exception>
     /// <remarks>
     /// <para><b>Timeout vs. cancellation:</b></para>
     /// <para>


core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,236 @@
...
+                $"{executablePath} {arguments}");
+        }
+        finally
+        {
+            // ProcessStreamReader.Dispose unsubscribes 'DataReceived' event handlers from the Process.
+            // Process.Dispose releases OS resources.
+        }
+    }
+
+    public JsonElement ParseJsonOutput(ProcessResult result)
```
anuchandy, line 196: The `ParseJsonOutput` method, and the types it uses `ParseError`, `ParseOutput` are existing (they just look new in this PR but were already part of the`ExternalProcessService`.)

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,236 @@
...
+    ///       signaled, and the process has not yet exited, a best-effort attempt is made to terminate the process
+    ///       (and its tree) before rethrowing the original <see cref="OperationCanceledException"/>.
+    ///     </description>
+    ///   </item>
+    /// </list>
+    /// <para>
+    /// This separation avoids the common ambiguity where timeouts are implemented by canceling the caller’s token.
+    /// Here: <see cref="TimeoutException"/> always means timeout; <see cref="OperationCanceledException"/>
+    /// always means external cancellation.
+    /// </para>
```
anuchandy, line 81: I kept the original approach of taking operation timeout and external cancellation parameters separately, as this aligns with established .NET guidance and avoids the ambiguity that comes from conflating the two.

1. Timeout and cancellation represent different semantics
   * Timeout is a failure of the operation itself - external process didn’t finish in time.
   * Cancellation is a request initiated by the caller or somewhere in the async-pipeline to stop the operation.

Because timeout and cancellation reflect fundamentally different states, they are surfaced as different exception types (`TimeoutException` vs `OperationCanceledException`).

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,236 @@
...
+    ///       captures the original exception that made the exit status <c>Indeterminate</c>.
+    ///     </description>
+    ///   </item>
+    /// </list>
+    ///
+    /// <para>
+    /// These fields are useful for diagnosing cases such as processes failing to terminate, OS-level handle
+    /// failures, corrupted process state, or unexpected platform behavior. All diagnostic data is attached
+    /// without modifying the original exception type, message, stack trace, or cancellation semantics.
+    /// </para>
```
anuchandy, line 124: Timeout or cancellation is always the "primary" failure of the `ExecuteAsync` _operation_. Any errors from `HasExited` or `Kill` are "secondary" cleanup issues as a side effect of "primary" failure. These are rare (mainly Win32Exception from the OS) and don’t represent the real cause of the _operation_ failing.

Instead of throwing those cleanup exceptions, we attach them to the `.Data` of the `TimeoutException` or `OperationCanceledException`. This preserves the correct exception semantics (timeout vs. cancellation) while still surfacing full diagnostics if needed.

Additionally, cleanup failures are also logged at "Warning" level, so they’re visible - especially in expected stdio use cases, VS-Code - where such logs surface directly but also without masking the actual timeout or cancellation that may returned to Copilot/LLM.

anuchandy (commented):
adding notes

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
 
-    internal record ParseError(
-        int ExitCode,
-        string Error,
-        string Command
-    );
-
-    internal record ParseOutput([property: JsonPropertyName("output")] string Output);
-
-    public void SetEnvironmentVariables(IDictionary<string, string> variables)
```
anuchandy, line 143: This API isn’t used anywhere, but more importantly, it’s not a good design to allow shared, mutable environment-variable state across ExecuteAsync. It’s safer and clearer to pass environment variables directly as parameters to ExecuteAsync, and this PR makes that change.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+            catch (Exception ex)
+            {
+                // Unsubscribing the handlers are the best-effort cleanup step; log and swallow any exceptions to avoid disposal throwing.
+                _logger.LogDebug(
+                    ex,
+                    "Unsubscribe from {StreamType} stream during disposal was skipped. Process: {ProcessName}, PID: {Pid}.",
+                    StreamType,
+                    process.SafeName(),
+                    process.SafeId());
+            }
```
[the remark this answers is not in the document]
anuchandy: By design - and as documented - this is the dispose path, and we don’t throw here.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
-                result.Error,
-                result.Command
-            );
-            return JsonSerializer.SerializeToElement(error, ServicesJsonContext.Default.ParseError);
+            return process.Id;
+        }
+        catch
+        {
+            return -1;
         }
```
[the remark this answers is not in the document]
anuchandy: by design

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+    ///
+    /// <param name="process">The process to terminate.</param>
+    /// <param name="logger">Logger for diagnostic messages.</param>
+    ///
+    /// <remarks>
+    /// The returned <see cref="ExitCheckResult"/> reflects the state observed <em>before</em> the kill attempt.
+    /// If <see cref="ExitStatus.NotExited"/> is reported, this method calls <see cref="Process.Kill(bool)"/>.
+    ///
+    /// Kill may throw:
+    /// <list type="bullet">
```
vukelich, line 640: Try out [https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d35-exception](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d35-exception)

tools/Azure.Mcp.Tools.Extension/src/Commands/AzCommand.cs
```diff
@@ -138,7 +138,7 @@ private async Task<bool> AuthenticateWithAzureCredentialsAsync(IExternalProcessS
             var azPath = FindAzCliPath() ?? throw new FileNotFoundException("Azure CLI executable not found in PATH or common installation locations. Please ensure Azure CLI is installed.");
 
             var loginCommand = $"login --service-principal -u {credentials.ClientId} -p {credentials.ClientSecret} --tenant {credentials.TenantId}";
-            var result = await processService.ExecuteAsync(azPath, loginCommand, 60, cancellationToken: cancellationToken);
+            var result = await processService.ExecuteAsync(azPath, loginCommand, null, 60, cancellationToken: cancellationToken);
```
hallipr: Named arguments would be helpful here:

```suggestion
            var result = await processService.ExecuteAsync(azPath, loginCommand,
                operationTimeoutSeconds: 60,
                cancellationToken: cancellationToken);
```

tools/Azure.Mcp.Tools.Extension/src/Commands/AzCommand.cs
```diff
@@ -180,7 +180,7 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
             await AuthenticateWithAzureCredentialsAsync(processService, _logger, cancellationToken);
 
             var azPath = FindAzCliPath() ?? throw new FileNotFoundException("Azure CLI executable not found in PATH or common installation locations. Please ensure Azure CLI is installed.");
-            var result = await processService.ExecuteAsync(azPath, command, _processTimeoutSeconds, cancellationToken: cancellationToken);
+            var result = await processService.ExecuteAsync(azPath, command, null, _processTimeoutSeconds, cancellationToken: cancellationToken);
```
hallipr: named arguments here too:

```suggestion
            var result = await processService.ExecuteAsync(azPath, command,
                operationTimeoutSeconds: _processTimeoutSeconds,
                cancellationToken: cancellationToken);
```

tools/Azure.Mcp.Tools.Extension/src/Commands/AzqrCommand.cs
```diff
@@ -103,7 +103,7 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
             command += " --json";
 
             var processService = context.GetService<IExternalProcessService>();
-            var result = await processService.ExecuteAsync(azqrPath, command, _processTimeoutSeconds, cancellationToken: cancellationToken);
+            var result = await processService.ExecuteAsync(azqrPath, command, null, _processTimeoutSeconds, cancellationToken: cancellationToken);
```
hallipr: and here:

```suggestion
            var result = await processService.ExecuteAsync(azqrPath, command,
                operationTimeoutSeconds: _processTimeoutSeconds,
                cancellationToken: cancellationToken);
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+        /// Cancellation of the overall external process operation is handled by <c>ExecuteAsync</c>,
+        /// which terminates the process on timeout or external cancellation. Once the process exits
+        /// (naturally or via kill), the stream read completes automatically.
+        ///
+        /// If <c>ProcessStreamReader</c> is ever used outside the normal <c>ExecuteAsync</c> workflow,
+        /// <see cref="Dispose"/> provides a safety net by completing the task even if the final
+        /// <c>e.Data == null</c> callback was never raised.
+        /// </remarks>
+        /// <returns>A task that completes when the stream has fully drained.</returns>
+        public Task<string> StartReading()
```
hallipr:
```suggestion
        public Task<string> StartReadingAsync()
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
         {
-            if (!process.HasExited)
+            ObjectDisposedException.ThrowIf(disposed, this);
+
+            if (readingStarted)
```
hallipr: Should we lock first?

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
hallipr: If this returns the entire contents of the stream as a string, but only after the EOF `null` has been reached, wouldn't `ReadToEndAsync()` be a better name?

hallipr: like: [StreamReader.ReadToEndAsync](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoendasync?view=net-9.0#system-io-streamreader-readtoendasync\(system-threading-cancellationtoken\))

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+    }
 
-        process.ErrorDataReceived += (sender, e) =>
+    private TimeoutException HandleTimeout(Process process, TimeSpan timeout, string executablePath, string arguments)
+    {
+        // Get the pre-kill exit state and any kill error (if termination was attempted).
+        (ExitCheckResult exitCheck, Exception? killException) = process.TryKill(_logger);
+        string command = $"{executablePath} {arguments}";
+
+        TimeoutException tex;
```
hallipr: no need to abbreviate here...
```suggestion
        TimeoutException exception;
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+        tex.Data["ProcessId"] = process.SafeId();
+        tex.Data["ProcessExitStatus"] = exitCheck.Status.ToString();
 
-        process.Start();
-        process.BeginOutputReadLine();
-        process.BeginErrorReadLine();
+        return tex;
+    }
+
+    private void HandleCancellation(Process process, string executablePath, string arguments, OperationCanceledException oce)
```
hallipr: no need to abbreviate:

```suggestion
    private void HandleCancellation(Process process, string executablePath, string arguments, OperationCanceledException exception)
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+    private sealed class ProcessStreamReader : IDisposable
+    {
+        private readonly Process process;
+        private readonly bool isErrorStream;
+        private readonly ILogger<ExternalProcessService> _logger;
+        private readonly StringBuilder buffer = new();
+        private readonly TaskCompletionSource<string> tcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly DataReceivedEventHandler handler;
+        private bool readingStarted;
+        private bool disposed;
```
hallipr: The field naming needs to be consistent and match .editorconfig style rules:

```suggestion
        private readonly Process _process;
        private readonly bool _isErrorStream;
        private readonly ILogger<ExternalProcessService> _logger;
        private readonly StringBuilder _buffer = new();
        private readonly TaskCompletionSource<string> _tcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
        private readonly DataReceivedEventHandler _handler;
        private bool _readingStarted;
        private bool _disposed;
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+
+        public ProcessStreamReader(Process process, bool isErrorStream, ILogger<ExternalProcessService> logger)
         {
-            await waitTask;
+            this.process = process ?? throw new ArgumentNullException(nameof(process));
```
hallipr: This repo prefers `_process` to `this.process`

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+        catch (System.ComponentModel.Win32Exception e)
+        {
+            // Failure to read status (access denied, invalid handle, etc.).
+            return new ExitCheckResult(ExitStatus.Indeterminate, CheckException: e);
+        }
+        catch (NotSupportedException e)
+        {
+            // Remote process or unsupported scenario – not a valid case for Azure MCP Server.
+            return new ExitCheckResult(ExitStatus.Indeterminate, CheckException: e);
+        }
```
hallipr: full exception arg names would make this cleaner

```suggestion
        catch (InvalidOperationException checkException)
        {
            // Official docs: "No process is associated with this object." - treat as "already gone".
            logger.LogDebug(
            checkException,
            "Process.HasExited reported no associated process. Treating as already exited. " +
            "Process: {ProcessName}, PID: {Pid}", process.SafeName(), process.SafeId());
            return new ExitCheckResult(ExitStatus.Exited, CheckException: null);
        }
        catch (System.ComponentModel.Win32Exception checkException)
        {
            // Failure to read status (access denied, invalid handle, etc.).
            return new ExitCheckResult(ExitStatus.Indeterminate, checkException);
        }
        catch (NotSupportedException checkException)
        {
            // Remote process or unsupported scenario – not a valid case for Azure MCP Server.
            return new ExitCheckResult(ExitStatus.Indeterminate, checkException);
        }
```

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,236 @@
...
+            // Normal completion: the process has exited, and stdout/stderr have fully drained.
+            return new ProcessResult(
+                process.ExitCode,
+                stdout,
+                stderr,
+                $"{executablePath} {arguments}");
+        }
+        finally
+        {
+            // ProcessStreamReader.Dispose unsubscribes 'DataReceived' event handlers from the Process.
```
hallipr: I may be missing something.  An empty Finally block and no Catch block makes the whole try/finally moot.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +241,475 @@ public async Task<ProcessResult> ExecuteAsync(
...
+            handler = (_, e) =>
+            {
+                if (e.Data is null)
+                {
+                    // Stream closed – finalize accumulated output and complete.
+                    tcs.TrySetResult(buffer.ToString());
+                }
+                else
+                {
+                    buffer.AppendLine(e.Data);
```
tmeschter: I'm not convinced this approach solves the problem. As best I can tell, in the event of a cancellation or timeout this call to `StringBuilder.AppendLine(...)` could happen concurrently with a call to `StringBuilder.ToString()`, and `StringBuilder` does not support concurrent access by multiple threads.

tmeschter: Fundamentally, `buffer` may be accessed concurrently by multiple threads and access needs to be controlled with some sort of synchronization primitive.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: Good suggestion, let's rename to `ReadToEndAsync`

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: Thanks Patrick. For this internal, tightly-controlled usage inside `ExternalProcessService`, where `ReadToEndAsync` is only ever called once per instance as part of the `ExecuteAsync` flow, the current pattern is fine.

I would absolutely worry about this if we were building a general-purpose, thread-safe public stream reader meant for arbitrary concurrent callers - but that’s not our scenario here.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: good catch, will update the names

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: You’re right - the only reason I added the empty finally block was to document that disposal occurs. I’ll remove the no-code finally and move disposal document at the end of the method.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: Thanks, Tom - I agree with you on the race.

In the context of this `ExternalProcessService`, I think we can avoid introducing locks. There are essentially two flows:
1. Success path: the process exits and the stream is fully drained. In this case, the handler completes the task with the final string before `Dispose` runs, so the handler and `Dispose` never touch the buffer concurrently.
2. Failure path: we throw a timeout or cancellation exception. In this case, we don’t care about partially buffered data because the caller receives an exception rather than a `ProcessResult`. So instead of calling `ToString()` on the shared `_buffer` in `Dispose`, we can safely complete the task with `string.Empty`. `TrySetResult` on the `TaskCompletionSource` is thread-safe, and we don’t care whether the handler or Dispose wins `TrySetResult`  race.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: done

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: done

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,236 @@
...
+        if (result.ExitCode != 0)
+        {
+            var error = new ParseError(
+                result.ExitCode,
+                result.Error,
+                result.Command);
+
+            return JsonSerializer.SerializeToElement(
+                error,
+                ServicesJsonContext.Default.ParseError);
```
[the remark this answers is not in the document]
anuchandy: it compiles

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: updated based on my previous comment

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -3,21 +3,230 @@
...
+        using Process process = CreateProcess(executablePath, arguments, environmentVariables);
+        using ProcessStreamReader stdoutReader = new(process, isErrorStream: false, logger);
+        using ProcessStreamReader stderrReader = new(process, isErrorStream: true, logger);
+
+        if (!process.Start())
+        {
+            throw new InvalidOperationException($"Failed to start process: {executablePath}");
+        }
+
+        Task<string> stdoutTask = stdoutReader.ReadToEndAsync();
```
hallipr, line 150: What happens to these tasks if we don't await them?   In the timeout or cancellation case, what happens to the task?  Is it still scheduled?

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -26,130 +235,486 @@ public async Task<ProcessResult> ExecuteAsync(
             throw new FileNotFoundException($"Executable not found at path: {executablePath}");
```
weshaggard: Is this necessary? It prevent tools to be passed by name so the process class can resolve them via the path. It requires us to duplicate the path resolution in a few places.

https://github.com/microsoft/mcp/blob/8145156f4cfbb22b5fc9d57796e0ff164c71e1b1/tools/Azure.Mcp.Tools.Extension/src/Commands/AzCommand.cs#L66
https://github.com/microsoft/mcp/blob/8145156f4cfbb22b5fc9d57796e0ff164c71e1b1/tools/Azure.Mcp.Tools.Extension/src/Commands/AzqrCommand.cs#L135
https://github.com/microsoft/mcp/blob/8145156f4cfbb22b5fc9d57796e0ff164c71e1b1/servers/Azure.Mcp.Server/tests/Azure.Mcp.Server.UnitTests/Infrastructure/CommandMetadataSyncTests.cs#L56

hallipr (approved):

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
hallipr: I can't remember if an incomplete task stays on the scheduler and if its references keep objects from being garbage collected.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: The `File.Exists` check here was just carried over from the original implementation. I agree with you - a general-purpose process runner like this shouldn’t require a full path. It should support both full paths and command names, where in case of command names letting .NET resolve them (via PATH). I'll remove this validation.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: Hi Patrick, the `stdoutTask` and `stderrTask` from `ProcessStreamReader` instances aren’t fire-and-forget. They’re part of aggregation that we await with a timeout 

```C#
Task.WhenAll(exitTask, stdoutTask, stderrTask)
```

On timeout or cancellation:
1. `Dispose()` is called on both stream readers.
2.  if EOF is never observed, `Dispose()` force-completes the TCS with string.Empty as a safety net.

So once our stream reader shuts down and the TCS (an internal state of reader) completes, these tasks transitions and the entire reader (and it's state backing task) becomes collectible.

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: I’ve renamed the parameter and removed the file-path existence check. Unfortunately, the [official](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.filename?view=net-10.0) .NET documentation for `ProcessStartInfo.FileName` doesn’t describe how executable path lookup works across OS (and .NET seems to defer to that OS-level behavior). The underlying Win32 API (`CreateProcessW`) does document how it is for Windows - https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

So I kept the doc simple - "_.. the operating system resolves the executable using its standard search rules..._"

core/Azure.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
anuchandy: thanks!, updated to follow this guideline for exception documentation.

hallipr (approved):

Issue #1132, opened 2025-11-10
[BUG] FastTranscriptionRecognizer setting default Authorization outbound header is not remote MCP server safe


vukelich:
### Describe the bug

FastTranscriptionRecognizer is a singleton DI service that has an `HttpClient` field that is reused and uses

```csharp
_httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", accessToken.Token);
```

In the remote MCP server with on-behalf-of outbound authentication, this presents a race condition that could allow two different MCP clients to use the same outbound access token which could over-privilege one of the callers.

### Expected behavior

FastTranscriptionRecognizer should get an instance of `HttpClient` from `IHttpClientService` or `IHttpClientFactory` (choice requires investigation) as per [.NET best practices](https://learn.microsoft.com/en-us/dotnet/core/extensions/httpclient-factory), and it must send the authorization header only for this single request without causing side-effects to other `HttpClient` instances.

### Actual behavior

Possible security bug

### Reproduction Steps

N/A

kvenkatrajan:
Closing this out - @g2vinay has completed with - https://github.com/microsoft/mcp/pull/1141

PR #1218, opened 2025-11-19
Optionally Inject HttpClient to `BaseAzureService`

scbedd:
This is method I'd use with less refactoring.

The _other_ option is we add `IHttpClientService` to constructor for `BaseAzureService`, then utilize said HttpClientService during creation of `GetArmClient()`.

I have tested this implementation with [AppConfig](https://github.com/microsoft/mcp/compare/main...scbedd:migrate-appconfig-injected-option?expand=1) and the direct usage of `ExecuteResourceQueryAsync` (where the appconfigclient isn't new-ed up) successfully records with this code.

Before I go on a refactor trip of the entire repo injecting `IHttpClientService`, I wanted to get the opinion of folks here.


core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -148,6 +149,9 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
         {
             using var host = CreateHost(options);
 
+            var httpClientService = host.Services.GetRequiredService<IHttpClientService>();
+            BaseAzureService.InitializeHttpClientTransport(httpClientService);
```
vukelich, line 153: Can you please detail the goal of this PR and the changes you want to make? If I understand the motivation, I can help us make sure we implement something safe for local and remote MCP server scenarios.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -16,6 +19,9 @@ public abstract class BaseAzureService
     private static string? s_userAgent;
     private static volatile bool s_initialized = false;
     private static readonly object s_initializeLock = new();
+    private static HttpPipelineTransport? s_sharedHttpTransport;
```
vukelich, line 22: For secure code, we need to avoid any state that might capture details of a particular user, especially in our singleton types injected in the DI container and double-y so for `static`s. 

We should be moving in the direction of the .NET best practices for using `HttpClient` as outlined in [IHttpClientFactory with .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/httpclient-factory)

vukelich: An example of reuse that could lead to issues was the recently fixed #1132.

core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
scbedd: Effectively, we need the the ArmClient exposed in BaseAzureService to get an inserted handler to redirect traffic for record/playback.

[The client needs this code run against it's transport.](https://github.com/microsoft/mcp/blob/1dfe5ca5d52335c09e05fb0c5f524d6c97edbe77/core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs#L71)

scbedd:
Closing in favor of #1239

PR #1239, opened 2025-11-24
Populate `HttpClientFactory`, utilize in `BaseAzureService`

scbedd:
Following up original closed PR #1218 .

Hey @vukelich please take a look when you have a couple minutes? Not entirely done yet but not much more to work through.

This PR is addressing the fact that the `GetArmClient()` implementation in `BaseAzureService` ignores the recording configuration that `HttpClientService` _doesn't_. We do this by:

- [x] Adds a `client configurator` that is basically copy/pasta of HttpClientService
- [x] Injects a configured IHttpClientFactory into the `azure mcp server` utilizing said configurator.
- [x] Injects `IHttpClientFactory` into `TenantService`
- [x] Utilizes new TenantService.CreateClient() while creating the base ArmClient within `BaseAzureService` 
- [x] I have verified that an AppConfig livetest can successfullly record with my code changes
- [x] ~~I _must_ be missing something with the default client from AddHttpClient(). I should be able to override the default without actually naming it right? Seems super weird that I can't.~~ Yep I was ignorant.
- [x] Fix failing unit test.
- [x] Configure `IHttpClientFactory` in Azure.Fabric and Azure.Template servers so that the changes to `TenantService` don't melt the world. We only don't see failures because of the fact that we're not running those servers.
- [x] [Verify unbroken fabric server](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5610163&view=results)

Todo:

- [x] ~~I am going to Inject `IHttpClientFactory` into `HttpClientService` and utilize the _clientFactory.CreateClient() over the existing CreateClientInternal(). From there should be easy to entirely replace IHttpClientService in favor of the clientFactory created client.~~ Very next PR. I want to unblock record/playback.
- [x] Accept feedback from @hallipr / @vukelich / whoever else
  - First round accepted. Need to deal with HttpClientService / the re-instantiation of the useragent string now.
  - Secound round also accepted.

### DIFF
--- a/core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
+++ b/core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
@@ -20,6 +20,8 @@ public static class HttpClientFactoryConfigurator
     private static readonly string s_framework;
     private static readonly string s_platform;
 
+    private static string? s_userAgent = null;
+
     static HttpClientFactoryConfigurator()
     {
         var assembly = typeof(HttpClientService).Assembly;
@@ -144,6 +146,7 @@ public static class HttpClientFactoryConfigurator
 
     private static string BuildUserAgent(string transport)
     {
-        return $"azmcp/{s_version} azmcp-{transport}/{s_version} ({s_framework}; {s_platform})";
+        s_userAgent ??= $"azmcp/{s_version} azmcp-{transport}/{s_version} ({s_framework}; {s_platform})";
+        return s_userAgent;
     }
 }
--- a/servers/Azure.Mcp.Server/src/Program.cs
+++ b/servers/Azure.Mcp.Server/src/Program.cs
@@ -195,15 +195,11 @@ internal class Program
         services.AddSingleton<ISubscriptionService, SubscriptionService>();
         services.AddSingleton<CommandFactory>();
 
-        services.AddHttpClient();
-        // customize the default client with Proxy, Record/Playback, and User-Agent defaults
-        services.ConfigureDefaultHttpClient();
-
         // !!! WARNING !!!
         // stdio-transport-specific implementations of ITenantService and ICacheService.
         // The http-traport-specific implementations and configurations must be registered
         // within ServiceStartCommand.ExecuteAsync().
-        services.AddAzureTenantService();
+        services.AddAzureTenantService(addUserAgentClient: true);
         services.AddSingleUserCliCacheService();
 
         foreach (var area in Areas)
--- a/servers/Fabric.Mcp.Server/src/Program.cs
+++ b/servers/Fabric.Mcp.Server/src/Program.cs
@@ -148,8 +148,6 @@ internal class Program
         services.AddSingleton<ISubscriptionService, SubscriptionService>();
         services.AddSingleton<CommandFactory>();
 
-        services.AddHttpClient();
-
         // !!! WARNING !!!
         // stdio-transport-specific implementations of ITenantService and ICacheService.
         // The http-traport-specific implementations and configurations must be registered
--- a/servers/Template.Mcp.Server/src/Program.cs
+++ b/servers/Template.Mcp.Server/src/Program.cs
@@ -140,8 +140,6 @@ internal class Program
         services.AddSingleton<IDateTimeProvider, DateTimeProvider>();
         services.AddSingleton<CommandFactory>();
 
-        services.AddHttpClient();
-
         // !!! WARNING !!!
         // stdio-transport-specific implementations of ICacheService.
         // The http-traport-specific implementations and configurations must be registered
--- a/core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using Azure.Mcp.Core.Services.Azure.Authentication;
+using Azure.Mcp.Core.Services.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 
@@ -36,7 +37,7 @@ public static class TenantServiceCollectionExtensions
     /// </item>
     /// </list>
     /// </remarks>
-    public static IServiceCollection AddAzureTenantService(this IServiceCollection services)
+    public static IServiceCollection AddAzureTenantService(this IServiceCollection services, bool addUserAgentClient = false)
     {
         // !!! HACK !!!
         // Program.cs for the CLI servers have their own DI containers vs ServiceStartCommand.
@@ -48,6 +49,12 @@ public static class TenantServiceCollectionExtensions
         // container to be populated with the correct authentication strategy, such as OBO for
         // running as a remote HTTP MCP service.
         services.AddSingleIdentityTokenCredentialProvider();
+
+        services.AddHttpClient();
+        if (addUserAgentClient)
+        {
+            services.ConfigureDefaultHttpClient();
+        }
 
         services.TryAddSingleton<ITenantService, TenantService>();
         return services;


core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,146 @@
...
+            .Replace("?", ".");
+
+        return $"^{regex}$";
+    }
+
+    private static string BuildUserAgent(string transport)
+    {
+        return $"azmcp/{s_version} azmcp-{transport}/{s_version} ({s_framework}; {s_platform})";
+    }
+}
```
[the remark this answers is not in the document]
scbedd: You're absolutely right. The intent is to entirely eliminate HttpClientService though.

core/Azure.Mcp.Core/src/Services/Http/RecordingRedirectHandler.cs
```diff
@@ -8,11 +8,11 @@ namespace Azure.Mcp.Core.Services.Http;
 /// <summary>
 /// DelegatingHandler that rewrites outgoing requests to a recording/replace proxy specified by TEST_PROXY_URL.
 /// It also sets the x-recording-upstream-base-uri header once per request to preserve the original target.
-/// 
+///
 /// This handler is intended to be injected as the LAST delegating handler (closest to the transport) so
 /// that it rewrites the final outgoing wire request.
 /// </summary>
-internal sealed class RecordingRedirectHandler : DelegatingHandler
+public class RecordingRedirectHandler : DelegatingHandler
```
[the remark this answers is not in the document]
scbedd: With the `ClientFactoryConfigurator` moved local here you're absolutely right. This started not this way.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,146 @@
...
+        s_version = assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version ?? "unknown";
+        s_framework = assembly.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName ?? "unknown";
+        s_platform = RuntimeInformation.OSDescription;
+    }
+
+    public static IHttpClientBuilder AddConfiguredHttpClient(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        // for curious eyes, if one doesn't use named clients, AddHttpClient() doesn't return a builder that can be further configured
```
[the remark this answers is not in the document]
scbedd: > This is technically incorrect - the parameterless services.AddHttpClient() does return an IHttpClientBuilder.

I actually would agree with you if I didn't see the compilation errors myself. I guess it's possible that I was seeing an old build error?

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
```diff
@@ -102,4 +102,7 @@ public interface ITenantService
     Task<TokenCredential> GetTokenCredentialAsync(
         string? tenantId,
         CancellationToken cancellationToken);
+
+
+    HttpClient GetClient();
```
vukelich, line 130: When you add doc comments for this, please add instructions for the expected lifetime of the `HttpClient` being used. Language is up to you, but some sort of summary of:

Do:
- Use for a single method or MCP tool invocation (unlimited times within it)
- Modify most aspects to the `HttpClient` instance knowing that it's not reused elsewhere EXCEPT for <whatever things you did in this PR>

Don't:
- Save to an object member UNLESS the lifetime of the owning object is limited to a single MCP tool invocation

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
```diff
@@ -13,16 +13,20 @@ public class TenantService : BaseAzureService, ITenantService
...
     private const string CacheGroup = "tenant";
     private const string CacheKey = "tenants";
     private static readonly TimeSpan s_cacheDuration = TimeSpan.FromHours(12);
 
     public TenantService(
         IAzureTokenCredentialProvider credentialProvider,
-        ICacheService cacheService)
+        ICacheService cacheService,
+        IHttpClientFactory clientFactory
+        )
```
vukelich: Nit: it's not standard C# styling to put method signature closing parentheses on their own line. Put it on the line with the last argument.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,146 @@
...
+        var builder = services.AddHttpClient("default");
```
vukelich: Why are we introducing a named instance called `"default"`? Can we use [`Microsoft.Extensions.Options.Options.DefaultName`](https://source.dot.net/#Microsoft.Extensions.Options/[Options.cs](https://source.dot.net/#Microsoft.Extensions.Options/Options.cs,16),16) (which is `string.Empty`) as the default instance for named instance-supporting DI [which is described in the AddHttpClient doc comment](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/[HttpClientFactoryServiceCollectionExtensions.cs](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs,105),105)?

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -193,6 +195,10 @@ internal static void ConfigureServices(IServiceCollection services)
         services.AddSingleton<ISubscriptionService, SubscriptionService>();
         services.AddSingleton<CommandFactory>();
 
+        // Named overload returns IHttpClientBuilder enabling handler configuration.
+        // without passing named overload, we cannot configure handler.
+        services.AddConfiguredHttpClient();
```
vukelich: Check if we need to also update `Program.cs` for the Fabric and Template projects. We've broken Fabric in the recent past by not updating their `Program.cs` file with DI changes.

Edit: I'm pretty confident that the other executables will break without calling this in the current state since `TenantService` relies on a particular named client. I hope the factory will throw if no such client was first configured before requested.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -215,4 +221,5 @@ internal static async Task InitializeServicesAsync(IServiceProvider serviceProvi
         var telemetryService = serviceProvider.GetRequiredService<ITelemetryService>();
         await telemetryService.InitializeAsync();
     }
+
```
vukelich: Nit: remove empty line

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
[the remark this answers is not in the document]
vukelich: Super technically, there is no parameterless `Microsoft.Extensions.DependencyInjection.HttpClientFactoryServiceCollectionExtensions.AddHttpClient` method. All of them are extension methods with the first method argument being `IServiceCollection`.

After that, there is an `AddHttpClient` method which only takes in `IServiceCollection` as a method argument, but it's for _typed_ instances rather than _named_ instances as reflected in its type arguments. [See here.](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs,240)

I'd recommend instead using `AddHttpClient(IServiceCollection)` followed by [`ConfigureHttpClientDefaults(IServiceCollection, Action<IHttpClientBuilder>)`](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs,74). [The real default named instance (i.e., the one with `string.Empty`) is what can be resolved from DI](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/[HttpClientFactoryServiceCollectionExtensions.cs](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs,66),66) if a scoped or transient type wants to be directly instantiated with it. Otherwise, said type needs to DI with the factory then ask for our customly named instance.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        builder.ConfigureHttpClient((serviceProvider, client) =>
+        {
+            var httpClientOptions = serviceProvider.GetService<IOptions<HttpClientOptions>>()?.Value ?? new HttpClientOptions();
+            client.Timeout = httpClientOptions.DefaultTimeout;
+
+            var transport = serviceProvider.GetService<IOptions<ServiceStartOptions>>()?.Value.Transport ?? TransportTypes.StdIo;
+            client.DefaultRequestHeaders.UserAgent.ParseAdd(BuildUserAgent(transport));
```
vukelich, line 50: The delegate parameter sent to `ConfigureHttpClient` will be invoked every time the associated `HttpClient` needs to be constructed. This means we'll run `BuildUserAgent` to recreate the same string many times. While not a huge deal, it's somewhat inefficiently recurring string allocation. This is partly why past implementations created the user agent string was set to static fields.

vukelich (commented):
Some quick early comments to unblock. Will look more later, too.

servers/Azure.Mcp.Server/src/Program.cs
scbedd: Yeah I figured I'd fail `dotnet format` . It'll get fixed with following commits.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
scbedd: Going to resolve to indicate I updated this comment, but I still welcome feedback 👍

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
scbedd: So it seems to me the best way to fix _this_ if we don't want to build the user agent string is to register a tiny service _before_ anything else. Have that service pull in the service start options to determine the transport, and then have a static user agent string that it provides?

I don't want to inject ANOTHER dependency into this, but that's the way I see if we don't want to check out random statics initialized in the ConfigureServices from `program.cs`. Thoughts?

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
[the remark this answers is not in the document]
scbedd: Done! The default should carry this behavior now. Going to double check my record/playback but it SHOULD be fine.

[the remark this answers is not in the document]
scbedd: This should also mean that we need to simply add `AddHttpClient()` to the fabric/template servers, and they'll all just work :tm:

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
```diff
@@ -102,4 +102,22 @@ public interface ITenantService
     Task<TokenCredential> GetTokenCredentialAsync(
         string? tenantId,
         CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Gets a new instance of <see cref="HttpClient"/> configured with the following options:
+    /// - Proxy
```
vukelich: There's a native list syntax for doc comments. Examples are in our various [`Program.cs`](https://github.com/microsoft/mcp/blob/18ed5f2cfd9373cd7775c93e5c10860cd63f90a7/servers/Azure.Mcp.Server/src/Program.cs#L141-L160) files for a big comment I added warning of dragons. [Official docs.](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#list)

servers/Template.Mcp.Server/src/Program.cs
```diff
@@ -140,6 +140,8 @@ internal static void ConfigureServices(IServiceCollection services)
         services.AddSingleton<IDateTimeProvider, DateTimeProvider>();
         services.AddSingleton<CommandFactory>();
 
+        services.AddHttpClient();
```
scbedd: We need our ConfigureDefaultHttpClientHandler call from Configurator to get record/playback. I need to make certain that I'm not messing with the existing user agent strings etc if the fabric/template servers change what we're using in azure.mcp.server.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs
```diff
@@ -102,4 +105,10 @@ public async Task<TokenCredential> GetTokenCredentialAsync(string? tenantId, Can
     {
         return await _credentialProvider.GetTokenCredentialAsync(tenantId, cancellationToken);
     }
+
+    /// <inheritdoc/>
+    public HttpClient GetClient()
+    {
+        return _httpClientFactory.CreateClient();
+    }
```
vukelich, line 113: This is a nice and clean method. Love it.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,149 @@
...
+        services.ConfigureHttpClientDefaults(static builder => ConfigureHttpClientBuilder(builder));
+
+        return services;
+    }
+
+    private static void ConfigureHttpClientBuilder(IHttpClientBuilder builder)
+    {
+        builder.ConfigureHttpClient((serviceProvider, client) =>
+        {
+            var httpClientOptions = serviceProvider.GetService<IOptions<HttpClientOptions>>()?.Value ?? new HttpClientOptions();
```
vukelich: Any reason why we're using `GetService` rather than `GetRequiredService`? Do we want to allow for a dependency to not exist? If so, please add a comment describing the atypical, but valid, leniency.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,149 @@
...
+        s_version = assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version ?? "unknown";
+        s_framework = assembly.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName ?? "unknown";
+        s_platform = RuntimeInformation.OSDescription;
+    }
+
+    public static IServiceCollection ConfigureDefaultHttpClient(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        services.ConfigureHttpClientDefaults(static builder => ConfigureHttpClientBuilder(builder));
```
vukelich: what is `static` on this line? I don't think I've seen that syntax before. Is it describing the delegate as `static`, and thus, the same delegate on each invocation of `ConfigureDefaultHttpClient`?

vukelich (commented):
Another dump of comments as I continue reading.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
scbedd: The static effectively makes the lambda an isolated closure. Right now it will make _zero_ difference in actual functional code. If we ever add any code to the ConfigureHttpClientDefaults() that accesses external code, we'll immediately error though.

I figured this was literally one place that we call this, so it wouldn't do any harm to have the extra safety.

scbedd: I will explain it in comment in my next commit, but I honestly am not married to it's presence. Just a little piece of info I saw encouraged while digging through various asp.net articles over the weekend.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
scbedd: I was coming at this from the aspect of

> make this as flexible as possible

But given that we have total control of where we're firing this from, I think it's just adding difficulty down the line. I'll swap these to GetRequiredService() and check for breaks.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/ITenantService.cs
scbedd: Fixed.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
vukelich: Thanks for the details. I don't think you needed to add the comment to explain a language feature that I was too lazy to look up myself.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
vukelich: Since DI can't be checked at compile time, I tend to prefer inflexibility. Silently falling back on uninitialized values leads to unknown bugs.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,151 @@
...
+    private static WebProxy? CreateProxy(HttpClientOptions options)
+    {
+        string? proxyAddress = options.AllProxy ?? options.HttpsProxy ?? options.HttpProxy;
+
+        if (string.IsNullOrEmpty(proxyAddress))
+        {
+            return null;
+        }
+
+        if (!Uri.TryCreate(proxyAddress, UriKind.Absolute, out var proxyUri))
```
vukelich, line 94: Future PR suggestion:
If there's a typo in the URL to cause this to fail, like `http//localhost:1010` where the `:` is missing, the user has no way of noticing. It'll silently not use a proxy. 

I'd recommend finding a good way to validate this value at the startup of the application. If it's not an absolute URL, then we should kill the application.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private static string ConvertGlobToRegex(string globPattern)
+    {
+        if (string.IsNullOrEmpty(globPattern))
+        {
+            return string.Empty;
+        }
+
+        var escaped = globPattern
```
vukelich, line 126: Future PR suggestion: 

Calling `string.Replace` creates a new `string` each time because `string` contents cannot be modified. It'd be more efficient if we could use a sort of `ReadOnlySpan<char>.Replace` method (or extension method) to just use stack-allocated memory, but it looks to be an [open issue for .NET.](https://github.com/dotnet/runtime/issues/29758)

Recreating the regex is fine if we want to support changes to `HttpClientOptions` during the runtime of the application, which would require some amount of recalculating (or at least a simple cache) each `HttpClient`/`HttpMessageHandler` creation.

I'm calling this "future PR suggestion" because I don't know if we expect this to run much in the wild or only for our test automation.

vukelich: Also, I'm not a glob or regex expert, so please reach out to someone you think would be good at giving this a green light. This kind of thing usually scares me because it's a similar class of easy vulnerabilities caused by incomplete escaping, like turning input into URLs or SQL/KQL queries.

servers/Template.Mcp.Server/src/Program.cs
vukelich: You make a good point; we could have the user-agent value different per MCP server. 

I think that's out of scope for your current PR. There's a lot already tagging the runtime as the "azure MCP server", including [how we set up telemetry.](https://github.com/microsoft/mcp/blob/18ed5f2cfd9373cd7775c93e5c10860cd63f90a7/core/Azure.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs#L106)

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,151 @@
...
+        s_platform = RuntimeInformation.OSDescription;
+    }
+
+    public static IServiceCollection ConfigureDefaultHttpClient(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        // builder is static to prevent to access instance members or external state beyond what's been injected
+        // within the service collection.
+        services.ConfigureHttpClientDefaults(static builder => ConfigureHttpClientBuilder(builder));
```
hallipr: I think it would be better to just pass the method group instead of wrapping a static in a static.

```suggestion
        services.ConfigureHttpClientDefaults(ConfigureHttpClientBuilder);
```

This doesn't form a closure.

scbedd: > This doesn't form a closure.

I'll just change it.

servers/Fabric.Mcp.Server/src/Program.cs
```diff
@@ -148,6 +148,8 @@ internal static void ConfigureServices(IServiceCollection services)
         services.AddSingleton<ISubscriptionService, SubscriptionService>();
         services.AddSingleton<CommandFactory>();
 
+        services.AddHttpClient();
```
vukelich: If `services.AddHttpClient()` is here because the `TenantService` dependency on `IHttpClientFactory`, then I'd recommend just putting that call directly inside `AddAzureTenantService`. The extension method model for `services.AddXyz` can/should be used like a dependency graph: the extension method should call the extension methods that add the original's dependencies.

vukelich: It's kind of "separation of concerns" thinking. For example, by calling `AddHttpClient`, you're not worried about the dependencies of the services it's adding to the DI. [It starts by calling `AddOptions` and `AddLogging`](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/[HttpClientFactoryServiceCollectionExtensions.cs](https://source.dot.net/#Microsoft.Extensions.Http/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs,30),30), and even those add a breadth of extra things that `AddHttpClient` doesn't need to concern itself with directly.

vukelich: Once you get more complicated, like we have with different implementations or dynamic dependency injection graphs for `ITenantService` and `ICacheService`, this model starts to hit a wall.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
vukelich: Any thoughts about a simple lazy initalization?

```csharp
private static string? s_userAgent = null;

private static string BuildUserAgent(string transport)
{
    s_userAgent ??= $"azmcp/{s_version} azmcp-{transport}/{s_version} ({s_framework}; {s_platform})";
    return s_userAgent;
}
```

vukelich: This could also be a follow up PR. I'd recommend chatting with @srnagar.

vukelich (approved):
Suggestions but all things that could be follow ups.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
hallipr: Does transport have low cardinality?   Would it make sense to have a dictionary of transport -> useragent, or just a single static user agent for the assembly?

hallipr: Ooh... is transport here the mcp transport?

scbedd: Yes. This is exactly the issue. We need the startup option which is why it's injecting those services.

scbedd: > Any thoughts about a simple lazy initalization?
> 
> ```cs
> private static string? s_userAgent = null;
> 
> private static string BuildUserAgent(string transport)
> {
>     s_userAgent ??= $"azmcp/{s_version} azmcp-{transport}/{s_version} ({s_framework}; {s_platform})";
>     return s_userAgent;
> }
> ```

This should be fine -- as the transport type is set at startup and doesn't change after. @hallipr that's not off base right?

servers/Template.Mcp.Server/src/Program.cs
scbedd: Ok. I'll make certain these are unbroken and merge with this. We'll revisit in near future.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
scbedd: Yeah this is inherited code from IHttpClientService. I'll put a pin in this on my notepad and followup after this PR.

core/Azure.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        var proxy = new WebProxy(proxyUri);
+
+        if (!string.IsNullOrEmpty(options.NoProxy))
+        {
+            var bypassList = options.NoProxy
+                .Split(',', StringSplitOptions.RemoveEmptyEntries)
+                .Select(s => s.Trim())
+                .Where(s => !string.IsNullOrEmpty(s))
```
hallipr, line 104: trimming can be done in the split call now, so you don't need to RemoveEmptyEntries, Trim, then remove entries you just made empty:

```cs
	"BASIC,, LEADING,TRAILING , BOTH , ,\tTABS\t,\t \n"
		.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
		.Select(x => $"|{x}|")
```
```
|BASIC|
|LEADING|
|TRAILING|
|BOTH|
|TABS|
```

```suggestion
                .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
```

scbedd: Next PR.

PR #1315, opened 2025-12-04
Ensure Failed Deployments output

scbedd:
@hallipr @alzimmermsft ran into a situation where a failed post-deployment script error got swallowed by our invocation. 

Everywhere has `ErrorActionPreference = Stop` set, but i'd expect output to immediately pop. Nevertheless, I absolutely saw alan's experience where a failure of the post-deployment script due to missing az module never showed up on the console.

### DIFF
--- a/eng/scripts/Deploy-TestResources.ps1
+++ b/eng/scripts/Deploy-TestResources.ps1
@@ -138,7 +138,7 @@ if ($jobInputs.Count -eq 1 -or !$Parallel) {
             Deploy-TestResources @jobInput
         } catch {
             Write-Host "Deployment failed for '$($jobInput.Path)'" -ForegroundColor Red
-            Write-Error $_
+            LogError $_
             $failedPaths += $jobInput.Path
         }
     }


eng/scripts/Deploy-TestResources.ps1
```diff
@@ -137,6 +137,8 @@ if ($jobInputs.Count -eq 1 -or !$Parallel) {
         try {
             Deploy-TestResources @jobInput
         } catch {
+            Write-Host "Deployment failed for '$($jobInput.Path)'" -ForegroundColor Red
+            Write-Error $_
```
hallipr:
```suggestion
            LogError $_
```

hallipr: with `$ErrorActionPreference = 'Stop'`, Write-Error will halt, where we want to just log the error and add the failed path to the list of failed paths.

We'll `exit 1` at the end of the script if there are any failed paths

scbedd: Alrighty then we'll just log it per your suggestion.

hallipr (approved):

Issue #77, opened 2025-08-28
[ONBOARD] Azure Function App


colbytimm:
### Service / Tool Name

Azure.Functionapp

### Contacts

Colby Timm

### Intended Agent Scenarios

Bringing over this issue from the `azure-mcp` repo [onboard function app issue](https://github.com/Azure/azure-mcp/issues/482). Made some edits and improvements and updated the list on what has been completed.

# MCP Roadmap: Azure Function Apps

## 🔎 Discover & Inspect

- [x] List all Function Apps  
- [x] Get details for a Function App (includes runtime, stack, site config, and settings)  
- [ ] List functions in a Function App  
- [ ] Get details for a specific function (bindings, routes, status)  
- [ ] List deployment slots  
- [ ] Show host and function keys  
- [ ] View diagnostics summary (failures, cold starts, quotas)  
- [ ] Download host.json or function.json files  
- [ ] View Kudu/SCM info for a Function App  

---

## 🛠️ Create & Setup

- [ ] Create Function App (basic defaults)  
- [ ] Create Function App (advanced options)  
- [ ] Create Function App (custom container)  
- [ ] Create Function App (Flex or Premium plan)  
- [ ] Create Slot  
- [ ] Create New Project (local)  
- [ ] Create New Containerized Project (local)  
- [ ] Add New Function to Project  
- [ ] Initialize Project for VS Code  
- [ ] Install or Update Azure Functions Core Tools  

---

## ⚙️ Configure & Manage

- [ ] Add Binding to a Function  
- [ ] Add Function App Identity Connection  
- [ ] Add Local Project Identity Connection  
- [ ] Add New Application Setting  
- [ ] Edit Application Setting  
- [ ] Rename Application Setting  
- [ ] Delete Application Setting  
- [ ] Toggle Setting as Slot Setting  
- [ ] Upload Local Settings to Azure  
- [ ] Download Remote Settings from Azure  
- [ ] Encrypt Local Settings  
- [ ] Decrypt Local Settings  
- [ ] Set AzureWebJobsStorage  
- [ ] Enable System-Assigned Identity  
- [ ] Assign Managed Identity to Function App  
- [ ] Configure Deployment Source (GitHub, local zip, container registry)  
- [ ] Connect to GitHub Repository  
- [ ] Disconnect from GitHub Repository  
- [ ] Restart Function App  
- [ ] Report Issue (feedback/help)  
- [ ] Uninstall Azure Functions Core Tools  

---

[...]

PR #1327, opened 2025-12-08
Add functionapp create command

colbytimm:
## What does this PR do?

Adds azmcp functionapp create command to create Azure Function Apps with automatic dependency provisioning.

### Features
- Multiple hosting plans: Consumption (default), Flex Consumption, Premium, App Service, Container App
- Runtime support: .NET, Node.js, Python, Java, PowerShell with automatic OS selection
- Auto-provisioning: Creates Storage accounts and App Service plans when not specified
- Smart defaults: Appropriate SKUs and configurations per hosting type

## GitHub issue number?
#77 

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:


servers/Azure.Mcp.Server/src/Resources/consolidated-tools.json
```diff
@@ -2196,6 +2196,39 @@
...
+          "value": false,
+          "description": "Running this operation multiple times with the same arguments produces the same result without additional effects."
+        },
+        "openWorld": {
+          "value": false,
+          "description": "This tool's domain of interaction is closed and well-defined, limited to a specific set of entities."
+        },
+        "readOnly": {
+          "value": false,
+          "description": "This tool only performs read operations without modifying any state or data."
```
JasonYeMSFT, line 4646: This description should describe what the tool/tools will do to support this metadata value. Given that readOnly is false, it shouldn't say this tool only performs read operations. You can look at other tools with readOnly=false. 

Same comment for other metadata descriptions.

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+        * premium / functionspremium -> EP1 (Elastic Premium)
+        * appservice   -> B1 (Basic) unless overridden by --plan-sku
+    * containerapp -> Creates a Container App instead of an App Service plan/site (no plan created). Container App will reuse the function-app name.
+    - plan-sku: Explicit App Service plan SKU (e.g. B1, S1, P1v3). Overrides --plan-type SKU selection (ignored for containerapp).
+    - runtime: FUNCTIONS_WORKER_RUNTIME (dotnet|dotnet-isolated|node|python|java|powershell). Default: dotnet.
+    - runtime-version: Specific runtime version; if omitted a default per runtime is applied (see defaults below).
+    - os: windows|linux. Default: windows unless runtime/plan requires linux (python, flex consumption, containerapp). Overridden to linux automatically when required. Python & flex consumption do not support Windows.
+
+    Automatic resources & defaults:
+    - Storage account: Always created (Standard_LRS, HTTPS only, blob public access disabled). Name pattern: <sanitized-functionapp>[random6]. Connection string injected as AzureWebJobsStorage.
```
JasonYeMSFT: Azure Function supports using managed identity to connect to storage account. If it is not too hard we should recommend users to use that instead of the connection string since connection strings use storage account keys.

https://learn.microsoft.com/en-us/azure/azure-functions/storage-considerations?tabs=azure-cli#storage-account-connection-setting

tools/Azure.Mcp.Tools.FunctionApp/src/Options/FunctionAppOptionDefinitions.cs
```diff
@@ -6,6 +6,15 @@ namespace Azure.Mcp.Tools.FunctionApp.Options;
...
     public const string FunctionAppName = "function-app";
+    public const string LocationName = "location";
+    public const string AppServicePlanName = "app-service-plan";
+    public const string PlanTypeName = "plan-type";
+    public const string PlanSkuName = "plan-sku";
+    public const string RuntimeName = "runtime";
+    public const string RuntimeVersionName = "runtime-version";
+    public const string OperatingSystemName = "os";
+    public const string StorageAccountName = "storage-account";
+    public const string ContainerAppsEnvironmentName = "container-apps-environment";
```
JasonYeMSFT: If I understand correctly, this option will be ignored when not using the containerapp plan, if not rejected by the argument validation logic.

I don't feel good that some arguments will be completely ignored depending on the value of some other arguments. This looks like a signal that there should be two commands. One is for creating a traditional Azure Function App using an App Service Plan. The other is for creating a container app that runs an Azure Function App.

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
colbytimm: I see. I was following what Azure Portal does by default. That's been my pattern. I don't disagree that using managed identities are best practices.

tools/Azure.Mcp.Tools.FunctionApp/src/Options/FunctionAppOptionDefinitions.cs
colbytimm: I can see how that would be clearer. I'll work on breaking these out into two separate commands.

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -118,8 +214,579 @@ private static void TryAddFunctionApp(WebSiteResource site, List<FunctionAppInfo
...
+        var inputs = ValidateAndNormalizeInputs(
+            subscription, resourceGroup, functionAppName, location,
+            runtime, runtimeVersion, hostingKind, sku, os,
+            storageAccountName, containerAppsEnvironmentName);
+        var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken);
+
+        ResourceGroupResource rg;
+        var resourceGroupResponse = await subscriptionResource.GetResourceGroupAsync(resourceGroup, cancellationToken);
+        if (resourceGroupResponse?.Value == null)
+        {
```
jongio: Bug: Resource group auto-creation is broken. `GetResourceGroupAsync` throws `RequestFailedException` with status 404 when the RG doesn't exist  - it does NOT return null. This means the `if (resourceGroupResponse?.Value == null)` branch is dead code, and the documented "resource-group: Resource group (created if missing)" feature silently fails with an unhandled exception.

The same PR adds `ExistsAsync` patterns for plans and storage accounts (see `EnsureAppServicePlan` and `EnsureStorageForFunctionApp`), but uses a different (broken) pattern here.

Fix: Use the `ExistsAsync` / `GetAsync` pattern consistent with the rest of this PR, or use `CreateOrUpdateAsync` directly (which is idempotent for RGs).

```suggestion
        ResourceGroupResource rg;
        var rgCollection = subscriptionResource.GetResourceGroups();
        if (await rgCollection.ExistsAsync(resourceGroup, cancellationToken))
        {
            rg = (await rgCollection.GetAsync(resourceGroup, cancellationToken)).Value;
        }
        else
        {
            var rgData = new ResourceGroupData(location);
            var rgOp = await rgCollection.CreateOrUpdateAsync(WaitUntil.Completed, resourceGroup, rgData, cancellationToken);
            rg = rgOp.Value;
        }
```

core/Azure.Mcp.Core/src/Areas/Server/Resources/consolidated-tools.json
```diff
@@ -2196,6 +2196,39 @@
...
         "speech_tts_synthesize"
       ]
+    },
+    {
+      "name": "create_azure_app_resource",
+      "description": "Create Azure application platform services, such as Azure Functions.",
+      "toolMetadata": {
+        "destructive": {
+          "value": true,
+          "description": "This tool performs only additive updates without deleting or modifying existing resources."
```
jongio: **Metadata description contradicts boolean value.** `destructive.value` is `true` (tool IS destructive), but the description says "This tool performs only additive updates without deleting or modifying existing resources" which describes a non-destructive tool.

For a create command, `destructive: true` is correct (it creates new Azure resources that cost money). The description should match, e.g.: "This tool creates new Azure resources including Function Apps, Storage accounts, and App Service plans."

core/Azure.Mcp.Core/src/Areas/Server/Resources/consolidated-tools.json
```diff
@@ -2196,6 +2196,39 @@
...
+        },
+        "idempotent": {
+          "value": false,
+          "description": "Running this operation multiple times with the same arguments produces the same result without additional effects."
```
jongio: **Metadata description contradicts boolean value.** `idempotent.value` is `false` (tool is NOT idempotent), but the description says "Running this operation multiple times with the same arguments produces the same result" which describes an idempotent tool.

`idempotent: false` is correct (re-running would fail with a 409 Conflict on the function app name, and auto-generated storage account names include random GUID suffixes). The description should match, e.g.: "Running this operation multiple times may fail due to name conflicts or create additional auto-provisioned resources."

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -118,8 +214,579 @@ private static void TryAddFunctionApp(WebSiteResource site, List<FunctionAppInfo
...
+        CancellationToken cancellationToken = default)
+    {
+        var inputs = ValidateAndNormalizeInputs(
+            subscription, resourceGroup, functionAppName, location,
+            runtime, runtimeVersion, hostingKind, sku, os,
+            storageAccountName, containerAppsEnvironmentName);
+        var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken);
+
+        ResourceGroupResource rg;
+        var resourceGroupResponse = await subscriptionResource.GetResourceGroupAsync(resourceGroup, cancellationToken);
```
jongio: **Duplicated logic: RG creation exists in core but isn't used.** This same PR adds `ResourceGroupService.CreateOrUpdateResourceGroup` to the core library (see `core/Azure.Mcp.Core/src/Services/Azure/ResourceGroup/ResourceGroupService.cs`), but `FunctionAppService.CreateFunctionApp` creates resource groups directly via the SDK instead of using it. This duplicates logic and bypasses the centralized error handling in the core service.

Consider injecting `IResourceGroupService` and calling `CreateOrUpdateResourceGroup`, or remove the core addition if it's not intended for shared use.

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        base.RegisterOptions(command);
+        command.Options.Add(OptionDefinitions.Common.ResourceGroup.AsRequired());
+        command.Options.Add(_functionAppNameOption);
+        command.Options.Add(_locationOption);
+        command.Options.Add(_appServicePlanOption);
+        command.Options.Add(_planTypeOption);
+        command.Options.Add(_planSkuOption);
+        command.Options.Add(_runtimeOption);
+        command.Options.Add(_runtimeVersionOption);
```
jongio: **`--function-app` option is not registered as required.** The Description documents it as a required parameter ("function-app: Globally unique Function App name"), and `--location` is enforced by the parser (`Required = true` in definition). However, `--function-app` has `Required = false` in its definition and is not added with `.AsRequired()`.

Currently, if omitted, the null value passes through to the service layer's `ValidateRequiredParameters` which throws a generic `ArgumentException`. The parser should enforce this earlier for a better error message, matching the pattern used for `ResourceGroup`:
```csharp
command.Options.Add(FunctionAppOptionDefinitions.FunctionApp.AsRequired());
```
Also consider using `.AsRequired()` for `_locationOption` explicitly for consistency, even though it has `Required = true` in its definition.

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+using Microsoft.Mcp.Core.Models.Option;
+
+namespace Azure.Mcp.Tools.FunctionApp.Commands.FunctionApp;
+
+public sealed class FunctionAppCreateCommand(ILogger<FunctionAppCreateCommand> logger)
+    : BaseFunctionAppCommand<FunctionAppCreateOptions>
+{
+    private const string CommandTitle = "Create Azure Function App";
+    private readonly ILogger<FunctionAppCreateCommand> _logger = logger;
+
```
jongio: **Convention deviation: private readonly option fields.** The repo convention (per AGENTS.md: "Use static OptionDefinitions for command options (never readonly fields)") and the existing `FunctionAppGetCommand` in this same toolset reference static definitions directly:
```csharp
command.Options.Add(FunctionAppOptionDefinitions.FunctionApp);
```
This command instead stores them in `private readonly` instance fields, adding unnecessary indirection. While functionally equivalent, it's inconsistent with the pattern used by `FunctionAppGetCommand` and other toolsets like Storage.

Suggestion: Remove lines 22-31 and reference `FunctionAppOptionDefinitions.*` directly in `RegisterOptions` and `BindOptions`.

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -21,6 +27,93 @@ public sealed class FunctionAppService(
     private const string CacheGroup = "functionapp";
```
jongio: **File exceeds 500-line guideline.** `FunctionAppService.cs` was 114 lines and this PR adds ~670 lines, bringing it to ~784 lines. The repo guidelines recommend ≤500 lines per file.

Consider splitting into focused classes:
- `FunctionAppValidation` — `ValidateAndNormalizeInputs`, `ValidateParameterCombinations`, `ResolveOs`
- `FunctionAppStorageProvisioner` — `EnsureStorageForFunctionApp`, `CreateStorageAccountName`, `BuildConnectionString`, `CreateStorageAccountOptions`
- `FunctionAppPlanProvisioner` — `EnsureAppServicePlan`, `CreatePlan`, `ValidateExistingPlan`
- Keep `AppServiceStrategy` and `ContainerAppsStrategy` as top-level internal classes

The static helper methods make this refactoring straightforward.

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -118,8 +214,579 @@ private static void TryAddFunctionApp(WebSiteResource site, List<FunctionAppInfo
...
+    internal static StorageAccountCreateOrUpdateContent CreateStorageAccountOptions(string location)
+    {
+        return new StorageAccountCreateOrUpdateContent(
+            new StorageSku(StorageSkuName.StandardLrs),
+            StorageKind.StorageV2,
+            location)
+        {
+            AccessTier = StorageAccountAccessTier.Hot,
+            EnableHttpsTrafficOnly = true,
+            AllowBlobPublicAccess = false,
```
jongio: **Unnecessary list allocation for storage key retrieval.** This iterates all storage account keys into a `List<StorageAccountKey>` then takes `FirstOrDefault()`. Only the first key is needed.

More efficient approach:
```csharp
StorageAccountKey? primary = null;
await foreach (var key in storage.GetKeysAsync())
{
    primary = key;
    break;
}
if (primary is null)
    throw new InvalidOperationException($"No keys found for storage account '{accountName}'");
return BuildConnectionString(accountName, primary.Value);
```

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -118,8 +214,579 @@ private static void TryAddFunctionApp(WebSiteResource site, List<FunctionAppInfo
...
+    }
+
+    internal static async Task<string> EnsureStorageForFunctionApp(SubscriptionResource subscription, ResourceGroupResource rg, string functionAppName, string location, string? storageAccountName = null)
+    {
+        var accountName = storageAccountName ?? CreateStorageAccountName(functionAppName);
+        var storageAccounts = rg.GetStorageAccounts();
+
+        StorageAccountResource storage;
+        if (await storageAccounts.ExistsAsync(accountName))
+        {
```
jongio: Minor: Use range syntax `[..6]` instead of `Substring(0, 6)` for consistency with modern C# patterns used elsewhere in this file (e.g., `version[..^2]` at line 526).

```suggestion
        var suffix = Guid.NewGuid().ToString("N")[..6];
```

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppService.cs
```diff
@@ -118,8 +214,579 @@ private static void TryAddFunctionApp(WebSiteResource site, List<FunctionAppInfo
...
+        if (options.HostingKind == HostingKind.FlexConsumption)
+            SanitizeAppSettingsForFlexConsumption(appSettings);
+        await site.UpdateApplicationSettingsAsync(appSettings);
+    }
+
+    internal static bool IsFunctionApp(WebSiteData siteData)
+    {
+        return siteData.Kind?.Contains("functionapp", StringComparison.OrdinalIgnoreCase) == true;
+    }
+
```
jongio: Nit: Double blank line. There are a few instances of consecutive blank lines in this file (here after `ApplyAppSettings`, and after `ExtractMajorVersion`). Single blank lines are the convention.

jongio (changes requested):
This is a large addition (1878 lines) that adds `azmcp functionapp create` with support for multiple hosting plans, runtimes, and auto-provisioning of dependencies. The architecture (strategy pattern for AppService vs ContainerApp) is clean, and the unit test coverage is broad (714 lines). However, there are several issues that need addressing before merge.

- 1 CRITICAL: Resource group auto-creation is broken due to incorrect null-check pattern
- 2 HIGH: Metadata contradictions in consolidated-tools.json; duplicated RG creation logic
- 4 MEDIUM: Missing required option enforcement, convention deviations, file size, allocation inefficiency
- 2 LOW: Minor style issues

- Strategy pattern (`AppServiceStrategy` / `ContainerAppsStrategy`) cleanly separates hosting concerns
- Thorough input validation with `ValidateAndNormalizeInputs` and `ValidateParameterCombinations`
- Broad unit tests covering validation, normalization, SKU inference, OS resolution, and configuration building
- Good error messages in `GetErrorMessage` override with specific HTTP status handling
- Proper AOT-safe JSON serialization context registration

Some issues already raised by other reviewers (JasonYeMSFT, Copilot) are not duplicated here. Specifically: readOnly metadata description (JasonYeMSFT), managed identity recommendation (JasonYeMSFT), Java version mismatch in description (Copilot), `Substring(0, 18)` range syntax (Copilot), description formatting (Copilot).

tools/Azure.Mcp.Tools.FunctionApp/src/Options/FunctionAppOptionDefinitions.cs
colbytimm: This is addressed

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (!string.IsNullOrWhiteSpace(options.FunctionAppName))
+            {
+                var len = options.FunctionAppName.Length;
+                if (len < 2 || len > 43)
+                {
+                    context.Response.Status = HttpStatusCode.BadRequest;
+                    context.Response.Message = "function-app name must be between 2 and 43 characters.";
+                    return context.Response;
+                }
```
jongio: Both `FunctionAppCreateCommand` and `FunctionAppCreateContainerAppCommand` have identical name-length validation blocks (2-43 chars). Extract this into `FunctionAppValidation` alongside `ValidateAndNormalizeInputs` - that way a rule change only needs one update.

jongio (commented):
Incremental re-review. All 10 items from my previous review are addressed:

- RG auto-creation bug: fixed (now uses `IResourceGroupService.CreateOrUpdateResourceGroup`)
- Metadata descriptions: fixed (destructive/idempotent values match descriptions)
- Duplicated RG logic: fixed (delegates to core service)
- `--function-app` not required: fixed (`.AsRequired()`)
- Readonly option fields: fixed (references `FunctionAppOptionDefinitions.*` directly)
- File over 500 lines: fixed (split into FunctionAppValidation, StorageProvisioner, PlanProvisioner, ContainerAppStrategy)
- Unnecessary list allocation: fixed (`await foreach` + `break`)
- Range syntax, blank lines: fixed

One new finding on duplicated validation (see inline comment).

Current blockers before this can leave draft:
- Merge conflicts need resolution
- CI failing - multiple files missing `using Azure;` for `WaitUntil.Completed` (already flagged in bot comments)
- ~30 bot comments outstanding worth triaging

tools/Azure.Mcp.Tools.FunctionApp/src/Commands/FunctionApp/FunctionAppCreateCommand.cs
colbytimm: Should be addressed

jongio (commented):
Re-review after rebase. My 10 items from the prior review are all addressed - clean work. A few new things caught my eye on this pass.

The biggest issue is the hardcoded core.windows.net endpoint suffix in FunctionAppStorageProvisioner. This breaks sovereign cloud deployments (Azure Gov, Azure China). The repo already has the pattern for this in StorageService, and even this PR's own 	est-resources.bicep uses nvironment().suffixes.storage instead of hardcoding.

The rest is lower-stakes - visibility modifiers, a silent fallback for unrecognized plan types, and one item from my last review that's still open.

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppStorageProvisioner.cs
```diff
@@ -0,0 +1,113 @@
...
+        if (string.IsNullOrEmpty(baseName))
+            baseName = "func";
+
+        var trimmed = baseName.Length > 18 ? baseName[..18] : baseName;
+        var suffix = Guid.NewGuid().ToString("N")[..6];
+        return $"{trimmed}{suffix}";
+    }
+
+    public static string BuildConnectionString(string accountName, string key) =>
+        $"DefaultEndpointsProtocol=https;AccountName={accountName};AccountKey={key};EndpointSuffix=core.windows.net";
```
jongio: EndpointSuffix=core.windows.net is hardcoded here and again in BuildDeploymentStorageForAccount (the lob.core.windows.net URI on line 110). This won't work for Azure Government (core.usgovcloudapi.net) or Azure China (core.chinacloudapi.cn).

The existing StorageService handles this via _tenantService.CloudConfiguration.CloudType - same pattern would work here. One option: pass the endpoint suffix as a parameter to both methods so the caller (which has access to the tenant service) can provide it. That keeps FunctionAppStorageProvisioner static.

`suggestion
    public static string BuildConnectionString(string accountName, string key, string endpointSuffix = "core.windows.net") =>
        \$"DefaultEndpointsProtocol=https;AccountName={accountName};AccountKey={key};EndpointSuffix={endpointSuffix}";
`

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppStorageProvisioner.cs
```diff
@@ -0,0 +1,113 @@
...
+using Azure.ResourceManager.AppService.Models;
+using Azure.ResourceManager.Resources;
+using Azure.ResourceManager.Storage;
+using Azure.ResourceManager.Storage.Models;
+
+namespace Azure.Mcp.Tools.FunctionApp.Services;
+
+public readonly record struct StorageProvisioningResult(string AccountName, string ConnectionString);
+
+public static class FunctionAppStorageProvisioner
```
jongio: This class and FunctionAppPlanProvisioner are both public. They're implementation details of the FunctionApp toolset, not part of a public API surface. internal would be more appropriate and reduces the maintenance surface.

tools/Azure.Mcp.Tools.FunctionApp/src/Services/FunctionAppValidation.cs
```diff
@@ -0,0 +1,220 @@
...
+        };
+    }
+
+    public static HostingKind ParseHostingKind(string? planType) => planType switch
+    {
+        "flex" or "flexconsumption" => HostingKind.FlexConsumption,
+        "premium" or "functionspremium" => HostingKind.Premium,
+        "appservice" => HostingKind.AppService,
+        "containerapp" or "containerapps" => HostingKind.ContainerApp,
+        _ => HostingKind.Consumption
```
jongio: ParseHostingKind silently maps unrecognized --plan-type values (including typos like premmium) to Consumption. Users won't get any signal that their input was ignored. Consider throwing an ArgumentException for truly unknown values instead of falling back. The catch-all at the bottom could become:

`cs
_ => throw new ArgumentException(\$"Unsupported plan type '{normalizedPlanType}'. Supported values: consumption, flex, premium, appservice, containerapp.")
`

colbytimm:
Thanks @jongio I'll work on addressing those later today.

jongio (commented):
Addresses my previous feedback. The sovereign cloud fix threads the endpoint suffix cleanly through the entire call chain with a sensible default constant. ParseHostingKind now rejects unknown values instead of silently falling back, and the tests cover the new error path. Visibility changes are correct.

One small thing: ParseHostingKind is now case-sensitive (uppercase "Consumption" throws). That works because ValidateAndNormalizeInputs lowercases before calling it, but worth a one-line comment noting this method expects pre-normalized input so a future caller doesn't hit a confusing exception.

jongio (commented):
All previous findings addressed across the last 4 review rounds. The sovereign cloud endpoint fix, visibility changes, ParseHostingKind validation, and documentation comment are all correct.

One housekeeping item before I approve: please resolve the outdated review threads from my earlier reviews so the PR conversation is clean.

colbytimm:
@jongio Done

colbytimm:
@jongio what are the next steps here?

colbytimm:
I'll work on addressing the conflicts. Anything else you need me to do @jongio?

colbytimm:
@jongio What are next steps here?

g2vinay:
@colbytimm 
is this something you're still working on ?
can you fix the CI errors ?

colbytimm:
@g2vinay I can. I just haven't heard from anyone so wasn't sure if there was going to be any movement here. I'm happy to get it into a better position, just want to make sure I can get reviews.

PR #1356, opened 2025-12-10
Only run impacted recorded tests

scbedd:
With the introduction of `ManagedLustre` we're introducing a guaranteed perf hit to every mcl - pullrequest build when we don't need to do that.

This takes care of scoping down where necessary.

### DIFF
--- a/eng/scripts/Test-Code.ps1
+++ b/eng/scripts/Test-Code.ps1
@@ -69,8 +69,8 @@ function FilterTestProjects {
             $changedPaths = $BuildInfo.pathsToTest | ForEach-Object { $_.path }
 
             $testProjects = $testProjects | Where-Object {
-                $prefix = $_.Relative
-                ($changedPaths | Where-Object { $prefix.StartsWith($_) }).Count -gt 0
+                $testProjectPath = $_.Relative
+                ($changedPaths | Where-Object { $testProjectPath.StartsWith($_) }).Count -gt 0
             }
         }
     }


eng/scripts/Test-Code.ps1
```diff
@@ -11,7 +11,8 @@ param(
     [switch] $CollectCoverage,
     [switch] $OpenReport,
     [switch] $TestNativeBuild,
-    [switch] $OnlyBuild
+    [switch] $OnlyBuild,
+    [string] $ScopingBuildInfoPath = $null
```
hallipr, line 15: I would just call it BuildInfoPath to be consistent with all the other scripts that handle build_info.json

eng/scripts/Test-Code.ps1
```diff
@@ -47,13 +56,23 @@ function FilterTestProjects {
...
             Test-Path -Path (Join-Path -Path $projectDirectory -ChildPath 'assets.json')
         }
+
+        # if provided a buildinfo, further scope the recorded tests to only those impacted by changes
+        if ($BuildInfo){
+            $changedPaths = $BuildInfo.pathsToTest | ForEach-Object { $_.path }
+
+            $testProjects = $testProjects | Where-Object {
+                $prefix = $_.Relative
+                ($changedPaths | Where-Object { $prefix.StartsWith($_) }).Count -gt 0
```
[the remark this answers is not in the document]
scbedd: Yeaaaaaahh fine.

hallipr (approved):

eng/scripts/Test-Code.ps1
scbedd: I started here, but given that I'm only using this for scoping I'm going to keep this change.

PR #1419, opened 2025-12-19
[ONBOARD] Add Support for Microsoft.FileShares Resource Provider to Azure MCP Server

ankushbindlish2:
## What does this PR do?
This pull request adds comprehensive Azure Storage Sync (File Sync) support to the Azure MCP Server with 24 new commands organized into 5 resource categories:

- StorageSyncService (5 commands): Create, delete, get, list, and update Storage Sync Services for managing cloud synchronization of file shares.
- 
- RegisteredServer (5 commands): Get, list, register, unregister, and update servers registered with a Storage Sync Service.
- 
- SyncGroup (4 commands): Create, delete, get, and list sync groups that define synchronization topology between cloud and server endpoints.
- 
- CloudEndpoint (5 commands): Create, delete, get, list, and trigger change detection on cloud endpoints representing Azure file shares.
- 
- ServerEndpoint (5 commands): Create, delete, get, list, and update server endpoints with cloud tiering and tiering policy configuration.

All commands include proper error handling, metadata annotations (destructive, idempotent, readOnly flags), comprehensive unit tests (49 passing tests), and live test infrastructure with Azure resource deployment support.

## Open Questions
  - [ ] **Managed Identity Support ?**
  - [ ] ** RoleAssignments creation as part of resource creation ? **
  - [ ] ** Hybrid API which has no significance without being run at a windows server like PUT REGISTEREDSERVER ? **
  - [ ] ** PrivateEndpointConnections API for connection management ?**

## GitHub issue number?
[[ONBOARD] Add Support for Microsoft.StorageSync Resource Provider to Azure MCP Server #1354](https://github.com/microsoft/mcp/issues/1354)

## Pre-merge Checklist
- [x] Required for All PRs
    - [ ] **Read [contribution guidelines](https://github.com/microsoft/mcp/blob/main/CONTRIBUTING.md)**
    - [x] PR title clearly describes the change
    - [x] Commit history is clean with descriptive messages ([cleanup guide](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md))
    - [x] Added comprehensive tests for new/modified functionality
    - [x] Updated `servers/Azure.Mcp.Server/CHANGELOG.md` and/or `servers/Fabric.Mcp.Server/CHANGELOG.md` for product changes (`features, bug fixes, UI/UX, updated dependencies`)
- [ ] For MCP tool changes:
    - [x] **One tool per PR**: This PR adds or modifies only one MCP tool for faster review cycles
[...]

PR #1472, opened 2026-01-09
Throw fewer exceptions from AppLens

tmeschter:
## What does this PR do?

The AppLens tool has a very low success rate. Looking through the code there are several common, non-exceptional cases where we're throwing an exception that escapes the tool. The MCP server infrastructure captures these exceptions and treats them as failures.

For example, we will currently throw an exception if the resource can't be found. This is pretty normal--the user may have mistyped the name, or the resource exists but not in the specified subscription. Rather than throwing exceptions we should simply return a message telling the LM that the resource wasn't found and to check the spelling.

Most of the changes here are concerned with bubbling existing messages up to the LM in this manner. This is not a complete solution to the problem but does address some low-hanging fruit. Beyond this we'll need to invest in better telemetry for AppLens so that we can see where things tend to go sideways.

## GitHub issue number?
https://github.com/microsoft/mcp-pr/issues/172

## Pre-merge Checklist
- [ ] Required for All PRs
- [x] For MCP tool changes:
    - [x] **One tool per PR**: This PR adds or modifies only one MCP tool for faster review cycles
    - [x] Updated `servers/Azure.Mcp.Server/README.md` and/or `servers/Fabric.Mcp.Server/README.md` documentation
    - [x] Validate README.md changes using script at `eng/scripts/Process-PackageReadMe.ps1`. See [Package README](https://github.com/microsoft/mcp/blob/main/CONTRIBUTING.md#package-readme)
    - [x] Updated command list in `/servers/Azure.Mcp.Server/docs/azmcp-commands.md` and/or `/docs/fabric-commands.md`
    - [x] Run `.\eng\scripts\Update-AzCommandsMetadata.ps1` to update tool metadata in azmcp-commands.md (required for CI)
    - [x] For new or modified tool descriptions, ran [`ToolDescriptionEvaluator`](https://github.com/microsoft/mcp/blob/main/eng/tools/ToolDescriptionEvaluator/Quickstart.md) and obtained a score of `0.4` or more and a top 3 ranking for all related test prompts
    - [x] For tools with new names, including new tools or renamed tools, update [`consolidated-tools.json`](https://github.com/microsoft/mcp/blob/main/core/Azure.Mcp.Core/src/Areas/Server/Resources/consolidated-tools.json)
    - [x] For new tools associated with Azure services or publicly available tools/APIs/products, add URL to documentation in the PR description

### DIFF
--- a/servers/Azure.Mcp.Server/CHANGELOG.md
+++ b/servers/Azure.Mcp.Server/CHANGELOG.md
@@ -10,6 +10,8 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Bugs Fixed
 
+- Fixed various exceptions being thrown by the AppLens tool. [[#1419](https://github.com/microsoft/mcp/pull/1419)]
+
 ### Other Changes
 
 ## 2.0.0-beta.10 (2026-01-09)
--- a/tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
+++ b/tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
@@ -39,34 +39,35 @@ public class AppLensService(IHttpClientService httpClientService, ISubscriptionS
         // Step 1: Get the resource ID
         var findResult = await FindResourceIdAsync(resource, subscriptionResource.Data.SubscriptionId, resourceGroup, resourceType, cancellationToken);
 
-        if (findResult is DidNotFindResourceResult notFound)
+        if (findResult is Failure<FoundResourceData> notFoundResult)
         {
-            throw new InvalidOperationException(notFound.Message);
+            throw new InvalidOperationException(notFoundResult.Message);
         }
 
-        var foundResource = (FoundResourceResult)findResult;
+        var successfulResult = (Success<FoundResourceData>)findResult;
+        var resourceData = successfulResult.Data;
 
         // Step 2: Get AppLens session
-        var session = await GetAppLensSessionAsync(foundResource.ResourceId, tenantId, cancellationToken);
+        var getSessionResult = await GetAppLensSessionAsync(resourceData.ResourceId, tenantId, cancellationToken);
 
-        if (session is FailedAppLensSessionResult failed)
+        if (getSessionResult is Failure<AppLensSession> failed)
         {
             throw new InvalidOperationException(failed.Message);
         }
 
-        var successfulSession = (SuccessfulAppLensSessionResult)session;
+        var successResult = (Success<AppLensSession>)getSessionResult;
 
         // Step 3: Ask AppLens the diagnostic question
-        var insights = await CollectInsightsAsync(successfulSession.Session, question, cancellationToken);
+        var insights = await CollectInsightsAsync(successResult.Data, question, cancellationToken);
 
         return new DiagnosticResult(
             insights.Insights,
             insights.Solutions,
-            foundResource.ResourceId,
-            foundResource.ResourceTypeAndKind);
+            resourceData.ResourceId,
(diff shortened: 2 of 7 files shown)


alzimmermsft (approved):
Does this change create larger implications on what we determine to be a successful or failed MCP tool call? For example, does the failure to find a resource change to a successful tool call but the operation itself failed due to the resource not existing.

tmeschter:
>Does this change create larger implications on what we determine to be a successful or failed MCP tool call? For example, does the failure to find a resource change to a successful tool call but the operation itself failed due to the resource not existing.

Yes. Currently if we can't find the resource we will throw an exception and the tool call will be recorded as a failure. After this change it will be recorded as a success. That's intentional; it's entirely normal and expected that the input resource won't exist. The most likely explanation is that the user mistyped the name; we will now provide a "successful" response to that effect.

As another example, the resource might exist and be found, but the resource type isn't supported by AppLens. That's a limitation of the system and may not provide a useful response to the user, but nothing has actually gone wrong when that happens.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
```diff
@@ -39,34 +39,35 @@ public async Task<DiagnosticResult> DiagnoseResourceAsync(
         // Step 1: Get the resource ID
         var findResult = await FindResourceIdAsync(resource, subscriptionResource.Data.SubscriptionId, resourceGroup, resourceType, cancellationToken);
 
-        if (findResult is DidNotFindResourceResult notFound)
+        if (findResult is Failure<FoundResourceData> notFoundResult)
         {
-            throw new InvalidOperationException(notFound.Message);
+            return Result<AppLensInsights>.Failure(notFoundResult.Message);
```
kvenkatrajan, line 44: It would be good to have telemetry to understand how often or not the LLM gets the resource it expects.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
```diff
@@ -77,15 +78,15 @@ private Task<FindResourceIdResult> FindResourceIdAsync(
 
         if (string.IsNullOrEmpty(subscription) || string.IsNullOrEmpty(resourceGroup))
```
kvenkatrajan, line 79: Same here - telemetry capturing would be good here. ideally subscription should not be null so might be good to understand how many times the subscription or resourcegroup is null

servers/Azure.Mcp.Server/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Bugs Fixed
 
+- Fixed various exceptions being thrown by the AppLens tool. [[#1419](https://github.com/microsoft/mcp/pull/1419)]
```
jongio, line 13: **[MEDIUM]** Quality: Wrong PR link in changelog.

This entry references #1419 but this is PR #1472.

```suggestion
- Fixed various exceptions being thrown by the AppLens tool. [[#1472](https://github.com/microsoft/mcp/pull/1472)]
```

jongio (commented):
Changelog entry links to wrong PR.

Issues to address:
- CHANGELOG.md:13 - references #1419 instead of #1472

Minor nit: the direct casts in AppLensService.cs (lines ~47, ~58) after negative type checks are safe with sealed subclasses but pattern matching would be more idiomatic. Not blocking.

Existing open feedback from Copilot (string interpolation bug, missing Failure-path tests) and @kvenkatrajan (telemetry) is still unaddressed.

PR #1501, opened 2026-01-14
Update ModelContextProtocol to latest

stephentoub:
## What does this PR do?

Updates ModelContextProtocol dependency to 0.6.0-preview.1. Also bump Microsoft.Extensions.* to latest.

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/BaseToolLoaderTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/BaseToolLoaderTests.cs
@@ -238,12 +238,27 @@ public class BaseToolLoaderTests
         var capabilities = new ClientCapabilities
         {
             Elicitation = new ElicitationCapability()
+            {
+                Form = new(),
+            }
         };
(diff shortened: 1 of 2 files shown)


core/Azure.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -638,11 +638,12 @@ await request.Server.NotifyProgressAsync(progressToken.Value,
 
         var samplingRequest = new CreateMessageRequestParams
         {
+            MaxTokens = 1000,
```
alzimmermsft: Mind adding a comment for where `MaxTokens = 1000` came from and why. Right now, if feels opaque and will result in this being something never touched again due to that, but could be a candidate for further research and improvement.

stephentoub: It's arbitrary. The MCP protocol requires a max tokens to be specified. Previously the C# SDK was defaulting the value (to 1000), but we removed that default and force the consumer to do what the MCP protocol requires, which is specify a max tokens.
```suggestion
            MaxTokens = 1000, // arbitrary limit that can be changed in the future as needed; MCP requires a max be specified
```

PR #1509, opened 2026-01-15
Support OAuth protected registry server

JasonYeMSFT:
## What does this PR do?

This PR extends the RegistryServerProvider to support adding registry servers protected by OAuth for all its MCP accesses.

`registry.json` now let's MCP servers with http transport to optionally declare an `oauthScopes` parameter. When this property is defined, the RegistryServerProvider will instantiate the corresponding McpClient object with a custom HttpClient. This custom HttpClient object has a delegate handler that fetches an access token and sets it in the Authorization header in each request.

Test scenarios

- local stdio + environment credential
- local stdio + user credential
- remote + environment credential
- remote + user credential obo

The server may apply additional restrictions that limit access, such as the tenant restriction. 

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/AccessTokenHandlerTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/AccessTokenHandlerTests.cs
@@ -126,31 +126,6 @@ public class AccessTokenHandlerTests
     }
 
     [Fact]
-    public async Task SendAsync_PassesCancellationTokenToTokenProvider()
-    {
-        // Arrange
-        var handler = new AccessTokenHandler(_mockTokenCredentialProvider, _oauthScopes);
-        var mockInnerHandler = new MockHttpMessageHandler();
-        handler.InnerHandler = mockInnerHandler;
-
-        using var client = new HttpClient(handler);
-        var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com/api/test");
-        using var cts = new CancellationTokenSource();
-
-        // Act
-        await client.SendAsync(request, cts.Token);
-
-        // Assert
-        await _mockTokenCredentialProvider.Received(1).GetTokenCredentialAsync(
-            null,
-            cts.Token);
-
-        await _mockTokenCredential.Received(1).GetTokenAsync(
-            Arg.Any<TokenRequestContext>(),
-            cts.Token);
-    }
-
-    [Fact]
     public async Task SendAsync_CallsBaseHandlerAfterSettingAuthorizationHeader()
     {
         // Arrange


core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -88,8 +93,30 @@ private async Task<McpClient> CreateHttpClientAsync(McpClientOptions clientOptio
...
+        {
+            var FetchAccessToken = async (CancellationToken cancellationToken) =>
+            {
+                var credential = await _tokenCredentialProvider.GetTokenCredentialAsync(tenantId: null, cancellationToken);
+                var tokenContext = new Azure.Core.TokenRequestContext(_serverInfo.OAuthScopes);
+                var token = await credential.GetTokenAsync(tokenContext, cancellationToken);
+                return token.Token;
+            };
+
+            var client = _httpClientService.CreateClientWithAccessToken(FetchAccessToken);
```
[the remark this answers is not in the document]
JasonYeMSFT: We should avoid imposing default behavior than necessary and let the MCP SDK handle what to do for requests.

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
```diff
@@ -62,6 +63,41 @@ public HttpClient CreateClient(Uri? baseAddress, Action<HttpClient> configureCli
         return client;
     }
 
+    /// <summary>
+    /// Creates a new HttpClient instance with a customized function to acquire an access token for its outgoing requests.
+    /// </summary>
+    /// <param name="accessTokenProvider">A function to acquire access token.</param>
+    /// <param name="baseAddress">The base address for the HttpClient</param>
+    /// <returns>A new HttpClient instance.</returns>
+    public HttpClient CreateClientWithAccessToken(Func<CancellationToken, Task<string>> accessTokenProvider, Uri? baseAddress)
```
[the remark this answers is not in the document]
JasonYeMSFT: Added test coverage to make sure: 
- the client can be created with proxy options
- the client is created with correct user agent
- the client does fetch access token using the provided token method
- token fetching can be cancelled by the upstream cancellation token

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/Discovery/RegistryServerProviderTests.cs
```diff
@@ -5,13 +5,24 @@
...
 
 namespace Azure.Mcp.Core.UnitTests.Areas.Server.Commands.Discovery;
 
 public class RegistryServerProviderTests
 {
+    private static RegistryServerProvider CreateServerProvider(string id, RegistryServerInfo serverInfo)
+    {
+        var httpClientService = NSubstitute.Substitute.For<IHttpClientService>();
+        httpClientService.CreateClientWithAccessToken(Arg.Any<Func<CancellationToken, Task<string>>>(), Arg.Any<Uri?>())
+            .Returns(new HttpClient());
```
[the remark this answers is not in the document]
JasonYeMSFT: Adopted the suggestion.

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
```diff
@@ -191,4 +227,27 @@ public void Dispose()
             _disposed = true;
         }
     }
+
+    /// <summary>
+    /// DelegatingHandler that adds a Bearer access token to each outgoing request.
+    /// </summary>
+    private sealed class AccessTokenHandler : DelegatingHandler
```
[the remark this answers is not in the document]
JasonYeMSFT: The testing is covered by HttpClientService.CreateClientWithAccessToken.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -88,8 +93,30 @@ private async Task<McpClient> CreateHttpClientAsync(McpClientOptions clientOptio
...
+            // HttpClientTransportOptions offers an OAuth property to configure client side OAuth parameters, such as RedirectUri and ClientId.
+            // When OAuth property is set, the MCP client will attempt to complete the Auth flow following the MCP protocol.
+            // However, there is a gap between what MCP protocol requires the OAuth provider to implement and what Entra supports. This MCP client will always send a resource parameter to the token endpoint because it is required by the MCP protocol but Entra doesn't support it. More details in issue #939 and related discussions in modelcontextprotocol/csharp-sdk GitHub repo.
         };
-        var clientTransport = new HttpClientTransport(transportOptions);
+
+        HttpClientTransport clientTransport;
+        if (_serverInfo.OAuthScopes is not null)
+        {
+            var FetchAccessToken = async (CancellationToken cancellationToken) =>
```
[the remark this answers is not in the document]
JasonYeMSFT: Fixed the naming.

core/Azure.Mcp.Core/src/Areas/Server/Resources/registry.json
```diff
@@ -12,6 +12,12 @@
             "title": "Azure Developer CLI",
             "description": "Azure Developer CLI (azd) includes a suite of tools to help build, modernize, and manage applications on Azure. It simplifies the process of developing cloud applications by providing commands for project initialization, resource provisioning, deployment, and monitoring. Use this tool to streamline your Azure development workflow and manage your cloud resources efficiently.",
             "installInstructions": "The Azure Developer CLI (azd) is either not installed or requires an update. The minimum required version that works with MCP tools is 1.20.0.\n\nTo install or upgrade, follow the instructions at https://aka.ms/azd/install\n\nAfter installation you may need to restart the Azure MCP server and your IDE."
+        },
+        "arm": {
+            "url": "https://tbd.com",
+            "title": "Azure Resource Manager",
+            "description": "Query Azure Resource Graph for Azure Resource information",
+            "oauthScopes": [ "api://tbd_app_id/.default" ]
```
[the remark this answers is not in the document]
JasonYeMSFT: Removed.

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
```diff
@@ -62,6 +63,41 @@ public HttpClient CreateClient(Uri? baseAddress, Action<HttpClient> configureCli
         return client;
     }
 
+    /// <summary>
+    /// Creates a new HttpClient instance with a customized function to acquire an access token for its outgoing requests.
+    /// </summary>
+    /// <param name="accessTokenProvider">A function to acquire access token.</param>
+    /// <param name="baseAddress">The base address for the HttpClient</param>
+    /// <returns>A new HttpClient instance.</returns>
+    public HttpClient CreateClientWithAccessToken(Func<CancellationToken, Task<string>> accessTokenProvider, Uri? baseAddress = null)
```
anuchandy: In the RegistryServerProvider, can we use DI injected `HttpClientFactory` to obtain the HttpClient instead of `HttpClientService`? We’re in the process of removing our dependency on HttpClientService and avoiding explicit HttpClient creation across the repo.

JasonYeMSFT: Yes, I will see if it can be easily done. I didn't know we are migrating from HttpClientService to HttpClientFactory.

anuchandy: thanks @JasonYeMSFT, the details are here https://github.com/microsoft/mcp-pr/issues/152

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
```diff
@@ -15,9 +17,11 @@ namespace Azure.Mcp.Core.Areas.Server.Commands.Discovery;
 /// </summary>
 /// <param name="options">Options for configuring the service behavior.</param>
 /// <param name="logger">Logger instance for this discovery strategy.</param>
-public sealed class RegistryDiscoveryStrategy(IOptions<ServiceStartOptions> options, ILogger<RegistryDiscoveryStrategy> logger) : BaseDiscoveryStrategy(logger)
+public sealed class RegistryDiscoveryStrategy(IOptions<ServiceStartOptions> options, ILogger<RegistryDiscoveryStrategy> logger, IHttpClientService httpClientService, IAzureTokenCredentialProvider tokenCredentialProvider) : BaseDiscoveryStrategy(logger)
```
vukelich: Are you able to use `IHttpClientFactory` instead here? We started moving away from `IHttpClientService` with #1239 because `...Factory` is part of the .NET SDK that does more than our bespoke `...Service`.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -12,10 +14,14 @@ namespace Azure.Mcp.Core.Areas.Server.Commands.Discovery;
...
 /// <param name="id">The unique identifier for the server.</param>
 /// <param name="serverInfo">Configuration information for the server.</param>
-public sealed class RegistryServerProvider(string id, RegistryServerInfo serverInfo) : IMcpServerProvider
+/// <param name="httpClientService">The HTTP client service for creating HTTP clients.</param>
+/// <param name="tokenCredentialProvider">The token credential provider for OAuth authentication.</param>
+public sealed class RegistryServerProvider(string id, RegistryServerInfo serverInfo, IHttpClientService httpClientService, IAzureTokenCredentialProvider tokenCredentialProvider) : IMcpServerProvider
 {
     private readonly string _id = id;
     private readonly RegistryServerInfo _serverInfo = serverInfo;
+    private readonly IHttpClientService _httpClientService = httpClientService;
```
vukelich: Same as before, can we use `IHttpClientFactory` here?

vukelich (changes requested):
Chatted offline. Jason was already mid-change for switching to `IHttpClientFactory`. Will return for review when that's complete.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
JasonYeMSFT: I pushed the changes for creating custom HttpClient objects for these registered MCP servers using IHttpClientFactory.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
JasonYeMSFT: I pushed the changes for creating custom HttpClient objects using IHttpClientFactory.

core/Azure.Mcp.Core/src/Areas/Server/Models/RegistryServerInfo.cs
```diff
@@ -19,11 +19,19 @@ public sealed class RegistryServerInfo
     public string? Name { get; set; }
 
     /// <summary>
-    /// Gets the URL endpoint (deprecated - no longer used).
+    /// URL of the remote server.
```
vukelich: Doc comments for C# properties have a historical wording convention based on https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1623.md, a previously common style analyzer. This is why a lot of other properties in this file, properties across the .NET SDK, and tools within Visual Studio all have or create descriptions that start with `/// Gets ....` or `/// Gets or sets...`. Please utilize that style.

core/Azure.Mcp.Core/src/Areas/Server/Models/RegistryServerInfo.cs
```diff
@@ -38,6 +46,7 @@ public sealed class RegistryServerInfo
 
     /// <summary>
     /// Gets the transport type, e.g., "stdio".
+    /// This should be undefined if url is defined.
```
vukelich: Since this is C# code, we don't have an `undefined`, and we should use the corresponding `null` language word instead.

```
/// This should be <see langword="null"/> if <see cref="Url"/> is non-<see langword="null"/>.
```

Image

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
```diff
@@ -56,7 +60,66 @@ public static IServiceCollection AddAzureTenantService(this IServiceCollection s
...
+        services.AddHttpClientForMcpRegistry();
+
         services.TryAddSingleton<ITenantService, TenantService>();
         return services;
     }
+
+    /// <summary>
+    /// Add HttpClient for each registry server with OAuthScopes that knows how to fetch its access token.
+    /// </summary>
+    private static IServiceCollection AddHttpClientForMcpRegistry(this IServiceCollection services)
```
vukelich: For clean code organization, please put this code somewhere close to adding the registry tool loader and discovery strategy into the DI container. This code is not relevant for the `TenantService`.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
```diff
@@ -56,7 +60,66 @@ public static IServiceCollection AddAzureTenantService(this IServiceCollection s
...
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var resourceName = assembly
```
vukelich: I'd strongly recommend leaving a comment as a warning to other maintainers both here and in `RegistryDiscoveryStrategy` that we have two different files reading and acting on `registry.json`. As an example, if one of them changes the name of the file, then the other will break.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
```diff
@@ -56,7 +60,66 @@ public static IServiceCollection AddAzureTenantService(this IServiceCollection s
...
+            .GetManifestResourceNames()
+            .FirstOrDefault(n => n.EndsWith("registry.json", StringComparison.OrdinalIgnoreCase));
+        if (resourceName is null)
```
vukelich: As an alternative to opening and parsing this file within this method and tossing out the results, we could load the results into an object and add it into the DI container such as adding `RegistryRoot` as a type into the DI for `RegistryDiscoveryStrategy` to be constructed with.

vukelich: This allows the application to have a "read once" rather than multiple times in separate spots.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
```diff
@@ -15,9 +16,11 @@ namespace Azure.Mcp.Core.Areas.Server.Commands.Discovery;
 /// </summary>
 /// <param name="options">Options for configuring the service behavior.</param>
 /// <param name="logger">Logger instance for this discovery strategy.</param>
-public sealed class RegistryDiscoveryStrategy(IOptions<ServiceStartOptions> options, ILogger<RegistryDiscoveryStrategy> logger) : BaseDiscoveryStrategy(logger)
+public sealed class RegistryDiscoveryStrategy(IOptions<ServiceStartOptions> options, ILogger<RegistryDiscoveryStrategy> logger, IHttpClientFactory httpClientFactory, IAzureTokenCredentialProvider tokenCredentialProvider) : BaseDiscoveryStrategy(logger)
 {
     private readonly IOptions<ServiceStartOptions> _options = options;
+    private readonly IHttpClientFactory _httpClientFactory = httpClientFactory;
+    private readonly IAzureTokenCredentialProvider _tokenCredentialProvider = tokenCredentialProvider;
```
vukelich: `IAzureTokenCredentialProvider` is no longer used with the move to `IHttpClientFactory` within `RegistryDiscoveryStrategy` or `RegistryServerProvider`

vukelich (changes requested):

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
```diff
@@ -56,7 +60,66 @@ public static IServiceCollection AddAzureTenantService(this IServiceCollection s
...
+            var oauthScopes = kvp.Value.OAuthScopes;
+            if (oauthScopes.Length == 0)
+            {
+                continue;
+            }
+
+            services.AddHttpClient(RegistryServerHelper.GetRegistryServerHttpClientName(serverName))
+                .AddHttpMessageHandler((services) =>
+                {
+                    var fetchAccessToken = async (CancellationToken cancellationToken) =>
```
vukelich: Rather than passing around function callbacks, can we please construct `AccessTokenHandler` with the OAuth scopes and a credential provider, and the internal logic of `AccessTokenHandler` has this logic?

vukelich: This is very TypeScript/JavaScript-esque rather than OOB

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Net;
+using System.Net.Http.Headers;
```
vukelich: Can remove this change.

core/Azure.Mcp.Core/src/Areas/Server/Models/RegistryServerInfo.cs
JasonYeMSFT: I restored the "Gets" pattern following the convention. I fixed an AI hallucination in the old comment since this parameter has never been deprecated.

core/Azure.Mcp.Core/src/Areas/Server/Models/RegistryServerInfo.cs
JasonYeMSFT: I decorated the new comment using the special doc string syntax.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
JasonYeMSFT: The logic for registring RegistryServer related DI depdendencies have been moved to RegistryServerServiceCollectionExtension.cs.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
JasonYeMSFT: Now azmcp loads the registry.json once and registers the parsed object as a DI singleton.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
JasonYeMSFT: Removed.

core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantServiceCollectionExtensions.cs
JasonYeMSFT: It's modified to construct AccessTokenHandler with an IAzureTokenCredentialProvider and oauthScopes.

core/Azure.Mcp.Core/src/Services/Http/HttpClientService.cs
JasonYeMSFT: This file has been removed.

core/Azure.Mcp.Core/src/Areas/Server/RegistryServerServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    /// <summary>
+    /// Add HttpClient for each registry server with OAuthScopes that knows how to fetch its access token.
+    /// </summary>
+    public static IServiceCollection AddRegistryRoot(this IServiceCollection services)
+    {
+        var registry = RegistryServerHelper.GetRegistryRoot();
+        if (registry?.Servers is null)
+        {
+            return services;
```
vukelich, line 26: We must put something for `IRegistryRoot` in the DI container for `RegistryDiscoveryStrategy` to be resolved. If we bail out here early, then things will break.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+using Azure.Mcp.Core.Areas.Server.Commands.Discovery;
+using Azure.Mcp.Core.Areas.Server.Models;
+using Azure.Mcp.Core.Services.Azure.Authentication;
+
+namespace Azure.Mcp.Core.Helpers
+{
+    /// <summary>
+    /// DelegatingHandler that adds a Bearer access token to each outgoing request.
+    /// </summary>
+    public sealed class AccessTokenHandler : DelegatingHandler
```
vukelich: Idk if this repo is following the convention that each `.cs` file should have a single top-level class. Having `AccessTokenHandler` in `RegistryServerHelper.cs` might be frowned upon.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+using Azure.Core;
+using Azure.Mcp.Core.Areas.Server;
+using Azure.Mcp.Core.Areas.Server.Commands.Discovery;
+using Azure.Mcp.Core.Areas.Server.Models;
+using Azure.Mcp.Core.Services.Azure.Authentication;
+
+namespace Azure.Mcp.Core.Helpers
+{
+    /// <summary>
+    /// DelegatingHandler that adds a Bearer access token to each outgoing request.
```
vukelich: Use doc comment syntax when referring to types

```csharp
/// <see cref="DelegatingHandler "/> that adds a Bearer access token to each outgoing request.
```

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public AccessTokenHandler(IAzureTokenCredentialProvider tokenCredentialProvider, string[] oauthScopes)
+        {
+            _tokenCredentialProvider = tokenCredentialProvider;
+            _oauthScopes = oauthScopes;
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var credential = await _tokenCredentialProvider.GetTokenCredentialAsync(tenantId: null, cancellationToken);
```
vukelich: What should `SendAsync` be doing if there's already an `Authorization` header on the request? Should it skip adding one? Should it override, as the current code is doing? Whatever the decision, please add a comment for other maintainers to understand whatever chosen business logic is intentional rather than accidental. Also, add unit tests that verify the behavior to catch possible regressions.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+            var tokenContext = new TokenRequestContext(_oauthScopes);
+            var token = await credential.GetTokenAsync(tokenContext, cancellationToken);
+            if (!string.IsNullOrEmpty(token.Token))
```
vukelich: Is there a reason for allowing token creation to return a null or empty value? What's the design rationale? Are there scenarios where that's valid?

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        public AccessTokenHandler(IAzureTokenCredentialProvider tokenCredentialProvider, string[] oauthScopes)
+        {
+            _tokenCredentialProvider = tokenCredentialProvider;
+            _oauthScopes = oauthScopes;
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var credential = await _tokenCredentialProvider.GetTokenCredentialAsync(tenantId: null, cancellationToken);
+            var tokenContext = new TokenRequestContext(_oauthScopes);
```
vukelich: What should we do if `_oauthScopes` is an empty array? First of all, should we be adding validation for throwing on an empty array in the constructor? If we are ok with an empty array, then we should harden the code here to not try getting an access token.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+            return await base.SendAsync(request, cancellationToken);
+        }
+    }
+
+    public sealed class RegistryServerHelper
+    {
+        public static string GetRegistryServerHttpClientName(string serverName)
+        {
+            return $"azmcp.{nameof(RegistryServerProvider)}.{serverName}";
```
vukelich: I'd recommend not using `azmcp` as a prefix. @hallipr is trying to make some of the core code less azmcp specific. We could use `$"{typeof(RegistryServerProvider).FullName}.{serverName}"`, similar to what we chatted about before.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public static IRegistryRoot? GetRegistryRoot()
+        {
+            var assembly = Assembly.GetExecutingAssembly();
+            var resourceName = assembly
+                .GetManifestResourceNames()
+                .FirstOrDefault(n => n.EndsWith("registry.json", StringComparison.OrdinalIgnoreCase));
+            if (resourceName is null)
+            {
+                return null;
```
vukelich: Are there any scenarios where we're ok with not finding `registry.json`? If so, please document them. If we want to keep things simple and always expect it's there, then we should be aggressive at coding to that expectation.

One quirk is if this code path gets hit during unit tests. The unit test `.dll` likely won't have this embedded unless we enforce it.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            return $"azmcp.{nameof(RegistryServerProvider)}.{serverName}";
+        }
+
+        public static IRegistryRoot? GetRegistryRoot()
+        {
+            var assembly = Assembly.GetExecutingAssembly();
+            var resourceName = assembly
+                .GetManifestResourceNames()
+                .FirstOrDefault(n => n.EndsWith("registry.json", StringComparison.OrdinalIgnoreCase));
```
vukelich: It might be safer to explicitly match on the file name rather than `EndsWith`. A file like `AnotherRegistry.json` being included could be picked up instead, and that'd break this feature. I don't know what the exact resource name would be to match on, or if there's a directory structure to also deal with.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
vukelich: Can probably check for `registryRoot.Servers!` in the previous `if` statement rather than null-forgiving on this line. `.Where` will throw if given a `null` `IEnumerable`.

core/Azure.Mcp.Core/src/Areas/Server/Models/IRegistryRoot.cs
```diff
@@ -0,0 +1,16 @@
...
+/// <summary>
+/// Represents the root structure of the MCP server registry JSON file.
+/// Contains a collection of server configurations keyed by server name.
+/// </summary>
+public interface IRegistryRoot
+{
+    /// <summary>
+    /// Gets the dictionary of server configurations, keyed by server name.
+    /// </summary>
+    public Dictionary<string, RegistryServerInfo>? Servers { get; init; }
```
vukelich, line 15: Any reason to have this be nullable?

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
vukelich: Ok, I think I see why the code is working in the current state. 

First, `RegistryServerServiceCollectionExtensions.AddRegistryRoot` will skip creating a keyed `HttpClient` if `oauthScopes.Length == 0`.

Second, `RegistryServerProvider.CreateHttpClientAsync` also checks the `oauthScopes` JSON property, and it'll avoid the `if` branch that pulls a keyed `HttpClient` from DI.

This is fine as long as the code doesn't change, which is maybe an ok risk. Live tests should be able to catch any accidental regressions.

Alternatively, we could always add the keyed `HttpClient` into the DI and no-op if no scopes are given. One pro of that design is that it gives us a place to add other potential things on the outbound request, like if we want to add special headers. One con is more memory, but since these are currently singletons anyway, that's not a huge concern to me.

Totally your call. If you just want to get this PR in, I'd say go for it as-is. Revisit and iterate if you think this might be meaningful.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/SingleProxyToolLoaderTests.cs
```diff
@@ -15,6 +16,14 @@ namespace Azure.Mcp.Core.UnitTests.Areas.Server.Commands.ToolLoading;
 
 public class SingleProxyToolLoaderTests
 {
+    private static RegistryDiscoveryStrategy CreateStrategy(ServiceStartOptions options, ILogger<RegistryDiscoveryStrategy> logger)
```
vukelich, line 19: This `CreateStrategy` code is used a few times, and it'll be a pain to update each spot if the constructor to `RegistryDiscoveryStrategy` changes. Might be worth creating a helper method within the unit test project that adds semi-functional `RegistryDiscoveryStrategy`.

vukelich (changes requested):
Lots of comments. Sorry, I misused GitHub and posted PR comments rather than review comments. A good number of the comments are from pre-existing code that was moved elsewhere. For each of those that you don't want to fix, please comment explicitly saying so. Gives someone in the future looking at this PR historical confirmation about why code is the way it is.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: I am going to keep this as is for now since this is what the existing behavior.

core/Azure.Mcp.Core/src/Areas/Server/Models/IRegistryRoot.cs
JasonYeMSFT: The interface allows the Server to be nullable because the existing concrete class makes it nullable. I am going to leave it as is for now.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/SingleProxyToolLoaderTests.cs
JasonYeMSFT: I refactored the CreateStrategy function to be a shared helper method under Azure.Mcp.Core.UnitTests project.

core/Azure.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryDiscoveryStrategy.cs
JasonYeMSFT: Done.

core/Azure.Mcp.Core/src/Areas/Server/RegistryServerServiceCollectionExtensions.cs
JasonYeMSFT: Fixed. If registry or registry.Servers is null, an empty instance of RegistryRoot will be added to DI.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: Done.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: Fixed in your suggested way.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: As the time of this comment, the full filename that gets filtered out by this code is `Azure.Mcp.Core.Areas.Server.Resources.registry.json`.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: The unit tests can actually see this registry file. I'll keep this as is for now since the existing logic also returns null when the file cannot be found.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: No, there isn't and it's not what the typings indicate. I removed the null/empty check.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: I added a few unit tests to ensure the following behavior
1. Call the embedded IAzureTokenCredentialProvider
2. Overwrites existing auth header

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: I moved the AccessTokenHandler to its dedicated file.

core/Azure.Mcp.Core/src/Helpers/RegistryServerHelper.cs
JasonYeMSFT: For the near future, we can assume authors know to not define empty oauthScopes since it makes no sense. If that does beome a common mistake we can consider adding additional validation logic.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/AccessTokenHandlerTests.cs
```diff
@@ -0,0 +1,298 @@
...
+        await client.SendAsync(request, TestContext.Current.CancellationToken);
+
+        // Assert
+        await _mockTokenCredential.Received(1).GetTokenAsync(
+            Arg.Is<TokenRequestContext>(ctx => ctx.Scopes.SequenceEqual(customScopes)),
+            Arg.Any<CancellationToken>());
+    }
+
+    [Fact]
+    public async Task SendAsync_PassesCancellationTokenToTokenProvider()
```
vukelich: I like the spirit of this test. We want to make sure we don't drop `CancellationToken` notifications along the way. This is fine for now with how the current code is written, but `CancellationToken` structs can be tinkered with, linked together, replaced, etc. in valid ways across method calls. For example, let's say we had something that uses two `CancellationToken`s to bail out of async work: one from the caller and another from other system state or timeout logic.

Here's an AI generated example:

Prompt:
> I'm teaching someone how using unit tests to check for CancellationToken equality is fine but not ideal. My argument is that a method can do operations like linking to change what CancellationToken it effectively uses to call async methods it depends on. Show an example of this by creating this hypothetical system with a mini Program.cs.

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;

// === The dependency interface ===
public interface IDataFetcher
{
    Task<string> FetchDataAsync(CancellationToken cancellationToken);
}

// === The service under test ===
public class DataService
{
    private readonly IDataFetcher _fetcher;
    private readonly TimeSpan _timeout;

    public DataService(IDataFetcher fetcher, TimeSpan timeout)
    {
        _fetcher = fetcher;
        _timeout = timeout;
    }

    public async Task<string> GetDataWithTimeoutAsync(CancellationToken cancellationToken)
    {
        // Here's the problem: we create a LINKED token that combines:
        // 1. The caller's cancellation token
        // 2. A timeout-based cancellation
        using var timeoutCts = new CancellationTokenSource(_timeout);
        using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(
            cancellationToken, 
            timeoutCts.Token);

        // The token passed to FetchDataAsync is NOT the same as the one we received!
        return await _fetcher.FetchDataAsync(linkedCts.Token);
    }
}

// === Demo showing why token equality tests are fragile ===
public class Program
{
    public static async Task Main()
    {
        Console.WriteLine("=== CancellationToken Equality Demo ===\n");

        var cts = new CancellationTokenSource();
        var originalToken = cts.Token;

        // Create a mock that captures what token it receives
        var mockFetcher = new CapturingFetcher();
        var service = new DataService(mockFetcher, TimeSpan.FromSeconds(30));

        await service.GetDataWithTimeoutAsync(originalToken);

        // This is what a naive unit test might assert:
        Console.WriteLine($"Original token: {originalToken.GetHashCode()}");
        Console.WriteLine($"Received token: {mockFetcher.ReceivedToken.GetHashCode()}");
        Console.WriteLine($"Tokens are equal: {originalToken == mockFetcher.ReceivedToken}");
        Console.WriteLine();

        // But the behavior is still correct! Cancelling the original cancels the linked one:
        Console.WriteLine("Testing that cancellation still propagates...");
        cts.Cancel();
        Console.WriteLine($"Original is cancelled: {originalToken.IsCancellationRequested}");
        Console.WriteLine($"Received is cancelled: {mockFetcher.ReceivedToken.IsCancellationRequested}");
        Console.WriteLine();

        Console.WriteLine("=== Conclusion ===");
        Console.WriteLine("Token equality check FAILS, but the behavior is CORRECT.");
        Console.WriteLine("A better test would verify cancellation behavior, not token identity.");
    }
}

// === Simple mock that captures the token ===
public class CapturingFetcher : IDataFetcher
{
    public CancellationToken ReceivedToken { get; private set; }

    public Task<string> FetchDataAsync(CancellationToken cancellationToken)
    {
        ReceivedToken = cancellationToken;
        return Task.FromResult("data");
    }
}
```

vukelich: Another way to test is to focus on what we're trying to confirm: if the `CancellationToken` the caller provider is cancelled, then we should be reacting to it.

Fix or don't. Up to you. This feedback is expecting change that may never actually happen.

vukelich (approved):
LGTM. Skippable final comment mostly teaching about `CancellationToken`.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/AccessTokenHandlerTests.cs
JasonYeMSFT: Good comment. I'll do more experiment with this one later. If I cannot find a reasonable way to test the behavior I'll just remove this test to suppress potential false positives.

PR #1544, opened 2026-01-20
Bump `mcp-publisher` to align with `server.json` schema `2025-12-11`

scbedd:
Unbeknownst to me, this is a two part system.

1. You have `mcp-publisher.exe` that is installed in `Deploy-ServerJson.ps1`
2. You have a manifest file `server.json` which targets a schema via property `$schema`
3. Both must be updated. As the `mcp-publisher` needs to have the newer schema locally as well.

I failed to update mcp-publisher.exe version in my previous PR #1527 . And it's quite visible in the `publisher` tag: https://github.com/modelcontextprotocol/registry/releases/tag/v1.4.0

Going to check this against template release prior to merge.

### DIFF
--- a/eng/scripts/Deploy-ServerJson.ps1
+++ b/eng/scripts/Deploy-ServerJson.ps1
@@ -105,6 +105,8 @@ if (!(Test-Path $goInstallScriptPath)) {
 $env:GOEXPERIMENT = "systemcrypto"
 
 # Clone and build the publisher tool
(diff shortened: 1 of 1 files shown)


weshaggard (approved):

scbedd:
[all good](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5776860&view=results)

PR #1609, opened 2026-01-29
Add context-aware CORS policy to enhance security

g2vinay:
Add context-aware CORS policy for HTTP transport mode

- Development mode (unauthenticated): Restricts to localhost origins only to prevent CSRF attacks
  - Allows localhost/127.0.0.1/[::1] with any port (supports MCP Inspector, custom ports)
  - Protects against cross-site request forgery when --dangerously-disable-http-incoming-auth is used
  - Prevents malicious websites from accessing developer's Azure credentials via localhost endpoints

- Production mode (authenticated): Allows all origins safely
  - JWT Bearer authentication validates all requests regardless of origin
  - Required for MCP clients (GitHub Copilot in VS Code/Codespaces) from various origins
  - CORS is a browser mechanism; authenticated API doesn't need origin restrictions

Replaces permissive 'AllowAll' policy with context-aware 'McpCorsPolicy'
Implements ConfigureCors() method shared by both authenticated and unauthenticated HTTP hosts

### DIFF
--- a/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
+++ b/core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
@@ -526,7 +526,7 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
         // By default in development mode, we restrict to localhost origins for security.
         // In production (authenticated mode), allow configured origins or all origins if specified.
         // Non-browser clients are unaffected by CORS.
-        ConfigureCors(services, serverOptions);
+        ConfigureCors(services, builder.Environment, serverOptions);
 
         // Configure services
         ConfigureServices(services); // Our static callback hook
@@ -635,7 +635,7 @@ public sealed class ServiceStartCommand : BaseCommand<ServiceStartOptions>
         // By default in development mode, we restrict to localhost origins for security.
         // In production (authenticated mode), allow configured origins or all origins if specified.
         // Non-browser clients are unaffected by CORS.
-        ConfigureCors(services, serverOptions);
+        ConfigureCors(services, builder.Environment, serverOptions);
 
         // Configure services
         ConfigureServices(services); // Our static callback hook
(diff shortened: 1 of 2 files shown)


core/Azure.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -668,13 +652,69 @@ private IHost CreateIncomingAuthDisabledHttpHost(ServiceStartOptions serverOptio
...
+    /// In production (authenticated), allows all origins (safe due to authentication requirement).
+    /// </summary>
+    /// <param name="services">The service collection to configure.</param>
+    /// <param name="serverOptions">The server configuration options.</param>
+    private static void ConfigureCors(IServiceCollection services, ServiceStartOptions serverOptions)
+    {
+        // Check if running in development environment
+        string? environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
+        bool isDevelopment = string.Equals(environment, "Development", StringComparison.OrdinalIgnoreCase);
+
```
anuchandy, line 674: Can we check if the builder object at the call site can be used to determine if we're running in development mode? If it is possible then that would be more idiomatic for ASP.NET Core.

```C#
builder.Environment.IsDevelopment()
```

g2vinay: updated to use this API.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/ServiceStartCommandTests.cs
```diff
@@ -915,4 +916,315 @@ private static object GetAndAssertTagKeyValue(Activity activity, string tagName)
...
+
+        try
+        {
+            // Arrange environment
+            var environment = Substitute.For<IWebHostEnvironment>();
+            environment.EnvironmentName.Returns("Development");
+
+            // Act
+            var method = typeof(ServiceStartCommand).GetMethod("ConfigureCors",
+                System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
```
anuchandy, line 1042: An alternative to reflection would be making `ConfigureCors` internal and adding `[assembly: InternalsVisibleTo("Azure.Mcp.Core.UnitTests")]` to Azure.Mcp.Core's `AssemblyInfo.cs`. However, since the repo already uses reflection for similar test scenarios, I don't have a strong preference either way. Let's proceed with reflection as you suggested - just wanted to share this option for consideration.

anuchandy (approved):
thanks Vinay, lgtm

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/ServiceStartCommandTests.cs
g2vinay: if it becomes problematic at any point, we can pivot.

Issue #1457, opened 2026-01-08
CLI --help output | Please organize the output better


diberry:
Can the output of --help be organized better?

Suggested organization could be:

- group `CLI` contains `tools`
- group `MCP` contains `server` command
- group `recommended tools` contains `best practices`
- group `subscription management` contains `subscription, group`
- group `tools` contains individual tools 

@sandeep-sen 

<img width="1788" height="1632" alt="Image" src="https://github.com/user-attachments/assets/f029420a-603a-4b65-a863-007d759c19d6" />

joshfree:
Use Copilot:

1. max line length <= 72 characters
2. greatly reduce the help text for all of the commands.  Be concise and clear.

Add a regression test to enforce 1 and 2 above.

PR #1640, opened 2026-02-03
Organize CLI --help output

anannya03:
## What does this PR do?

1. Organize the CLI help output in the following categories- 
group `CLI` contains tools
group `MCP` contains server command
group `recommended tools` contains best practices
group `subscription management` contains subscription, group
group `Azure Services` contains individual tools

2. Add a section for `Example` which has 1-2 examples of using the commands.

image

3. Reduce the help output of the commands to be clear and concise. 
4. Enforce a strict <=72 characters per line limit for description in the CLI --help output.

## GitHub issue number?
https://github.com/microsoft/mcp/issues/1457


eng/scripts/Test-DescriptionLineLength.ps1
```diff
@@ -0,0 +1,206 @@
...
+        Write-Host "  $fixed"
+    }
+    Write-Host ""
+    Write-Host "`nReview the changes to check indentation." -ForegroundColor Yellow
+}
+else {
+    Write-Host "Run with -Fix to attempt automatic wrapping" -ForegroundColor Cyan
+}
+
+exit 1
```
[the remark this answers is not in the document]
anannya03: fixed.

core/Azure.Mcp.Core/src/Areas/Group/GroupSetup.cs
```diff
@@ -14,14 +14,20 @@ public sealed class GroupSetup : IAreaSetup
...
         services.AddSingleton<GroupListCommand>();
     }
 
     public CommandGroup RegisterCommands(IServiceProvider serviceProvider)
     {
-        var group = new CommandGroup(Name, "Resource group operations - Commands for listing and managing Azure resource groups in your subscriptions.", Title);
+        var group = new CommandGroup(Name,
+            """
+            Resource group operations - Commands for listing and managing Azure
+            resource groups in your subscriptions.
```
alzimmermsft: This will have the output string be multi-line, is that what we want?

I'm guessing all these were split to remain until 80 characters, is that right? But we should be doing 80 characters based on the description string, not the line length in IDE.

Also, is this even needed given there is logic to format commands and their descriptions in `CustomHelpAction.cs`? If not, could we revert all these description string changes to make this PR way simpler?

core/Microsoft.Mcp.Core/src/Commands/CustomHelpAction.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private static readonly Dictionary<CommandCategory, string> _categoryNames = new()
+    {
+        [CommandCategory.Cli] = "CLI",
+        [CommandCategory.Mcp] = "MCP",
+        [CommandCategory.RecommendedTools] = "Recommended Tools",
+        [CommandCategory.SubscriptionManagement] = "Subscription Management",
+        [CommandCategory.AzureServices] = "Azure Services"
+    };
```
alzimmermsft: I think a switch statement here would be better as that should make it more obvious to update this if more `CommandCategory`s are added.

eng/scripts/Test-DescriptionLineLength.ps1
alzimmermsft, line 1: Do we need this if we can handle wrapping descriptions before we output them?

core/Azure.Mcp.Core/src/Areas/Group/GroupSetup.cs
anannya03: Ideally, since we are wrapping the description up at 72 characters in CustomHelpAction, we don't really need to change all the descriptions. 
Makes sense, I'll revert the description length changes, and let the help action handle that. There'd still be some changes to the description in an effort to make it slightly concise, but that's just for better clarity.

eng/scripts/Test-DescriptionLineLength.ps1
anannya03: Nope. I'll revert the .ps1 file changes as well.

anannya03: done.

core/Microsoft.Mcp.Core/src/Commands/CustomHelpAction.cs
anannya03: I've added a switch case. Thanks!

core/Azure.Mcp.Core/src/Areas/Group/GroupSetup.cs
anannya03: I've now reverted these changes.

alzimmermsft (approved):

PR #1651, opened 2026-02-04
Multi-Architecture Docker Image Support

anuchandy:
### Summary
Adds ARM64 architecture support to Docker image publishing, creating multi-arch manifests that allow Docker to automatically pull the correct architecture for the host platform.

### Changes
* **Build pipeline**: Both `linux/amd64` and `linux/arm64` Docker images are now built in parallel using architecture-specific pools
* **Release pipeline**: Creates multi-arch manifests combining both architectures under single tags (`:latest`, `:<version>`)

### What it enables

In addition to the currently supported x64 architecture, this enables Azure MCP Server to run on ARM64 infrastructure across the Azure services:

| Service | ARM64 Docker Support |
| ------- | -------------------- |
| Azure Kubernetes Service (AKS) | ARM64 node pools (Ampere Altra VMs) |
| Azure VMs | Dpsv5, Dplsv5, Epsv5 series (Ampere Altra) |

#### Multi-arch manifest inspection:

```
$ docker manifest inspect azuresdkimages.azurecr.io/anu/azure-sdk/azure-mcp:latest
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "digest": "sha256:d5a7b84c92120ea15aeafa488c9f68ac50452b6eb5f2bb95821348fdcf695004",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "digest": "sha256:ebeb8252fcfeee569fbee1adcdeecb67a4b4ca83f79ed200d9a2b1ec6f6b183d",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

#### Docker pull on macOS ARM:

```
$ docker pull azuresdkimages.azurecr.io/anu/azure-sdk/azure-mcp:latest

$ docker inspect azuresdkimages.azurecr.io/anu/azure-sdk/azure-mcp:latest --format '{{.Os}}/{{.Architecture}} '

    linux/arm64

```

#### ACR repository:

image

### Out of Scope
Native Windows or macOS Docker images are intentionally excluded. linux/amd64 + linux/arm64 covers the vast majority of production use cases—Windows/macOS Docker images are rarely used in practice.

## GitHub issue number?

https://github.com/microsoft/mcp/issues/1659

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:


eng/pipelines/templates/jobs/docker/build-docker.yml
```diff
@@ -12,8 +12,8 @@ jobs:
   strategy:
     matrix: $[ ${{ parameters.ServerMatrix }}]
   pool:
-    name: $(LINUXPOOL)
-    image: $(LINUXVMIMAGE)
+    name: $(Pool)
+    image: $(VMImage)
```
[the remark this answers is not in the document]
anuchandy: `Prepare-Docker.ps1` now handles artifacts for both supported platforms.

```ps1
$dockerPlatforms = @("linux-musl-x64-docker", "linux-musl-arm64-docker")
```

eng/scripts/New-BuildInfo.ps1
```diff
@@ -593,23 +601,55 @@ function Get-ServerMatrix {
...
-            ExecutableName = $server.cliName + $executableExtension
-            DockerLocalTag = $imageName + ":" + $BuildId
+
+        foreach ($archConfig in $dockerArchConfigs) {
+            if ($archConfig.Architecture -eq 'arm64' -and $arm64DockerServers -notcontains $server.name) {
+                # Skip ARM64 for MCP servers not in the opt-in list
+                continue
+            }
+
+            $platform = $server.platforms | Where-Object { $_.name -eq $archConfig.PlatformName -and -not $_.native }
```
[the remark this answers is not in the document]
anuchandy: Given the current logic of `New-BuildInfo.ps1` where both docker platforms are defined $additionalPlatforms and always added, `$platform` will not be null.

eng/scripts/Publish-DockerImages.ps1
```diff
@@ -0,0 +1,259 @@
...
+    
+    Write-Host "Discovering tar files..."
+    $tarPattern = Join-Path $TarDirectory "$CliName-*-image.tar"
+    $tarFiles = Get-ChildItem -Path $tarPattern -ErrorAction SilentlyContinue
+    
+    if (-not $tarFiles -or $tarFiles.Count -eq 0) {
+        Write-Host "ERROR: No tar files found matching pattern $tarPattern" -ForegroundColor Red
+        Write-Host "Directory contents:"
+        Get-ChildItem -Path $TarDirectory | ForEach-Object { Write-Host "  $_" }
+        exit 1
```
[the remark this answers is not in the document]
anuchandy: I agree we should use `Write-Error` instead of `Write-Host "ERROR: .."`. However, in that case we don't need to explicitly exit or throw since `$ErrorActionPreference = 'Stop'` is set for this script - `Write-Error` will terminate execution automatically. I'll change to use `Write-Error`

eng/scripts/Publish-DockerImages.ps1
```diff
@@ -0,0 +1,259 @@
...
+}
+
+function New-MultiArchManifest {
+    param(
+        [string]$ManifestTag,
+        [string[]]$ArchTags
+    )
+    
+    Write-Host "Creating multi-arch manifest for $ManifestTag..."
+    Invoke-DockerCommand -Arguments (@('manifest', 'create', $ManifestTag) + $ArchTags)
```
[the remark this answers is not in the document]
anuchandy: Since we're using Microsoft-hosted pools, reruns always get fresh agents, so the local manifest is created anew each time - making the `--amend` flag technically unnecessary. However, I'm considering local dev scenarios where `--amend` makes sense and is low-risk. I'll update the PR.

eng/scripts/Prepare-Docker.ps1
```diff
@@ -55,40 +55,48 @@ Remove-Item -Path $OutputPath -Recurse -Force -ErrorAction SilentlyContinue -Pro
 Push-Location $RepoRoot
 try {
     $buildInfo = Get-Content $BuildInfoPath -Raw | ConvertFrom-Json
-    $platformName = "linux-musl-x64-docker"
+    $dockerPlatforms = @("linux-musl-x64-docker", "linux-musl-arm64-docker")
```
conniey: Do we need this here? I figured we would read these values from the build_info.json

eng/scripts/Publish-DockerImages.ps1
```diff
@@ -0,0 +1,229 @@
+<#
```
conniey, line 1: This feels like it's doing more than publishing, it's also writing this manifest item? I feel like Publish-DockerImages, should just be "push x,y,z to docker" rather than computing all these values, writing a manifest, etc.

Could "Prepare-Docker" contain all this logic?

eng/pipelines/templates/jobs/docker/build-docker.yml
```diff
@@ -12,8 +12,8 @@ jobs:
   strategy:
```
conniey, line 12: I'm a bit surprised that we didn't have to update the dockerfile to support this.

eng/scripts/Prepare-Docker.ps1
anuchandy: Today we don't write these values to build_info.json as key-value pairs indicating the Docker platform types, though these are two string constants used for different MCP Server platform names in build-info. I think it makes sense to write to and read from build_info.json. @hallipr , what do you think - does it make sense to add an entry in `build_info.json`?

```json
"dockerPlatforms": [
  "linux-musl-x64-docker", "linux-musl-arm64-docker"
]
```

eng/scripts/Publish-DockerImages.ps1
anuchandy: The docker `manifest create/push` commands are Docker's native way to publish multi-arch images - "Publish" in Docker context includes tagging, pushing, and manifest creation.

We need all these operations in the context of a single Docker daemon (load from tars -> tag -> push arch images -> create manifest -> push manifest). Originally this was an inline bash script in `release-docker.yml` that handled load, tag, and push for a single image. With multi-arch support (2 images now), the inline script route became unreadable, so I extracted it into a dedicated script that `release-docker.yml` invokes. So two steps

* Build jobs: build image -> save to tar -> stage tar artifact
* Release job: download tars -> load image into local daemon, tag -> push, create manifest -> push (aka publish, needs to happen in one daemon)

eng/pipelines/templates/jobs/docker/build-docker.yml
anuchandy: Good point! We don't use this PS1 in CI - it's for local development. It seems to have an entry for linux/arm64 i will need to try it

anuchandy: My bad- I mistook `build-docker.yml` (which you commented on) for `build-docker.ps1`, which we use locally. 

But yes, we didn't need to change the `Dockerfile`. There are two arch-specific elements in that file: the base image and the azmcp binary. 

1. The base image "mcr.microsoft.com/dotnet/aspnet:10.0-alpine" that we're using is multi-arch, so the user's Docker client will automatically pull the correct architecture based on the host machine. 
2. As for the azmcp binary, `Prepare-Docker.ps1` sets the correct dist/ directory containing the azmcp built for the same arch we're building docker image for.

eng/scripts/Prepare-Docker.ps1
anuchandy: Thinking about it a bit more, we rarely add new architectures - ARM64 docker support took years to become mainstream, so the chances of needing another entry are almost nil. That said, I'm fine either way: duplicating and hardcoding here (as it was before my change) vs. change build-info and read from there

anuchandy: I changed it to read from `build_info.jon` instead of hardcoding in `Prepare-Docker.ps1`.  This ensures that constant naming only needs to be updated in one place - for example, if we ever switch libc from musl to glibc and want these constant string to reflect that change.

Reg arch, If we ever need to support a new architecture for Docker, the steps would be:

1. Work with agent pool owners to get a native agent for that new architecture to build the Docker image, and define it in `image.yml`. Alternatively, we could cross-arch docker build - assuming we can enable QEMU emulation on an existing agent and the version of Docker supports QEMU emulation based build for that architecture.
2. Update `New-BuildInfo.ps1` to include - new pool, image and const `linux-musl-{arch}-docker`.
3. Add a new input path in `release-docker.yml` for that architecture's Docker artifacts.

(Outside of the Docker build, we'd also need to be able to do a .NET build to produce the azmcp binary for that architecture.)

eng/pipelines/templates/jobs/docker/build-docker.yml
```diff
@@ -5,15 +5,15 @@ parameters:
   type: string
 
 jobs:
-- job: BuildDockerImages
-  displayName: "Build Docker Images"
+- job: BuildDockerImage
```
hallipr, line 8: If the server matrix contains more than one server, we'll still be building multiple images.   I'd keep it plural

anuchandy: Right, when there are multiple entries in the matrix, we call this template for each Docker platform separately. So there's one template call for the arm64 image and another for amd64. In the DevOps pipeline, this results in two tasks rendered in UI as: 

* "Build Docker Images Azure.Mcp.Server_amd64"
* "Build Docker Images Azure.Mcp.Server_arm64"

I noticed the job name uses plural **"Images"** even though each job only builds a single image. This initially confused me, so I changed it to singular. However, if our EngSys pattern is to keep it plural, I'm happy to revert this change. Just wanted to share my reasoning and get your take on it.

anuchandy: discussed offline, we're good.

PR #1751, opened 2026-02-18
Fix  error propagation about missing param back to the LLM/user.

g2vinay:
Fixes the issue where missing parameter wasn't being propagated in the response to the LLM.

### DIFF
--- a/servers/Azure.Mcp.Server/changelog-entries/1771830706972.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1771830706972.yaml
@@ -1,3 +1,3 @@
 changes:
   - section: "Bugs Fixed"
-    description: "Fixed error handling test to ensure error messages are preserved for missing parameters"
+    description: "Fixed error handling to ensure error messages are preserved for missing parameters"


core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
```diff
@@ -556,6 +589,35 @@ await Assert.ThrowsAsync<ArgumentNullException>(async () =>
             await options.Handlers.ElicitationHandler.Invoke(null!, TestContext.Current.CancellationToken));
     }
 
+    [Fact]
+    public async Task CallToolHandler_WithMissingRequiredOptions_PreservesSpecificErrorMessage()
+    {
+        // Arrange
+        // Service Bus commands require --subscription parameter
```
joshfree: are there any other tools that have a required --subscription parameter that you should add tests for at the same time?

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -380,13 +380,19 @@ private async Task<CallToolResult> InvokeChildToolAsync(
                 var childToolSpecJson = GetChildToolJson(request, namespaceName, command);
 
                 _logger.LogWarning("Namespace {Namespace} command {Command} requires additional parameters.", namespaceName, command);
+
+                // Extract the specific error message from the response
+                var errorMessage = string.IsNullOrEmpty(commandResponse.Message)
```
joshfree, line 385: this needs a CHANGELOG entry

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
g2vinay: This is actually, going to fix it for all the tools.
Servicebus just became the source to trace this bug.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
g2vinay: added

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -380,13 +380,19 @@ private async Task<CallToolResult> InvokeChildToolAsync(
...
+                    : commandResponse.Message;
+
                 var finalResponse = new CallToolResult
                 {
                     Content =
                     [
                         new TextContentBlock {
                                 Text = $"""
-                                    The '{command}' command is missing required parameters.
+                                    {errorMessage}
```
anuchandy, line 395: Trying to understand - so previously we never bubbled the actual error message to the LLM (whether one existed or not), and instead always returned the generic text _"The .. command is missing required parameters."_ Now, if an error message exists we return it; otherwise we fall back to the same generic text.

1. So, if a command throws a `System.ArgumentNullException` when a param missing, the error message would be "Value cannot be null. (Parameter 'foo')" - which is  more actionable for the LLM. Is this understanding correct?
2. nit; Should we update the pr title as this is a generic fix

g2vinay: Yeah previously, we just sent back a generic message, now we extract the info about missing param name and wire it back to the LLM as Missing required options: subscription id etc, if the error contains the info.

g2vinay: updated PR title.

PR #1771, opened 2026-02-19
Migrate Foundry MCP server to use external integration.

g2vinay:
This PR migrates Microsoft Foundry from in-process tool files to the external MCP server
at `https://mcp.ai.azure.com`, then adds the authentication plumbing needed for the Azure MCP
server to call that external server on the user's behalf.

Changes:

1. Foundry → external MCP server (`registry.json`)
   All Foundry source files (commands, services, options, models, tests, setup) have been deleted.
   Instead, `registry.json` now contains a `"foundry"` entry that points the runtime at the hosted
   server:

   The Foundry project references have also been removed from both `Microsoft.Mcp.slnx` and
   `Azure.Mcp.Server.slnx`.

3. `AccessTokenHandler` — `TokenCredential`-based constructor

    A second constructor was added so the registry HTTP client can be wired directly to a
    `TokenCredential` instance rather than always going through `IAzureTokenCredentialProvider`:

4. `CustomChainedCredential` — `forceBrowserFallback` parameter

    A new `forceBrowserFallback` constructor parameter allows callers to opt into
    `InteractiveBrowserCredential` as a last resort without requiring `AZURE_TOKEN_CREDENTIALS` to 
    be unset globally. The flag is suppressed when `AZURE_TOKEN_CREDENTIALS=prod`, so CI/headless
    environments are never interrupted by a browser popup.

5. `SafeTokenCredential` — `normalizeScopes` parameter
 
    Non-MSAL credentials (`AzureCliCredential`, `ManagedIdentityCredential`, etc.) use a
    resource-based auth model: they expect a scope of the form `https://host/.default` and strip the
    `/.default` suffix internally to derive the resource URL. They cannot handle arbitrary MSAL
    permission scopes like `https://mcp.ai.azure.com/Foundry.Mcp.Tools`.

    `SafeTokenCredential` now accepts `normalizeScopes: true`, which transparently converts any
    non-`/.default` scope to its `resource/.default` form before forwarding to the inner credential.


core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -150,8 +150,13 @@ private static TokenCredential CreateCredential(string? tenantId, ILogger<Custom
         // 1. AZURE_TOKEN_CREDENTIALS is not set (default behavior)
         // 2. AZURE_TOKEN_CREDENTIALS explicitly requests it
         // 3. AZURE_TOKEN_CREDENTIALS="dev" (development credentials with interactive fallback)
+        // 4. forceBrowserFallback=true AND AZURE_TOKEN_CREDENTIALS is not "prod"
+        //    (registry server callers want interactive fallback for local/hosting-identity scenarios,
+        //     but must respect "prod" which means headless/CI — no browser popup)
         // Do NOT add it for "prod" or specific credential names (user wants only those credentials)
```
JasonYeMSFT: If tokenCredentials is a specific credential (but not "prod") and forceBrowserFallback is true the browser fallback will be added now.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT, line 63: This doc string needs to be updated to reflect the new branching condition to determine whether to add the browser fallback.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: yeah, that's the intent, only for registry.json, auth browser (it will silent auth using default broker account) fallback will be true.
if user has a specific credential set and that doesn't work, we will catch the fallback in that case, to ensure we auth with the fallback auth and list the foundry tools.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -111,7 +111,6 @@ private static IAreaSetup[] RegisterAreas()
             new Azure.Mcp.Tools.ConfidentialLedger.ConfidentialLedgerSetup(),
             new Azure.Mcp.Tools.EventHubs.EventHubsSetup(),
             new Azure.Mcp.Tools.FileShares.FileSharesSetup(),
-            new Azure.Mcp.Tools.Foundry.FoundrySetup(),
```
xiangyan99, line 117: Do we need FoundryExtensionsSetup?

g2vinay: Yeah, I already pushed it above.

Microsoft.Mcp.slnx
```diff
@@ -237,14 +237,7 @@
...
   </Folder>
-  <Folder Name="/tools/Azure.Mcp.Tools.Foundry/" />
-  <Folder Name="/tools/Azure.Mcp.Tools.Foundry/src/">
-    <Project Path="tools/Azure.Mcp.Tools.Foundry/src/Azure.Mcp.Tools.Foundry.csproj" />
-  </Folder>
-  <Folder Name="/tools/Azure.Mcp.Tools.Foundry/tests/">
-    <Project Path="tools/Azure.Mcp.Tools.Foundry/tests/Azure.Mcp.Tools.Foundry.LiveTests/Azure.Mcp.Tools.Foundry.LiveTests.csproj" />
-    <Project Path="tools/Azure.Mcp.Tools.Foundry/tests/Azure.Mcp.Tools.Foundry.UnitTests/Azure.Mcp.Tools.Foundry.UnitTests.csproj" />
-  </Folder>
+
```
xiangyan99, line 249: Do we need to include the FoundryExtensions project?

JasonYeMSFT: Probably not a huge deal functionality wise since this only affects if the project shows up in VS. If you don't use VS you may not even notice this.

g2vinay: Yeah, wired it now.

xiangyan99 (approved):
Maybe you also want to include the project in servers/Azure.Mcp.Server/Azure.Mcp.Server.slnx. :)

PR #1908, opened 2026-03-03
add DeviceCodeCredential support for CLI Usage.

g2vinay:
Adds DeviceCode authentication support, enabling Azure MCP to authenticate in headless environments (Docker, WSL, SSH tunnels, CI) or any situation/ env a broker auth is unavailable. Device Code auth doesn't come into play if server is running in stdio / http mode.

Supports device code Authentication when:
1. AZURE_TOKEN_CREDENTIALS env var is set to DeviceCodeCredential.
2. All forms of auth including browser Broker auth is not available. Device code auth is the fallback in that case.

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
+++ b/core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
@@ -29,7 +29,7 @@ namespace Azure.Mcp.Core.Services.Azure.Authentication;
 /// </listheader>
 /// <item>
 /// <term>"dev"</term>
-/// <description>Visual Studio → Visual Studio Code → Azure CLI → Azure PowerShell → Azure Developer CLI → InteractiveBrowserCredential</description>
+/// <description>Visual Studio → Visual Studio Code → Azure CLI → Azure PowerShell → Azure Developer CLI → InteractiveBrowserCredential → DeviceCodeCredential (CLI mode; fallbacks suppressed in server transport mode)</description>
 /// </item>
 /// <item>
 /// <term>"prod"</term>
@@ -37,7 +37,7 @@ namespace Azure.Mcp.Core.Services.Azure.Authentication;
 /// </item>
 /// <item>
 /// <term>"DeviceCodeCredential"</term>
-/// <description>Device code flow — prints a URL and one-time code to stdout; works in headless environments (Docker, WSL, SSH, CI)</description>
+/// <description>Device code flow — displays a URL and one-time code on the console; works in headless environments (Docker, WSL, SSH, CI). Not available in server transport mode (stdio/http).</description>
 /// </item>
 /// <item>
 /// <term>Specific credential name</term>
@@ -45,7 +45,7 @@ namespace Azure.Mcp.Core.Services.Azure.Authentication;
 /// </item>
 /// <item>
 /// <term>Not set or empty</term>
-/// <description>Development chain (Environment → Visual Studio → Visual Studio Code → Azure CLI → Azure PowerShell → Azure Developer CLI) + InteractiveBrowserCredential fallback</description>
+/// <description>Development chain (Environment → Visual Studio → Visual Studio Code → Azure CLI → Azure PowerShell → Azure Developer CLI) + InteractiveBrowserCredential + DeviceCodeCredential as last-resort fallbacks (CLI mode only; both suppressed in server transport mode)</description>
 /// </item>
 /// </list>
 /// <para>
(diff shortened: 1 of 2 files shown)


core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -178,6 +189,19 @@ private static TokenCredential CreateCredential(string? tenantId, ILogger<Custom
             creds.Add(CreateBrowserCredential(tenantId, authRecord));
         }
 
+        // Add DeviceCodeCredential as a fallback for headless environments (Docker, WSL, SSH, CI)
+        // when the default or dev chain is active. Unlike InteractiveBrowserCredential it only needs
+        // a terminal, not a GUI browser. Only added in CLI mode (ActiveTransport empty) because in
+        // server mode stdout is the MCP protocol pipe (stdio) or there is no attached terminal (http).
+        bool shouldAddDeviceCodeFallback = !isPinnedCredentialMode &&
```
xiangyan99, line 200: Is DeviceCodeCredential  a pinned credential?

I see

"Any explicit AZURE_TOKEN_CREDENTIALS value other than "dev" or "InteractiveBrowserCredential"
is treated as a pinned credential choice — interactive browser must not be injected."

in our comments.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -178,6 +189,19 @@ private static TokenCredential CreateCredential(string? tenantId, ILogger<Custom
...
+                                          !isExplicitBrowserMode &&
+                                          string.IsNullOrEmpty(ActiveTransport);
+
+        if (shouldAddDeviceCodeFallback)
+        {
+            AddDeviceCodeCredential(creds, tenantId);
```
xiangyan99, line 206: If I understand correctly, in the case that uses DeviceCodeCredential, it should be the only credential in the chain, right?

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: yeah, if AZURE_TOKEN_CREDENTIALS is set to DeviceCodeCredential, then it is pinned to it, in that case, we don't need to add any device code fallback, because it is already pinned to use Device code auth.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: if AZURE_TOKEN_CREDENTIALS is set to device code auth yeah, it will be the only credential in the chain.
This block won't execute to add as fallback.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
xiangyan99: Then in the DeviceCodeCred case, isPinnedCredentialMode  should be true?

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -423,6 +451,49 @@ private static void AddAzureDeveloperCliCredential(List<TokenCredential> credent
...
+        string? authRecordJson = Environment.GetEnvironmentVariable(AuthenticationRecordEnvVarName);
+        if (!string.IsNullOrEmpty(authRecordJson))
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(authRecordJson);
+            using MemoryStream stream = new(bytes);
+            deviceCodeOptions.AuthenticationRecord = AuthenticationRecord.Deserialize(stream);
+        }
+
+        credentials.Add(new SafeTokenCredential(new DeviceCodeCredential(deviceCodeOptions), "DeviceCodeCredential"));
+    }
```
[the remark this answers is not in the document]
g2vinay: Callback is optional, by default it writes to the terminal.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: yeah it will be true.

g2vinay: IsPinnedCredentialMode will be true, if AZURE_TOKEN_CREDENTIALS is set Device code Cred, in that case, device code fallback will not be added. 
and just the device code credential will be added to the chain by the CreateDefaultCredential helper beforehand.

xiangyan99: Do we need != "prod"?

g2vinay: logically, nope, as prod puts pinned credential mode to true and device code fallback requries pinned credential mode to be false.
But for readability, I think we can add it explicitly, else it can confuse the reader/maintainer later.

PR #1948, opened 2026-03-05
Integrate azure load testing into CI

xiangyan99:
## What does this PR do?

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:


eng/pipelines/load-test.yml
```diff
@@ -0,0 +1,196 @@
...
+            displayName: ACR Build & Push
+            inputs:
+              azureSubscription: $(AZURE_SERVICE_CONNECTION)
+              scriptType: bash
+              scriptLocation: inlineScript
+              inlineScript: |
+                az acr build \
+                  --registry $(AZURE_ACR_NAME) \
+                  --image $(imageName):$(imageTag) \
+                  --file servers/Azure.Mcp.Server/src/Dockerfile \
```
jongio: The Dockerfile at this path doesn't exist - it lives at the repo root (`./Dockerfile`). Fixing the path alone isn't enough though: that Dockerfile is a post-publish wrapper and requires `--build-arg PUBLISH_DIR=...` and `--build-arg EXECUTABLE_NAME=...`, plus the published binary already staged in `PUBLISH_DIR`. You need either a `dotnet publish` step before this that passes both build args through, or a different Dockerfile that builds from source. See `eng/scripts/Build-Docker.ps1` for how the existing build invokes it.

jongio (commented):
A few things to sort out before this can actually run end-to-end.

1. The pipeline won't build the image. `az acr build --file servers/Azure.Mcp.Server/src/Dockerfile .` points at a path that doesn't exist - the Dockerfile is at the repo root (`./Dockerfile`). That Dockerfile is also a post-publish packaging step: it requires `--build-arg PUBLISH_DIR=...` and `--build-arg EXECUTABLE_NAME=...` plus a prior `dotnet publish` to produce the binary. See `eng/scripts/Build-Docker.ps1` for the normal flow. As written, Stage 1 fails every run.

2. The Container App won't pull the image. The Bicep gives the Container App a system-assigned identity and references ACR with `identity: 'system'`, but doesn't assign AcrPull to it. Same pattern is done correctly in `tools/Azure.Mcp.Tools.Acr/tests/test-resources.bicep` lines 37-48.

3. The pipeline parameters don't do anything. `engineInstances`, `users`, `spawnRate`, `duration` are declared with queue-time defaults, but nothing wires them into the run - `config.yaml` hardcodes all four values and the `AzureLoadTest@1` task doesn't take parameter overrides.

4. Title and PR body are out of sync with actual behavior. Title says `into CI` but the pipeline is `trigger: none` / `pr: none` - manual only. The README gets this right; the title and PR description don't. The PR body is also still the unfilled template.

Smaller things inline.

eng/load-tests/load-test-resources.bicep
```diff
@@ -0,0 +1,139 @@
...
+      logAnalyticsConfiguration: {
+        customerId: logAnalytics.properties.customerId
+        sharedKey: logAnalytics.listKeys().primarySharedKey
+      }
+    }
+  }
+}
+
+// ---------- Container App (MCP Server – auth disabled) ----------
+resource containerApp 'Microsoft.App/containerApps@2024-03-01' = {
```
jongio, line 86: The Container App uses `identity: 'system'` to pull from `$(acrName).azurecr.io`, but the system-assigned managed identity never gets AcrPull. On first deploy the image pull fails. Add a `Microsoft.Authorization/roleAssignments` with `roleDefinitionId` `7f951dda-4ed3-4680-a7ca-43fe172d538d` scoped to the ACR, targeting `containerApp.identity.principalId`. `tools/Azure.Mcp.Tools.Acr/tests/test-resources.bicep:37-48` shows the pattern used elsewhere in this repo.

eng/pipelines/load-test.yml
```diff
@@ -0,0 +1,196 @@
...
+#   - AZURE_LOAD_TEST_RESOURCE_GROUP   – resource group (ALT + ACA + ACR)
+#   - AZURE_LOAD_TEST_NAME            – name of the Azure Load Testing resource
+#   - AZURE_ACR_NAME                  – Azure Container Registry name (e.g. mcploadtestacr)
+#   - MCP_SUBSCRIPTION                – Azure subscription used in the tool call
+# ---------------------------------------------------------------------------
+
+trigger: none
+pr: none
+
+parameters:
```
jongio, line 24: These four parameters (`engineInstances`, `users`, `spawnRate`, `duration`) are declared but never used. `config.yaml` hardcodes `engineInstances: 1`, `users: 10`, `spawnRate: 2`, `runTime: 120s`, and the `AzureLoadTest@1` task doesn't take overrides. Either template-substitute the parameters into `config.yaml` before the load test step runs (e.g. a `sed`/`envsubst` step), or drop them from the pipeline so nobody thinks they're configurable.

eng/load-tests/load-test-resources.bicep
```diff
@@ -0,0 +1,139 @@
...
+            memory: '2Gi'
+          }
+          probes: [
+            {
+              type: 'Readiness'
+              httpGet: {
+                path: '/health'
+                port: 8080
+              }
+              initialDelaySeconds: 5
```
jongio: No startup probe, and readiness `initialDelaySeconds: 5` is aggressive for a .NET cold start - the server likely isn't serving `/health` within 5 seconds of container start, so the first few readiness checks will fail. Either add a startup probe with a generous `failureThreshold`, or bump readiness `initialDelaySeconds` to ~20-30s. Otherwise the first load test run may hit the 5-minute health wait ceiling in the pipeline.

eng/load-tests/config.yaml
```diff
@@ -0,0 +1,39 @@
...
+testPlan: locustfile.py
+
+engineInstances: 1
+
+failureCriteria:
+  - avg(response_time_ms) > 5000
+  - percentage(error) > 10
+
+env:
+  - name: MCP_SERVER_URL
```
jongio: `MCP_SERVER_URL` is declared here and mentioned in the locustfile docstring, but `locustfile.py` never reads it - Locust uses the `host` field from this file (or `--host` on the CLI) as the base URL for `self.client.post(`/mcp`, ...)`. The env var is dead. Either drop it or actually read it in the locustfile.

eng/load-tests/locustfile.py
```diff
@@ -0,0 +1,133 @@
...
+        with self.client.post(
+            "/mcp",
+            json=payload,
+            headers={"Content-Type": "application/json", "Accept": "application/json, text/event-stream"},
+            name="initialize",
+            catch_response=True,
+        ) as resp:
+            if resp.status_code == 200:
+                self.session_id = resp.headers.get("Mcp-Session-Id")
+                if not self.session_id:
```
jongio: If the initialize response comes back without `Mcp-Session-Id`, `session_id` stays `None` and every subsequent `tools/call` from this user goes out without the session header and gets rejected by the server. For a load test that's arguably fine (failures show up in results), but if the goal is to measure the steady-state tool-call path, consider stopping the user here or retrying the handshake, otherwise a session-handshake regression shows up as tool-call errors and gets misdiagnosed.

eng/pipelines/load-test.yml
```diff
@@ -0,0 +1,238 @@
...
+            inputs:
+              azureSubscription: $(AZURE_SERVICE_CONNECTION)
+              scriptType: bash
+              scriptLocation: inlineScript
+              inlineScript: |
+                echo "Removing image $(imageName):$(imageTag) from ACR..."
+                az acr repository delete \
+                  --name $(AZURE_ACR_NAME) \
+                  --image $(imageName):$(imageTag) \
+                  --yes || true
```
jongio, line 254: Cleanup deletes the Container App and the ACR image tag, but each run also creates a user-assigned managed identity (`${containerAppName}-uami`) and an AcrPull role assignment scoped to the ACR. Neither is cleaned up here, so they accumulate across runs.

Add a step to delete the identity (which cascade-removes its role assignments):

```bash
# Delete the user-assigned managed identity
az identity delete \
  --name $(containerAppName)-uami \
  --resource-group $(AZURE_LOAD_TEST_RESOURCE_GROUP) || true
```

jongio (commented):
Addresses all six of my previous inline comments - the build, identity, parameter wiring, probes, dead env var, and session handling are all fixed. The user-assigned identity + AcrPull pattern is the right call over system-assigned.

Two items from my earlier review body are still open:

1. PR description is still the unfilled template. Fill it in before this leaves draft.
2. Title says "into CI" but the pipeline is manual-only (trigger: none / pr: none). Something like "Add on-demand load testing pipeline" would match the actual behavior.

One new item inline.

jongio (commented):
My inline comment about the cleanup stage leaking user-assigned managed identities - addressed in ecaf912. The cascade-delete comment is a nice touch.

All technical findings from both my reviews are now resolved. Two non-blocking items remain from the PR description:
- PR description is still the template placeholder (once this moves out of draft, a real description helps reviewers and future archaeology)
- Title says "into CI" but this is a manually-triggered pipeline, not part of the CI gate

jongio (commented):
Solid iteration. 6 new commits since my last review, all improvements:

- `BUILD_ID` in testId/displayName prevents test collisions across concurrent pipeline runs
- `MCP_PATH` env var makes the endpoint configurable with backwards-compatible `/mcp` default
- Response parsing in locustfile.py is much cleaner. The old `except: pass` is gone, content-type is checked before parsing, and parse failures are explicitly reported
- `--name mcp-loadtest-$(Build.BuildId)` prevents ARM deployment name conflicts
- `loadTestName` wired from pipeline variable instead of relying on Bicep default
- curl format string fix (`%%{http_code}` to `%{http_code}`). Double percent was a bug in bash context

No new issues. Prior findings all remain addressed. Still draft/non-blocking notes from before: PR description is template placeholder, and title says "into CI" for what's actually a manual-trigger pipeline.

PR #1955, opened 2026-03-06
[Part 1] Clean up service mapping of exception and command definitions

alzimmermsft:
## What does this PR do?

General PR that is doing the following:

**Clean up exception remapping in `*Service.cs` files**

Removes cases where exceptions are caught and mapped to a new exception type. Doing this causes information on what went wrong to be lost. For example, `RequestFailedException`s being mapped lose the ability for `BaseCommand.HandleException` to extract the HTTP status code that caused the request to fail, muddling the ability to troubleshoot.

Spots that were doing this have the remapping replaced with a log message, to retain debugability, and then rethrow the exception as is so that upstream handling can extract appropriate information for logging, telemetry, and how information is returned to the agent.

**Command definition cleanup**

Cleans up miscellaneous things in commands:

- All validation for options must either be done through requiredness of the option or using `Validators` for the command parser.
- Follow development guidance on how to create responses by using terser `new()` syntax.
- Removes JSON property name specifications, as naming pattern of result parameters should serialize to the right format by default.

**Misc cleanup**

- Many replacements of terser `new()` syntax.
- Unused `using` clean ups.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/tools/Azure.Mcp.Tools.Advisor/src/Services/AdvisorService.cs
+++ b/tools/Azure.Mcp.Tools.Advisor/src/Services/AdvisorService.cs
@@ -22,24 +22,14 @@ public class AdvisorService(ISubscriptionService subscriptionService, ITenantSer
         RetryPolicyOptions? retryPolicy,
         CancellationToken cancellationToken = default)
     {
-        try
-        {
-            return await ExecuteResourceQueryAsync(
-                "Microsoft.Advisor/recommendations",
-                resourceGroup,
-                subscription,
-                retryPolicy,
-                ConvertToAdvisorRecommendationModel,
-                tableName: "advisorresources",
-                cancellationToken: cancellationToken);
-        }
-        catch (Exception ex)
-        {
-            _logger.LogError(ex,
-                "Error listing Advisor recommendations. Subscription: {Subscription}, ResourceGroup: {ResourceGroup}",
-                subscription, resourceGroup);
-            throw;
-        }
+        return await ExecuteResourceQueryAsync(
+            "Microsoft.Advisor/recommendations",
+            resourceGroup,
+            subscription,
+            retryPolicy,
+            ConvertToAdvisorRecommendationModel,
+            tableName: "advisorresources",
+            cancellationToken: cancellationToken);
     }
 
     private static Recommendation ConvertToAdvisorRecommendationModel(JsonElement item)
@@ -50,7 +40,6 @@ public class AdvisorService(ISubscriptionService subscriptionService, ITenantSer
         return new(
             ResourceId: advisorRecommendation.Properties?.ResourceMetadata?.ResourceId ?? "Unknown",
             RecommendationText: advisorRecommendation.Properties?.ShortDescription?.Problem ?? "Unknown",
-            Category: advisorRecommendation.Properties?.Category ?? "Unknown"
-        );
+            Category: advisorRecommendation.Properties?.Category ?? "Unknown");
     }
 }
--- a/tools/Azure.Mcp.Tools.AppService/src/Services/AppServiceService.cs
+++ b/tools/Azure.Mcp.Tools.AppService/src/Services/AppServiceService.cs
@@ -54,31 +54,21 @@ public class AppServiceService(
             (nameof(databaseName), databaseName),
             (nameof(subscription), subscription));
 
-        try
-        {
-            // Get Azure resources
-            var webApp = await GetWebAppResourceAsync(subscription, resourceGroup, appName, tenant, retryPolicy, cancellationToken);
+        // Get Azure resources
+        var webApp = await GetWebAppResourceAsync(subscription, resourceGroup, appName, tenant, retryPolicy, cancellationToken);
 
-            // Prepare connection string
-            var finalConnectionString = PrepareConnectionString(connectionString, databaseType, databaseServer, databaseName);
-            var connectionStringName = $"{databaseName}Connection";
+        // Prepare connection string
+        var finalConnectionString = PrepareConnectionString(connectionString, databaseType, databaseServer, databaseName);
+        var connectionStringName = $"{databaseName}Connection";
 
-            // Update web app configuration
-            await UpdateWebAppConnectionStringAsync(webApp, connectionStringName, finalConnectionString, databaseType, cancellationToken);
+        // Update web app configuration
+        await UpdateWebAppConnectionStringAsync(webApp, connectionStringName, finalConnectionString, databaseType, cancellationToken);
 
-            _logger.LogInformation(
-                "Successfully added database connection {ConnectionName} to App Service {AppName}",
-                connectionStringName, appName);
+        _logger.LogInformation(
+            "Successfully added database connection {ConnectionName} to App Service {AppName}",
+            connectionStringName, appName);
 
-            return CreateDatabaseConnectionInfo(databaseType, databaseServer, databaseName, finalConnectionString, connectionStringName);
-        }
-        catch (Exception ex)
-        {
-            _logger.LogError(ex,
-                "Failed to add database connection to App Service {AppName} in resource group {ResourceGroup}",
-                appName, resourceGroup);
-            throw;
-        }
+        return CreateDatabaseConnectionInfo(databaseType, databaseServer, databaseName, finalConnectionString, connectionStringName);
     }
 
     private async Task<WebSiteResource> GetWebAppResourceAsync(string subscription, string resourceGroup,
--- a/tools/Azure.Mcp.Tools.ApplicationInsights/src/Services/ApplicationInsightsService.cs
+++ b/tools/Azure.Mcp.Tools.ApplicationInsights/src/Services/ApplicationInsightsService.cs
@@ -46,22 +46,14 @@ public class ApplicationInsightsService(
     {
         ValidateRequiredParameters((nameof(subscription), subscription));
 
-        try
-        {
-            List<JsonNode> results = [];
-            var components = await GetApplicationInsightsComponentsAsync(subscription, resourceGroup, tenant, retryPolicy, cancellationToken).ConfigureAwait(false);
+        List<JsonNode> results = [];
+        var components = await GetApplicationInsightsComponentsAsync(subscription, resourceGroup, tenant, retryPolicy, cancellationToken).ConfigureAwait(false);
 
-            var insights = await _profilerDataClient.GetInsightsAsync(resourceIds: components.Select(c => c.Id), cancellationToken: cancellationToken).ConfigureAwait(false);
-            results.AddRange(insights);
+        var insights = await _profilerDataClient.GetInsightsAsync(resourceIds: components.Select(c => c.Id), cancellationToken: cancellationToken).ConfigureAwait(false);
+        results.AddRange(insights);
 
-            // Return all results for this resource group (outer method enforces global max)
-            return results;
-        }
-        catch (Exception ex)
-        {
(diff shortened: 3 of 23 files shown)


tools/Azure.Mcp.Tools.ApplicationInsights/tests/Azure.Mcp.Tools.ApplicationInsights.UnitTests/RecommendationListCommandTests.cs
```diff
@@ -63,7 +63,7 @@ public async Task ExecuteAsync_WhenServiceReturnsNoInsights_NoResults()
             Arg.Any<string?>(),
             Arg.Any<RetryPolicyOptions?>(),
             Arg.Any<CancellationToken>())
-            .Returns(Task.FromResult<IEnumerable<JsonNode>>(Array.Empty<JsonNode>()));
+            .Returns(Task.FromResult<IEnumerable<JsonNode>>([]));
```
conniey: The value of Array.Empty is that it returns it a single cached instance whereas [] will allocate something. If you know  your collection is never going to contain things its preferable to use Collection.Empty or Array.Empty.

tools/Azure.Mcp.Tools.Compute/src/Commands/Disk/DiskGetCommand.cs
```diff
@@ -64,7 +60,6 @@ protected override DiskGetOptions BindOptions(ParseResult parseResult)
     {
         var options = base.BindOptions(parseResult);
         options.Disk = parseResult.GetValueOrDefault<string>(ComputeOptionDefinitions.Disk.Name);
-        options.ResourceGroup ??= parseResult.GetValueOrDefault<string>(OptionDefinitions.Common.ResourceGroup.Name);
```
conniey, line 67: Is this not required?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmCreateCommand.cs
```diff
@@ -18,7 +16,7 @@
 namespace Azure.Mcp.Tools.Compute.Commands.Vm;
 
 public sealed class VmCreateCommand(ILogger<VmCreateCommand> logger)
-    : BaseComputeCommand<VmCreateOptions>()
+    : BaseComputeCommand<VmCreateOptions>(true)
```
conniey, line 19: Does this change anything or was the default value true and we were passing it now?

tools/Azure.Mcp.Tools.Compute/src/Commands/Disk/DiskGetCommand.cs
alzimmermsft: It is, I made a change to the command that this extends to take a `bool` in the constructor to determine whether a sub-type command requires resource group to centralize the handling.

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmCreateCommand.cs
alzimmermsft: This is a change to centralize whether resource group is marked as required or optional in the base command's binding of options, so each sub-type command doesn't need to check for whether resource group was passed when required.

tools/Azure.Mcp.Tools.ApplicationInsights/src/Services/ApplicationInsightsService.cs
```diff
@@ -46,19 +45,19 @@ private async Task<IEnumerable<JsonNode>> GetProfilerInsightsImpAsync(
...
 
-            IEnumerable<JsonNode> insights = await _profilerDataClient.GetInsightsAsync(resourceIds: components.Select(c => c.Id), cancellationToken: cancellationToken).ConfigureAwait(false);
+            var insights = await _profilerDataClient.GetInsightsAsync(resourceIds: components.Select(c => c.Id), cancellationToken: cancellationToken).ConfigureAwait(false);
             results.AddRange(insights);
 
             // Return all results for this resource group (outer method enforces global max)
             return results;
         }
-        catch (Exception ex) when (ex is not ArgumentNullException)
+        catch (Exception ex)
```
conniey: This changes the original behaviour because when an ArgumentNullException is thrown, it'll also fall into this catch.

tools/Azure.Mcp.Tools.ApplicationInsights/tests/Azure.Mcp.Tools.ApplicationInsights.UnitTests/RecommendationListCommandTests.cs
conniey: Would prefer this was reverted because it is a good use-case for using Array.Empty<>

tools/Azure.Mcp.Tools.AppService/src/Services/AppServiceService.cs
```diff
@@ -45,17 +45,17 @@ public async Task<DatabaseConnectionInfo> AddDatabaseAsync(
             "Adding database connection to App Service {AppName} in resource group {ResourceGroup}",
             appName, resourceGroup);
 
+        // Validate inputs
+        ValidateRequiredParameters(
```
conniey, line 49: Moving this out of the try/catch would change the original behaviour if this throws an exception.

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmUpdateCommand.cs
```diff
@@ -3,12 +3,10 @@
```
conniey, line 3: @KarishmaGhiya iirc, you helped with the compute MCP tools.  Can you look over these Azure.Mcp.Tool.Compute changes? There's been some logic changes and my limited knowledge may miss something.

tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
```diff
@@ -112,19 +112,8 @@ private string GetCosmosBaseUri(string accountName)
...
-        }
-        catch (CosmosException ex)
-        {
-            throw new Exception($"Failed to validate CosmosClient: {ex.StatusCode} - {ex.Message}", ex);
-        }
-        catch (Exception ex)
-        {
-            throw new Exception($"Unexpected error while validating CosmosClient: {ex.Message}", ex);
-        }
+        // Perform a lightweight operation to validate the client
```
conniey, line 109: By default, we defer to service side validation.  What if some service validation changes and an exception is thrown.  This would change that flow.

tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
```diff
@@ -221,34 +203,27 @@ public async Task<List<string>> ListDatabases(
         var client = await GetCosmosClientAsync(accountName, subscription, authMethod, tenant, retryPolicy, cancellationToken);
         var databases = new List<string>();
 
-        try
+        var iterator = client.GetDatabaseQueryStreamIterator();
```
conniey, line 200: What was the reason for removing this try/catch? Probably because I am only seeing a slice of the changes, I am missing something.

tools/Azure.Mcp.Tools.Deploy/src/Commands/Plan/GetCommand.cs
```diff
@@ -73,11 +73,8 @@ public override Task<CommandResponse> ExecuteAsync(CommandContext context, Parse
 
         try
         {
-            using (SHA256 sha256Hash = SHA256.Create())
-            {
-                byte[] bytes = sha256Hash.ComputeHash(Encoding.UTF8.GetBytes(options.ProjectName));
-                context.Activity?.AddTag(DeployTelemetryTags.ProjectName, BitConverter.ToString(bytes).Replace("-", "").ToLowerInvariant());
-            }
+            byte[] bytes = SHA256.HashData(Encoding.UTF8.GetBytes(options.ProjectName));
```
conniey: To align with your other changes, I think this would be `var`

tools/Azure.Mcp.Tools.ApplicationInsights/src/Services/ApplicationInsightsService.cs
alzimmermsft: Based on the flow of code, it is the same behavior, just adds in an error log. Nothing was checking for `ArgumentNullException` externally.

tools/Azure.Mcp.Tools.AppService/src/Services/AppServiceService.cs
alzimmermsft: To me, this should be an acceptable behavior change as all that happened before is we would catch the exception and rethrow it.

Thinking on this PR more, it might make more sense to just remove these try/catch blocks that just log and rethrow entirely as they are just caught again in the calling `*Command` tool.

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmUpdateCommand.cs
alzimmermsft: Logic change here is using the expected way to validator inputs to a tool call command. We shouldn't be validating in the `ExecuteAsync` method, we should validate in `Command.Validators`. That appropriately handles much more and will perform extensive validation rather than a fast failure.

tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
alzimmermsft: Would it change the overall behavior though? This was in a try/catch block that just rethrew a new wrapping exception.

tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
alzimmermsft: All it did was catch a thrown exception and wrapped it with a message that isn't really necessary as when the tool call failed with exception context knows that it failed on listing Cosmos databases (as that is the tool being used). The exception wrapping then muddles upstream logic that does some inspections on the exception type to return specific information, such as HTTP status codes.

tools/Azure.Mcp.Tools.Advisor/src/Services/AdvisorService.cs
```diff
@@ -21,38 +22,24 @@ public async Task<ResourceQueryResults<Recommendation>> ListRecommendationsAsync
...
     }
 
     private static Recommendation ConvertToAdvisorRecommendationModel(JsonElement item)
     {
-        Models.RecommendationData? advisorRecommendation = Models.RecommendationData.FromJson(item);
-        if (advisorRecommendation == null)
-        {
-            throw new InvalidOperationException("Failed to parse Advisor recommendation data");
-        }
+        Models.RecommendationData? advisorRecommendation = Models.RecommendationData.FromJson(item)
```
conniey, line 37: Is it possible for this to be null if we throw an InvalidOperationException?

```suggestion
        Models.RecommendationData advisorRecommendation = Models.RecommendationData.FromJson(item)
```

alzimmermsft: Given the size of this PR, and that this code change didn't actually modify behaviors, I'll look into this in a follow-up PR.

PR #1967, opened 2026-03-09
add foundry prefix to foundry tools.

g2vinay:
Adds foundry_ prefix to the tools loaded from external foundry mcp server.


JasonYeMSFT:
Do we need to add the prefix so that the tools don't have conflicts or ambiguities when running Azure MCP in non-namespace mode?

g2vinay:
> Do we need to add the prefix so that the tools don't have conflicts or ambiguities when running Azure MCP in non-namespace mode?

yes, and makes it easy for users to spot, as they are then grouped together in all tools mode.

PR #1978, opened 2026-03-10
Integrate Config based env retrieval

g2vinay:


MontaEllis8 (approved):
✅ Code Review 通过

代码审查通过，建议合并。

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -94,6 +94,20 @@ internal class CustomChainedCredential(string? tenantId = null, ILogger<CustomCh
     /// </summary>
     internal static IAzureCloudConfiguration? CloudConfiguration { get; set; }
 
+    /// <summary>
+    /// Configuration for reading settings. Set by DI container during service registration.
+    /// When set, configuration-based retrieval is preferred over direct environment variable access,
+    /// allowing values to come from non-environment sources (e.g. appsettings.json) in addition to environment variables.
+    /// </summary>
+    internal static IConfiguration? Configuration { get; set; }
```
anuchandy, line 102: I wonder if `IConfiguration` indirection layer  we've here earn its keep? -

1. "AZURE_CLIENT_ID" is a standard Azure Identity convention. The `EnvironmentCredential` in the same chain reads it directly from [env](https://github.com/Azure/azure-sdk-for-net/blob/0931022de5b78b7f0322d626841ce8fa5c8565e9/sdk/identity/Azure.Identity/src/EnvironmentVariables.cs#L16) via `EnvironmentCredentialOptions` - routing our read through `IConfiguration` while the azure-identity's own read bypasses it creates inconsistency within the same credential chain.
2. "VSCODE_PID" is process introspection set by VS Code it will never appear in `appsettings.json`.

These values are lazily initialized and cached, so env reads happen once per instance - IConfiguration indirection doesn't improve caching or correctness.

My two cents, I'd lean toward keeping `Environment.GetEnvironmentVariable`  as-is. If there's a specific scenario where appsettings.json sourcing is needed for certain values later, we could scope the change to just that variable rather than adding a blanket IConfiguration layer.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -462,14 +476,11 @@ private IHost CreateHttpHost(ServiceStartOptions serverOptions)
...
+        // Read once at host setup time — this value is process-wide and effectively static,
         // so there is no need to re-read it on every incoming request.
-        // Default to false; the env var must be present and parse to "true" to enable.
-        bool enableForwardedHeaders =
-            bool.TryParse(
-                Environment.GetEnvironmentVariable("AZURE_MCP_DANGEROUSLY_ENABLE_FORWARDED_HEADERS"),
-                out bool parsedEnvVar)
-            && parsedEnvVar;
+        // Default to false; the configuration value must parse to "true" to enable.
+        string? forwardedHeadersSetting = builder.Configuration["AZURE_MCP_DANGEROUSLY_ENABLE_FORWARDED_HEADERS"];
```
anuchandy, line 482: So we're now reading `ALLOW_INSECURE_EXTERNAL_BINDING`, `AZURE_MCP_DANGEROUSLY_DISABLE_HTTPS_REDIRECTION`, and `AZURE_MCP_DANGEROUSLY_ENABLE_FORWARDED_HEADERS` via WebApplicationBuilder - meaning they can now be set via "appsettings.json". These are security-sensitive opt-ins, so this is worth scrutinizing: a checked-in config file could accidentally enable insecure settings, whereas env vars require explicit runtime intent. Would be good to get a read from the security folks before we settle on this.

Tagging Xiang, Srikanta - \\cc @xiangyan99 , @srnagar

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -964,11 +982,20 @@ private static WebApplication UseHttpsRedirectionIfEnabled(WebApplication app)
...
+            isTelemetryEnabled = true;
+        }
+        else if (!bool.TryParse(telemetrySetting, out isTelemetryEnabled))
+        {
+            // Treat invalid/unknown values as telemetry disabled to avoid startup failures.
+            isTelemetryEnabled = false;
+        }
 
-        string? connectionString = Environment.GetEnvironmentVariable("APPLICATIONINSIGHTS_CONNECTION_STRING");
+        string? connectionString = configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"];
```
anuchandy, line 998: Routing through `IConfiguration` appears to unlock setting `APPLICATIONINSIGHTS_CONNECTION_STRING` via appsettings.json, but this flat key is an env var convention.

anuchandy (commented):
Thanks Vinay! @g2vinay, these changes route env var reads through IConfiguration, but most of the values here are env var conventions by design - so wondering if this routing is necessary. Since it shifts expectations around how these are configured, it might be worth holding off on including this in GA.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: I assume, @srnagar's direction was to make it similar to how we support Configuration in Azure Core in Java Azure SDKs. 
Read from both code level config, properties, env var etc. 
If that's still the goal, then the code level can override the env var, which is fine, as it won't regress users.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -94,6 +94,20 @@ internal class CustomChainedCredential(string? tenantId = null, ILogger<CustomCh
...
+
+    /// <summary>
+    /// Reads a configuration value, preferring <see cref="Configuration"/> when available
+    /// and falling back to <see cref="Environment.GetEnvironmentVariable"/> otherwise.
+    /// </summary>
+    private static string? GetConfigValue(string key) =>
+        Configuration?[key] ?? Environment.GetEnvironmentVariable(key);
```
alzimmermsft, line 109: I'd make this check a bit more verbose as there's a different between `Configuration` is null and fallback to `Environment.GetEnvironmentVariable` and `Configuration` returned null from the lookup, as `Configuration` will fallback to looking at `Environment.GetEnvironmentVariable`.


```suggestion
    private static string? GetConfigValue(string key) =>
        (Configuration == null)
            ? Environment.GetEnvironmentVariable(key)
            : Configuration[key];
```

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: We can add this selectively too, by design we can exclude certain areas where we want to restrict to env var only. 
Will bring downside of incosnsitency, but if its justified by design, its fine, for Credential purposes, I don't see strong reason to restrict it, its fine to support config there too.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
g2vinay: We can opt to selectively restrict it to env here by design.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
anuchandy: Yeah, I was trying to understand the general pattern libs use. 

Taking the telemetry SDK as an example: `APPLICATIONINSIGHTS_CONNECTION_STRING` (ALL_CAPS) is a flat "env var" read directly via `Environment.GetEnvironmentVariable`, not through `IConfiguration`. 

The appsettings.json equivalent is the hierarchical `ApplicationInsights:ConnectionString` section, which maps to `ApplicationInsights__ConnectionString` (double underscore) "env-var"  for IConfiguration resolution.

So my main point was that reading ALL_CAPS "env vars" through IConfiguration is uncommon - users understand IConfiguration work with the double-underscore "env var" form of a hierarchical section. That said, I don't see a concrete issue with us supporting ALL_CAPS via appsettings.json - yet.

g2vinay: I see, then perhaps we can also honor the double-underscore format as best practice.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -849,7 +859,13 @@ private static void InitializeListingUrls(WebApplicationBuilder builder, Service
             throw new InvalidOperationException($"Explicit external binding is not supported for '{url}'.");
         }
 
-        if (isWildcard && !EnvironmentHelpers.GetEnvironmentVariableAsBool("ALLOW_INSECURE_EXTERNAL_BINDING"))
+        var allowInsecureExternalBindingRaw = builder.Configuration["ALLOW_INSECURE_EXTERNAL_BINDING"];
+        bool allowInsecureExternalBinding = false;
+        if (!string.IsNullOrWhiteSpace(allowInsecureExternalBindingRaw))
+        {
+            _ = bool.TryParse(allowInsecureExternalBindingRaw, out allowInsecureExternalBinding);
```
jongio, line 866: **[HIGH]** Logic: `bool.TryParse` doesn't accept "T" or "1" - breaking change

The original `EnvironmentHelpers.GetEnvironmentVariableAsBool` accepted "true", "True", "T", and "1". `bool.TryParse` only accepts "true"/"false" (case-insensitive). Deployments using `ALLOW_INSECURE_EXTERNAL_BINDING=1` or `=T` will now get an `InvalidOperationException` on startup with no indication the accepted values changed.

This is the only boolean in ServiceStartCommand.cs that previously used the permissive helper - the other booleans already used `bool.TryParse` in the original code, so they don't regress.

Consider reusing the same truthy-value set as `ShouldUseOnlyBrokerCredential` via a shared helper.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -120,13 +134,14 @@ public override ValueTask<AccessToken> GetTokenAsync(TokenRequestContext request
 
     private static bool ShouldUseOnlyBrokerCredential()
     {
-        return EnvironmentHelpers.GetEnvironmentVariableAsBool(OnlyUseBrokerCredentialEnvVarName);
+        string? value = GetConfigValue(OnlyUseBrokerCredentialEnvVarName);
+        return value is "true" or "True" or "T" or "1";
```
jongio, line 138: **[MEDIUM]** Logic: Boolean parsing inconsistent within this PR

This method preserves the original truthy set ("true", "True", "T", "1") from `EnvironmentHelpers.GetEnvironmentVariableAsBool`. But `ALLOW_INSECURE_EXTERNAL_BINDING` in ServiceStartCommand.cs switched to `bool.TryParse`, which rejects "T" and "1".

The PR now has two different boolean parsing contracts for config values. Consider extracting a shared `ParseBoolConfig(string? value)` helper that all boolean config reads use.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -156,7 +171,7 @@ private static TokenCredential CreateCredential(string? tenantId, ILogger<Custom
         var creds = new List<TokenCredential>();
 
         // Check if we are running in a VS Code context. VSCODE_PID is set by VS Code when launching processes, and is a reliable indicator for VS Code-hosted processes.
-        bool isVsCodeContext = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSCODE_PID"));
+        bool isVsCodeContext = !string.IsNullOrEmpty(GetConfigValue("VSCODE_PID"));
```
jongio, line 174: **[MEDIUM]** Security: VSCODE_PID is a runtime process signal, not a user-configurable setting

VSCODE_PID is injected by VS Code into child process environments. Reading it via `GetConfigValue` (which includes appsettings.json) means a checked-in config file could fake a VS Code context, changing the credential chain priority. This isn't a setting users should be able to override through configuration files.

Consider keeping this as `Environment.GetEnvironmentVariable("VSCODE_PID")` - it's a process detection signal, not a tunable parameter.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -198,7 +198,21 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
...
-            using var tracerProvider = AddIncomingAndOutgoingHttpSpans(options);
+            // Build configuration using the same sources the host will use so AddIncomingAndOutgoingHttpSpans
+            // has access to IConfiguration before the host DI container is constructed.
+            // This mirrors Host.CreateDefaultBuilder / WebApplication.CreateBuilder source order:
+            // appsettings.json → appsettings.{env}.json → environment variables.
+            string hostEnvironment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
+                ?? Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT")
+                ?? "Production";
+            IConfiguration preHostConfiguration = new ConfigurationBuilder()
+                .SetBasePath(AppContext.BaseDirectory)
```
jongio, line 209: **[MEDIUM]** Architecture: preHostConfiguration doesn't mirror the full host configuration

This builder includes appsettings.json + env vars, but misses user-secrets (in Development), command-line args, and any custom providers that `WebApplication.CreateBuilder()` adds. If someone sets `AZURE_MCP_COLLECT_TELEMETRY` via user-secrets or CLI args, this config won't see it while the actual host config would.

Two options: (1) defer telemetry init until after the host is built, or (2) document that pre-host telemetry config only reads JSON files and env vars.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
```diff
@@ -400,6 +401,66 @@ public void PinnedCredentialMode_DoesNotAddDeviceCodeFallback_CreatesCredentialS
         Assert.IsAssignableFrom<TokenCredential>(credential);
     }
 
+    /// <summary>
+    /// Tests that when Configuration is set, token credentials are read from IConfiguration
```
jongio, line 405: **[MEDIUM]** Tests: Coverage gaps for the new config path

The two new tests only verify AZURE_TOKEN_CREDENTIALS through IConfiguration. Missing coverage:
- ShouldUseOnlyBrokerCredential via config (the truthy-value pattern is untested through the IConfiguration path)
- Value precedence when both IConfiguration and env var provide different values
- Edge cases: "TRUE" (all-caps) returns false from the pattern match, which may surprise users
- Empty-string config values (IConfiguration returns "" not null, which blocks the env var fallback in GetConfigValue)

jongio (changes requested):
Issues to address:
- ServiceStartCommand.cs:866 - ALLOW_INSECURE_EXTERNAL_BINDING drops "T"/"1" as accepted truthy values (breaking change from EnvironmentHelpers)
- CustomChainedCredential.cs:138 - boolean parsing inconsistent (ShouldUseOnlyBrokerCredential keeps "T"/"1", ALLOW_INSECURE_EXTERNAL_BINDING doesn't)
- CustomChainedCredential.cs:174 - VSCODE_PID is a process detection signal, shouldn't be config-file-settable
- ServiceStartCommand.cs:209 - preHostConfiguration misses user-secrets, CLI args that the real host config includes
- CustomChainedCredentialTests.cs - test gaps for broker credential config path, value precedence, edge cases

Also worth noting: the static mutable `Configuration` property follows the existing `CloudConfiguration` pattern, but expands the surface area for cross-container interference in multi-tenant/OBO scenarios. The telemetry comment at line 994 says "to avoid startup failures" but the old code also disabled telemetry for invalid values - the comment is misleading about what changed.

PR #2006, opened 2026-03-10
Enable OTLP exporting as part of development build

alzimmermsft:
## What does this PR do?

Enables OTLP exporters by default for development builds to allow for richer debugging without needing to configure the environment. Replaces `"AZURE_MCP_COLLECT_TELEMETRY": "false"` with `"AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT": "false"` which controls more specific behaviors, which we want, to allow custom Application Insights to be enabled without additional configuration in development builds.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:
    - [ ] For tools with new names, including new tools or renamed tools, update [`consolidated-tools.json`](https://github.com/microsoft/mcp/blob/main/core/Microsoft.Mcp.Core/src/Areas/Server/Resources/consolidated-tools.json)


saikoumudi (approved):

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -248,9 +248,13 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
     /// Using <see cref="IConfiguration"/> configures <see cref="McpServerConfiguration"/>.
     /// </summary>
     /// <param name="services">Service Collection to add configuration logic to.</param>
-    public static void InitializeConfigurationAndOptions(this IServiceCollection services)
+    public static void InitializeConfigurationOptionsAndOpenTelemetry(this IServiceCollection services)
     {
+#if DEBUG
```
conniey, line 256: The reason for having DOTNET_ENVIRONMENT is so we don't have to recompile to make test changes. We can rerun the application with different settings.  We shouldn't have to if/def things based on the build type.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -248,9 +248,13 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
     /// Using <see cref="IConfiguration"/> configures <see cref="McpServerConfiguration"/>.
     /// </summary>
     /// <param name="services">Service Collection to add configuration logic to.</param>
-    public static void InitializeConfigurationAndOptions(this IServiceCollection services)
+    public static void InitializeConfigurationOptionsAndOpenTelemetry(this IServiceCollection services)
```
conniey, line 254: Is there a reason we have to configure open telemetry in this method as well? This method now has multiple responsibilities rather than just setting options.

conniey (changes requested):

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -92,7 +93,7 @@ private static void EnableAzureMonitor(this IServiceCollection services)
 #if RELEASE
```
conniey, line 93: Frankly, I think we could have avoided this by using default appsettings.json = false and used the running environment to set it or not. Because now if we want to test this in a debug environment, it's impossible.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
conniey: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-10.0

alzimmermsft: I was finding the DOTNET_ENVIRONMENT was blank when running debug builds, so we fell back to `Production`. If you know of a better way to fallback to `Development` when running a debug build do let me know.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
alzimmermsft: I couldn't find a cleaner way to get the IConfiguration singleton bound here when we called `OpenTelemetryExtensions.ConfigureOpenTelemetry`. If you know of a way to cleanly pull that out of the `IServiceCollection` I'd be more than happy to revert this change as I'm not a huge fan of it either.

servers/Azure.Mcp.Server/src/appsettings.Development.json
```diff
@@ -1,8 +1,8 @@
 {
-    // Generally, AZURE_MCP_COLLECT_TELEMETRY will be set as an environment variable.
-    // For Development purposes, we never want to publish telemetry so it is explicitly
-    // set here.
-    "AZURE_MCP_COLLECT_TELEMETRY": "false",
+    // Disable AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT to prevent development telemetry from being collected by Microsoft
+    // instrumentation. This is to avoid polluting production telemetry with development data.
+    "AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT": "false",
```
jongio, line 4: **[HIGH]** Inconsistency: Fabric and Template dev configs not updated

Both `servers/Fabric.Mcp.Server/src/appsettings.Development.json` and `servers/Template.Mcp.Server/src/appsettings.Development.json` still have the old `AZURE_MCP_COLLECT_TELEMETRY: false` and lack the new `AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT` and `AZURE_MCP_ENABLE_OTLP_EXPORTER` keys. Their Program.cs files were updated to call the renamed method, but their dev configs are stale. Devs running Fabric or Template servers in Development will get different telemetry behavior than Azure.Mcp.Server.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
jongio, line 254: **[MEDIUM]** Breaking change: Public API renamed without backward-compat shim

`InitializeConfigurationAndOptions` is a public extension method. Renaming it to `InitializeConfigurationOptionsAndOpenTelemetry` (plus changing `ConfigureOpenTelemetry`'s signature to require `IConfiguration`) is a source and binary break for any external consumer of `Microsoft.Mcp.Core`. Consider adding `[Obsolete]` forwarders that delegate to the new methods for one release, or confirm no external callers exist.

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -80,7 +81,7 @@ private static void EnableAzureMonitor(this IServiceCollection services)
                     .AddTelemetrySdk();
             });
 
-        var userProvidedAppInsightsConnectionString = Environment.GetEnvironmentVariable("APPLICATIONINSIGHTS_CONNECTION_STRING");
+        var userProvidedAppInsightsConnectionString = configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"];
```
jongio, line 84: **[MEDIUM]** Config: Connection string now readable from checked-in JSON

Switching from `Environment.GetEnvironmentVariable` to `configuration[...]` means `APPLICATIONINSIGHTS_CONNECTION_STRING` can now come from `appsettings.json` or `appsettings.Development.json` - not just env vars. If someone adds a connection string to a checked-in config file, it becomes a committed secret. The env-var-only approach was a natural guard against this. Consider keeping this particular value as env-var-only, or document that connection strings should only be set via env vars / user-secrets.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/OpenTelemetryExtensionsTests.cs
```diff
@@ -24,7 +26,7 @@ public void ConfigureOpenTelemetry_RegistersTelemetryService()
         var services = new ServiceCollection();
 
         // Act
-        services.ConfigureOpenTelemetry();
+        services.ConfigureOpenTelemetry(Substitute.For<IConfiguration>());
```
jongio, line 29: **[MEDIUM]** Tests: Mock config doesn't verify new behavior

`Substitute.For<IConfiguration>()` returns null for all reads, so this test only proves the method doesn't throw with empty config. No test verifies that setting `AZURE_MCP_ENABLE_OTLP_EXPORTER` to `true` actually registers OTLP exporters, or that `APPLICATIONINSIGHTS_CONNECTION_STRING` from config enables the user-provided exporter.

jongio (changes requested):
Enables OTLP dev export and migrates telemetry config reads to IConfiguration.

Issues to address:
- appsettings.Development.json - Fabric and Template servers still have the old config keys; need the same update
- OpenTelemetryExtensions.cs:84 - APPLICATIONINSIGHTS_CONNECTION_STRING now readable from checked-in JSON (was env-only)
- ServiceCollectionExtensions.cs:251 - public API renamed without obsolete forwarders for external consumers
- OpenTelemetryExtensionsTests.cs:29 - mock config doesn't exercise the new IConfiguration code paths

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -25,7 +26,7 @@ public static class OpenTelemetryExtensions
     /// </summary>
     private const string MicrosoftOwnedAppInsightsConnectionString = "InstrumentationKey=21e003c0-efee-4d3f-8a98-1868515aa2c9;IngestionEndpoint=https://centralus-2.in.applicationinsights.azure.com/;LiveEndpoint=https://centralus.livediagnostics.monitor.azure.com/;ApplicationId=f14f6a2d-6405-4f88-bd58-056f25fe274f";
 
-    public static void ConfigureOpenTelemetry(this IServiceCollection services)
+    public static void ConfigureOpenTelemetry(this IServiceCollection services, IConfiguration configuration)
```
jongio, line 29: **[MEDIUM]** `ConfigureOpenTelemetry` signature is also a breaking change

Adding a required `IConfiguration` parameter changes this public extension method's signature. External consumers calling `services.ConfigureOpenTelemetry()` won't compile. Same concern as the `InitializeConfigurationAndOptions` rename - needs either an `[Obsolete]` parameterless overload or a default parameter value.

jongio (commented):
One more item missed in my earlier review:
- OpenTelemetryExtensions.cs:29 - `ConfigureOpenTelemetry` signature also changed (added required `IConfiguration` param) - separate public API break from the rename
- OpenTelemetryExtensionsTests.cs:18 - doc comment still references old method name `InitializeConfigurationAndOptions`

jongio (dismissed):
The merge commit (520001) resolves conflicts with main but doesn't address the issues from my March/April reviews.

**Outstanding items:**

1. **Sibling config gap**: Fabric and Template appsettings.Development.json still use the old AZURE_MCP_COLLECT_TELEMETRY key, but both servers' Program.cs files now call InitializeConfigurationOptionsAndOpenTelemetry, which reads AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT from config. Dev builds of Fabric and Template won't properly disable Microsoft telemetry.

2. **Two public API breaks without forwarders**: InitializeConfigurationAndOptions was renamed and ConfigureOpenTelemetry gained a required IConfiguration param. No [Obsolete] shims for external consumers.

3. **Config-sourced connection string**: APPLICATIONINSIGHTS_CONNECTION_STRING now reads from IConfiguration (includes JSON files), where it was previously env-only.

4. **Test coverage**: Substitute.For<IConfiguration>() returns null for all reads, so the new config-based code paths (OTLP enable, App Insights connection string, Microsoft telemetry opt-out) aren't exercised.

5. **#if DEBUG for DOTNET_ENVIRONMENT default**: @conniey's concern about compile-time env detection vs. runtime configuration is also still open.

servers/Azure.Mcp.Server/src/appsettings.Development.json
jongio, line 4: Fabric and Template servers still have `AZURE_MCP_COLLECT_TELEMETRY`: `false` in their dev configs. Since the code now reads `AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT` from `IConfiguration`, dev builds of those servers won't opt out of Microsoft telemetry. They also won't get OTLP export enabled. Both sibling configs need the same update applied here.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -251,9 +251,13 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
...
     {
+#if DEBUG
```
jongio, line 256: Using `#if DEBUG` to choose the environment default means you can't test production config behavior without recompiling. This is a step backward from runtime configurability. Consider setting `DOTNET_ENVIRONMENT=Development` in launchSettings.json (standard .NET practice) and always defaulting to `Production` in code. That way the runtime environment is configurable without a rebuild.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -251,9 +251,13 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
     /// Using <see cref="IConfiguration"/> configures <see cref="McpServerConfiguration"/>.
     /// </summary>
     /// <param name="services">Service Collection to add configuration logic to.</param>
-    public static void InitializeConfigurationAndOptions(this IServiceCollection services)
+    public static void InitializeConfigurationOptionsAndOpenTelemetry(this IServiceCollection services)
```
jongio, line 254: Public API rename without a backward-compat shim. `InitializeConfigurationAndOptions` is a public extension method. Adding an `[Obsolete]` forwarder that calls the new name would avoid breaking external consumers.

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
jongio, line 29: Adding a required `IConfiguration` parameter here is a second public API break (separate from the rename). External consumers calling `services.ConfigureOpenTelemetry()` will fail at compile time. Consider an overload or a default parameter (`IConfiguration? configuration = null`) with a fallback to `Environment.GetEnvironmentVariable` when null.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/OpenTelemetryExtensionsTests.cs
jongio, line 29: `Substitute.For<IConfiguration>()` returns null for all indexer reads, so this test only proves the method doesn't throw with empty config. It doesn't verify that a non-null `APPLICATIONINSIGHTS_CONNECTION_STRING` or `AZURE_MCP_ENABLE_OTLP_EXPORTER` value actually wires up the expected exporters. Consider setting up the mock to return test values and asserting on the resulting service registrations.

jongio (commented):
Re-reviewing after my earlier review was dismissed. The merge commit (520001) resolved conflicts with main but the substantive issues remain. The PR also has merge conflicts with current main.

**Key items still outstanding:**

1. Sibling config gap: Fabric and Template `appsettings.Development.json` still reference the old `AZURE_MCP_COLLECT_TELEMETRY` key. Both servers now call the renamed method that reads `AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT` from config, so dev builds of those servers won't properly opt out of Microsoft telemetry.

2. Compile-time env detection: The `#if DEBUG` block defaults `DOTNET_ENVIRONMENT` to "Development" at compile time, removing the ability to test production config paths without recompiling. conniey's concern about this is still open.

3. Two public API breaks without `[Obsolete]` forwarders: `InitializeConfigurationAndOptions` was renamed and `ConfigureOpenTelemetry` gained a required `IConfiguration` parameter. External consumers calling either method will fail at compile time after upgrading.

4. Test coverage: `Substitute.For<IConfiguration>()` returns null for all indexer reads, so the new config-based code paths (OTLP enable, App Insights connection string, Microsoft telemetry opt-out) aren't exercised by the existing tests.

alzimmermsft:
Closing as this can be done in the future as manually modifying environment configurations locally isn't too much of an ask.

PR #2018, opened 2026-03-11
Reduce AppLens failures

tmeschter:
Related to https://github.com/microsoft/mcp-pr/issues/172.

## Description

The AppLens `diagnose` tool uses an AppLens endpoint to identify problems in Azure resources and provide solutions. It currently has a high failure rate, likely due to how parameters are presented to the LM and handled internally.

Previously, subscription, resource group, and resource type were all required to construct the ARM resource ID passed to the AppLens API, but were presented as if optional. This often caused the LM's first attempt to fail, requiring it to retry with additional information—or worse, hallucinate values leading to further failures.

### Changes

- **Made `--subscription`, `--resource-group`, and `--resource-type` truly optional** — only `--resource` (name) and `--question` are required
- **Added Azure Resource Graph (ARG) discovery** — queries ARG by resource name to find matching resources, then applies progressive filtering by subscription, resource group, and resource type
- **Added supported resource type validation** — whitelist of AppLens-supported types (`microsoft.web/sites`, `microsoft.containerservice/managedclusters`, `microsoft.apimanagement/service`), matching the CFA-AIService implementation
- **Changed command base class** from `SubscriptionCommand` to `GlobalCommand` to support truly optional subscription
- **Changed service base class** from `BaseAzureService` to `BaseAzureResourceService` for ARG query infrastructure
- **Disambiguation messaging** when multiple matching resources are found, guiding the LM to provide more specific parameters

### Files Changed

- `AppLensOptionDefinitions.cs` — Updated option definitions and descriptions
- `ResourceDiagnoseCommand.cs` — Changed to GlobalCommand, all params except resource/question are optional
- `ResourceDiagnoseOptions.cs` — Updated option properties
- `IAppLensService.cs` — Updated interface with nullable subscription
- `AppLensService.cs` — Major rewrite with ARG-based resource discovery
- `ResourceDiagnoseCommandTests.cs` — Updated and expanded tests (30 tests)

## Invoking Livetests

### DIFF
--- a/tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
+++ b/tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
@@ -100,29 +100,8 @@ public class AppLensService(
 
         var filteredResults = queryResults;
 
-        // Progressive filtering by subscription
-        if (!string.IsNullOrEmpty(subscription))
-        {
-            var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenantId, cancellationToken: cancellationToken);
-            var subscriptionId = subscriptionResource.Data.SubscriptionId;
-
-            var subscriptionFiltered = filteredResults
-                .Where(r => r.SubscriptionId.Equals(subscriptionId, StringComparison.OrdinalIgnoreCase))
-                .ToImmutableArray();
-
-            if (subscriptionFiltered.Length == 0)
-            {
-                var foundSubscriptions = string.Join("\n", filteredResults
-                    .Select(r => r.SubscriptionId)
-                    .Distinct(StringComparer.OrdinalIgnoreCase)
-                    .Select(s => $"- {s}"));
-                return new DidNotFindResourceResult(
-                    $"Found resources with name '{resourceName}' but not in subscription '{subscription}'.\n" +
-                    $"Resources were found in the following subscriptions:\n{foundSubscriptions}");
-            }
-
-            filteredResults = subscriptionFiltered;
-        }
+        // There is no need to filter by the given subscription (if any) because that would
+        // have been taken care of by the ARG query.
 
         // Progressive filtering by resource group
         if (!string.IsNullOrEmpty(resourceGroup))


tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
```diff
@@ -72,22 +77,228 @@ public async Task<DiagnosticResult> DiagnoseResourceAsync(
...
+    /// </summary>
+    private async Task<ImmutableArray<AppLensArgQueryResult>> ExecuteArgQueryAsync(
+        string resourceName,
+        string? subscription,
+        string? tenantId,
+        CancellationToken cancellationToken)
     {
-        // Construct a resource ID from the provided information
+        var escapedName = EscapeKqlString(resourceName);
+        var kqlQuery = $"resources | where name =~ '{escapedName}' | project id, subscriptionId, resourceGroup, type, kind, name | limit 50";
```
alzimmermsft, line 199: Any reason we're basically reinventing `BaseAzureResourceService.ExecuteResourceQueryAsync` here? It should be able to handle this by having `name =~ '{escapedName}` passed as the `additionalFilter` parameter to the method.

tools/Azure.Mcp.Tools.AppLens/src/Commands/Resource/ResourceDiagnoseCommand.cs
```diff
@@ -47,19 +48,21 @@ public sealed class ResourceDiagnoseCommand(ILogger<ResourceDiagnoseCommand> log
...
+        command.Options.Add(AppLensOptionDefinitions.Resource);
+        command.Options.Add(AppLensOptionDefinitions.Question);
     }
 
     protected override ResourceDiagnoseOptions BindOptions(ParseResult parseResult)
     {
         var options = base.BindOptions(parseResult);
-        options.ResourceGroup ??= parseResult.GetValueOrDefault<string>(OptionDefinitions.Common.ResourceGroup.Name);
+        options.Subscription = parseResult.GetValueOrDefault<string>(AppLensOptionDefinitions.Subscription.Name);
+        options.ResourceGroup = parseResult.GetValueOrDefault<string>(AppLensOptionDefinitions.ResourceGroup.Name);
```
xiangyan99, line 62: Both Subscription & ResourceGroup are optional, right?

tmeschter: Yes.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
tmeschter: I don't know what type of resource I'm querying for, which is a required parameter for `BaseAzureResourceService.ExecuteResourceQueryAsync`.

tools/Azure.Mcp.Tools.AppLens/src/Commands/Resource/ResourceDiagnoseCommand.cs
xiangyan99: Then should we use ?? for Subscription, ResourceGroup & ResourceType?

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
tmeschter: Also the subscription, which may or may not be provided.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
```diff
@@ -72,22 +77,228 @@ public async Task<DiagnosticResult> DiagnoseResourceAsync(
...
+
+        if (queryResults.Length == 0)
+        {
+            return new DidNotFindResourceResult($"No resources found with name '{resourceName}'.");
+        }
+
+        var filteredResults = queryResults;
+
+        // Progressive filtering by subscription
+        if (!string.IsNullOrEmpty(subscription))
```
xiangyan99: It appears that if the subscription is not null or empty, the information has already been processed in ExecuteArgQueryAsync.

Why is it necessary to use it again to filter the results?

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
```diff
@@ -72,22 +77,228 @@ public async Task<DiagnosticResult> DiagnoseResourceAsync(
...
+        return new FoundResourceResult(resource.Id, resourceTypeAndKind, null);
+    }
+
+    /// <summary>
+    /// Executes an Azure Resource Graph query to find resources by name.
+    /// </summary>
+    private async Task<ImmutableArray<AppLensArgQueryResult>> ExecuteArgQueryAsync(
+        string resourceName,
+        string? subscription,
+        string? tenantId,
```
alzimmermsft, line 195: Should this take `string? resourceGroup = null` to allow the Azure Resource Graph query to filter down results on the passed resource group if we have it.

tools/Azure.Mcp.Tools.AppLens/src/Commands/Resource/ResourceDiagnoseCommand.cs
tmeschter: No, it is fine that `Subscription`, `ResourceGroup`, and `ResourceType` are `null`, since they are optional.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
tmeschter: We want the results of the ARG query to be pretty broad as that allows for better error recovery.

Consider the case where the LM passes in a valid resource name but gets the resource group name wrong. In the current code we can tell the LM that we didn't find a resource of the given name in the specified group, but we _did_ find one in this other group; is that the one it meant? If we filter on the resource group in the ARG query we'll get no results at all.

tools/Azure.Mcp.Tools.AppLens/src/Services/AppLensService.cs
tmeschter: You are correct, and I will remove this block.

In the original implementation of the App Lens support from the GitHub Copilot for Azure VS Code extension the ARG query didn't work in quite the same way so it was still helpful to filter on subscription after.

PR #2033, opened 2026-03-12
Centralize arm token logic

g2vinay:
Centralize ARM token auth logic. Remove duplicated code.

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/BaseAzureServiceTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/BaseAzureServiceTests.cs
@@ -221,7 +221,7 @@ public class BaseAzureServiceTests
             CreateArmClientAsync(tenant, retryPolicy);
 
         public Task<AccessToken> GetArmAccessTokenPublicAsync(CancellationToken cancellationToken) =>
-            GetArmAccessTokenAsync(cancellationToken);
+            GetArmAccessTokenAsync(null, cancellationToken);
 
         public Task<AccessToken> GetArmAccessTokenPublicAsync(string? tenant, CancellationToken cancellationToken) =>
             GetArmAccessTokenAsync(tenant, cancellationToken);


chidozieononiwu (approved):

tools/Azure.Mcp.Tools.ResourceHealth/src/Services/ResourceHealthService.cs
anannya03: The changes look good. I have one small question- this method `ListAvailabilityStatusesAsync` also requests token for the ArmEnvironment Default Scope. In that case, would it make sense to use the `GetArmAccessTokenAsync()` method here as well? I was wondering if there was a specific reason for leaving it as is.

The same goes for `ListServiceHealthEventsAsync` method in Line 102. Just wanted to check if there was any context I might be missing.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -158,6 +158,27 @@ protected async Task<TokenCredential> GetCredential(string? tenant, Cancellation
         }
     }
 
+    /// <summary>
+    /// Gets an ARM access token using the ARM default scope.
+    /// </summary>
+    protected async Task<AccessToken> GetArmAccessTokenAsync(CancellationToken cancellationToken)
```
alzimmermsft: Do we want this convenience method?

I think we should remove it and make it clear that if you have `tenant` as a parameter you should pass it. And for those who don't make them pass `null` explicitly so we can more easily find any possible bugs with mismatch tenant credentials due to missing `tenant`.

tools/Azure.Mcp.Tools.ResourceHealth/src/Services/ResourceHealthService.cs
g2vinay: Yeah those should be covered too. good catch.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
g2vinay: not strictly, removed it.

PR #2052, opened 2026-03-16
fix(fabric): isolate tenant-switch auth path for cross-tenant MFA flows (#1797)

### LINKED ISSUE #1797: [BUG] Fabric MCP not able to log in to different tenant
### Describe the bug

Fabric MCP is not able to login to a different tenant: example: client xx's tenant with 2FA setup. The Fabric MCP triggered authentication page is missing 2nd  authentication where once password is provided, it won't trigger 'Authenticator' for  2 digits number prompting. 

### Expected behavior

the Fabric MCP server should be able to switch tenant login to support different authentication mode smoothly. Without the need to grant read access to Azure Subscription as well. Once Fabric tenant access is granted for a given ID to enable login to Fabric portal, Fabric MCP should be able to connect with the same credentials.

### Actual behavior

Image 

this could be hallucination from github copilot. 

### Reproduction Steps

Prompt used: 

Fabric MCP: Connect to Fabric tenant ID = xxx, with my login email = xxxby selecting the default subscription with name = xxx . I don’t have access to the subscription but can access the tenant.  The available workspace that I can access such as “Enterprise xxx”. I have access to multiple tenants. Make sure you log  me to the right tenant. 

### Environment

VS code - version 

Image

newventurecap:
## What does this PR do?

When a user requests authentication against a different tenant, the broker's sticky default account and any cached `AuthenticationRecord` from the current tenant are reused — bypassing the MFA challenge specific to the target tenant. This surfaces as the Authenticator app prompt never appearing when switching tenants.

This PR fixes the issue by adding `isolateTenantAuth` to `CustomChainedCredential`. When `true`, the credential skips the cached auth record and sets `UseDefaultBrokerAccount = false`, forcing a fresh interactive login scoped to the target tenant. The default credential chain behavior is completely unchanged.

Production mode (`AZURE_TOKEN_CREDENTIALS=prod`) is explicitly respected — the isolation early-return is skipped so non-interactive environments (CI, Managed Identity, Workload Identity) are never presented with a browser prompt.

`TenantAwareCredential` from the original draft has been removed — the fix is now a small, additive change to the existing class. Shared broker credential setup is extracted into `CreateBrokerCredential` and `CreateTimeoutCredential` helpers to eliminate duplication between `CreateBrowserCredential` and `CreateTenantIsolatedCredential`.

## GitHub issue number?

Fixes #1797

## Pre-merge Checklist

- [x] Required for All PRs
    - [x] Read [contribution guidelines](https://github.com/microsoft/mcp/blob/main/CONTRIBUTING.md)
    - [x] Commit history is clean with descriptive messages — one commit per file
    - [x] Updated `servers/Fabric.Mcp.Server/CHANGELOG.md` changelog entry for bug fix
- [x] Extra steps for Community (non-Microsoft team member) PRs:
    - [x] **Security review**: No new network calls, external endpoints, or serialization paths introduced. `CreateTenantIsolatedCredential` uses only existing `Azure.Identity.Broker` primitives (`InteractiveBrowserCredentialBrokerOptions`, `UseDefaultBrokerAccount`, `AuthenticationRecord`) already used throughout this class. The two flag changes (`UseDefaultBrokerAccount = false`, `AuthenticationRecord = null`) are standard Azure Identity options for tenant isolation.
    - [x] Manual tests — triggered `/azp run mcp - pullrequest - live` but Azure Pipelines returned "Commenter does not have sufficient privileges for PR 2052". (This is expected for community contributions). Requesting a maintainer to trigger the live test run.

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
@@ -443,7 +443,7 @@ public class CustomChainedCredentialTests
     /// <summary>
     /// Helper method to create CustomChainedCredential using reflection since it's an internal class.
     /// </summary>
-    private static TokenCredential CreateCustomChainedCredential(bool forceBrowserFallback = false)
+    private static TokenCredential CreateCustomChainedCredential(bool forceBrowserFallback = false, bool isolateTenantAuth = false)
     {
         var assembly = typeof(global::Microsoft.Mcp.Core.Services.Azure.Authentication.IAzureTokenCredentialProvider).Assembly;
         var customChainedCredentialType = assembly.GetType("Microsoft.Mcp.Core.Services.Azure.Authentication.CustomChainedCredential");
@@ -454,18 +454,34 @@ public class CustomChainedCredentialTests
             .FirstOrDefault(c =>
             {
                 var parameters = c.GetParameters();
-                return parameters.Length == 3 &&
+                return parameters.Length == 4 &&
                        parameters[0].ParameterType == typeof(string) &&
                        parameters[1].ParameterType == typeof(ILogger<>).MakeGenericType(customChainedCredentialType) &&
-                       parameters[2].ParameterType == typeof(bool);
+                       parameters[2].ParameterType == typeof(bool) &&
+                       parameters[3].ParameterType == typeof(bool);
             });
 
         Assert.NotNull(constructor);
 
-        var credential = constructor.Invoke([null, null, forceBrowserFallback]) as TokenCredential;
+        var credential = constructor.Invoke([null, null, forceBrowserFallback, isolateTenantAuth]) as TokenCredential;
         Assert.NotNull(credential);
 
         return credential;
+    }
+
+    /// <summary>
+    /// Tests that isolateTenantAuth=true creates a credential successfully for cross-tenant flows.
+    /// Expected: Returns a tenant-isolated credential that bypasses sticky broker account and cached auth records.
+    /// </summary>
+    [Fact]
+    public void IsolateTenantAuth_CreatesCredentialSuccessfully()
+    {
+        // Act
+        var credential = CreateCustomChainedCredential(isolateTenantAuth: true);
+
+        // Assert
+        Assert.NotNull(credential);
+        Assert.IsAssignableFrom<TokenCredential>(credential);
     }
 
     private static Type GetCustomChainedCredentialType()
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
@@ -470,18 +470,45 @@ public class CustomChainedCredentialTests
     }
 
     /// <summary>
-    /// Tests that isolateTenantAuth=true creates a credential successfully for cross-tenant flows.
-    /// Expected: Returns a tenant-isolated credential that bypasses sticky broker account and cached auth records.
+    /// Tests that isolateTenantAuth=true bypasses a cached auth record, forcing a fresh login.
+    /// Expected: CreateCredential succeeds even when AZURE_MCP_AUTHENTICATION_RECORD contains invalid JSON,
+    /// because the early-return path never reads it.
     /// </summary>
     [Fact]
-    public void IsolateTenantAuth_CreatesCredentialSuccessfully()
+    public void IsolateTenantAuth_IgnoresCachedAuthRecord()
     {
-        // Act
-        var credential = CreateCustomChainedCredential(isolateTenantAuth: true);
+        using var env = new EnvironmentScope("AZURE_MCP_AUTHENTICATION_RECORD", "AZURE_TOKEN_CREDENTIALS");
+        Environment.SetEnvironmentVariable("AZURE_MCP_AUTHENTICATION_RECORD", "not-valid-json");
 
-        // Assert
-        Assert.NotNull(credential);
-        Assert.IsAssignableFrom<TokenCredential>(credential);
+        var type = GetCustomChainedCredentialType();
+        var createCredential = type.GetMethod("CreateCredential", BindingFlags.NonPublic | BindingFlags.Static);
+        Assert.NotNull(createCredential);
+
+        // With isolation + non-null tenantId: fires before reading the auth record — must not throw
+        var result = createCredential.Invoke(null, ["test-tenant-id", null, false, true]);
+        Assert.NotNull(result);
+        Assert.IsAssignableFrom<TokenCredential>(result);
+    }
+
+    /// <summary>
+    /// Tests that isolateTenantAuth=true respects prod mode — no interactive browser in non-interactive environments.
+    /// Expected: When AZURE_TOKEN_CREDENTIALS=prod, the isolation early-return is skipped and the
+    /// prod credential chain is returned instead.
+    /// </summary>
+    [Fact]
+    public void IsolateTenantAuth_RespectsProductionMode()
+    {
+        using var env = new EnvironmentScope("AZURE_TOKEN_CREDENTIALS");
+        Environment.SetEnvironmentVariable("AZURE_TOKEN_CREDENTIALS", "prod");
+
+        var type = GetCustomChainedCredentialType();
+        var createCredential = type.GetMethod("CreateCredential", BindingFlags.NonPublic | BindingFlags.Static);
+        Assert.NotNull(createCredential);
+
+        // In prod mode, isolation early-return is skipped — must return the prod chain credential
+        var result = createCredential.Invoke(null, ["test-tenant-id", null, false, true]);
+        Assert.NotNull(result);
+        Assert.IsAssignableFrom<TokenCredential>(result);
     }
 
     private static Type GetCustomChainedCredentialType()
--- a/core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
+++ b/core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
@@ -234,14 +234,38 @@ internal class CustomChainedCredential(string? tenantId = null, ILogger<CustomCh
 
     private static TokenCredential CreateBrowserCredential(string? tenantId, AuthenticationRecord? authRecord)
     {
-        string? clientId = Environment.GetEnvironmentVariable(ClientIdEnvVarName);
+        var browserCredential = CreateBrokerCredential(
+            tenantId,
+            !ShouldUseOnlyBrokerCredential() && authRecord is null,
+            authRecord);
 
+        return CreateTimeoutCredential(browserCredential);
+    }
+
+    private static TokenCredential CreateTenantIsolatedCredential(string tenantId)
+    {
+        var browserCredential = CreateBrokerCredential(
+            tenantId,
+            useDefaultBrokerAccount: false,
+            authRecord: null);
+
+        return new SafeTokenCredential(
+            CreateTimeoutCredential(browserCredential),
+            "InteractiveBrowserCredential");
+    }
+
+    private static InteractiveBrowserCredential CreateBrokerCredential(
+        string? tenantId,
+        bool useDefaultBrokerAccount,
+        AuthenticationRecord? authRecord)
+    {
         IntPtr handle = WindowHandleProvider.GetWindowHandle();
+        string? clientId = Environment.GetEnvironmentVariable(ClientIdEnvVarName);
 
         InteractiveBrowserCredentialBrokerOptions brokerOptions = new(handle)
         {
-            UseDefaultBrokerAccount = !ShouldUseOnlyBrokerCredential() && authRecord is null,
             TenantId = string.IsNullOrEmpty(tenantId) ? null : tenantId,
+            UseDefaultBrokerAccount = useDefaultBrokerAccount,
             AuthenticationRecord = authRecord,
             TokenCachePersistenceOptions = new TokenCachePersistenceOptions()
             {
@@ -259,46 +283,11 @@ internal class CustomChainedCredential(string? tenantId = null, ILogger<CustomCh
             brokerOptions.ClientId = clientId;
         }
 
-        var browserCredential = new InteractiveBrowserCredential(brokerOptions);
-
-        // Check for timeout value in the environment variable
-        string? timeoutValue = Environment.GetEnvironmentVariable(BrowserAuthenticationTimeoutEnvVarName);
-        int timeoutSeconds = 300; // Default to 300 seconds (5 minutes)
-        if (!string.IsNullOrEmpty(timeoutValue) && int.TryParse(timeoutValue, out int parsedTimeout) && parsedTimeout > 0)
-        {
-            timeoutSeconds = parsedTimeout;
-        }
-        return new TimeoutTokenCredential(browserCredential, TimeSpan.FromSeconds(timeoutSeconds));
+        return new InteractiveBrowserCredential(brokerOptions);
     }
 
-    private static TokenCredential CreateTenantIsolatedCredential(string tenantId)
+    private static TokenCredential CreateTimeoutCredential(TokenCredential credential)
     {
-        IntPtr handle = WindowHandleProvider.GetWindowHandle();
-        string? clientId = Environment.GetEnvironmentVariable(ClientIdEnvVarName);
-
-        InteractiveBrowserCredentialBrokerOptions brokerOptions = new(handle)
-        {
-            TenantId = tenantId,
-            UseDefaultBrokerAccount = false,
-            AuthenticationRecord = null,
-            TokenCachePersistenceOptions = new TokenCachePersistenceOptions()
-            {
-                Name = TokenCacheName,
-            }
-        };
-
-        if (CloudConfiguration != null)
-        {
-            brokerOptions.AuthorityHost = CloudConfiguration.AuthorityHost;
-        }
-
-        if (clientId is not null)
-        {
-            brokerOptions.ClientId = clientId;
-        }
-
-        var browserCredential = new InteractiveBrowserCredential(brokerOptions);
-
         string? timeoutValue = Environment.GetEnvironmentVariable(BrowserAuthenticationTimeoutEnvVarName);
         int timeoutSeconds = 300;
         if (!string.IsNullOrEmpty(timeoutValue) && int.TryParse(timeoutValue, out int parsedTimeout) && parsedTimeout > 0)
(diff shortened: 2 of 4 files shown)


newventurecap:
> @newventurecap please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree

newventurecap:
Adding context for the current CI failure pattern after latest update.

Observed failing jobs:
- Build linux_x64
- Build linux_arm64
- Build windows_x64
- Build macos_x64

Common failing step:
- `Run recorded tests`

Representative failing lines from build `6013273` logs:
- `System.InvalidOperationException: Test proxy restore failed with exit code 1`
- `Git ran into an unrecoverable error. Test-Proxy is exiting.`
- `error: pathspec '.' did not match any file(s) known to git`
- `##[error]PowerShell exited with code '1'`

Additional macOS x64 signature seen in same run:
- `error: RPC failed; curl 56 LibreSSL SSL_reAdding context for the current CI failure pattern after latest update.

Observed failing jobs:
- Build linux_x64
- Build linux_arm64
- Build windows_x64
- Build macos_x64

Common failing step:
- `Run recorded tests`

Representative failing lines from build `6013273` logs:
- `System.InvalidOperationException: Test proxy restore -proxy restore issue path for this pipeline run context?

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Azure.Mcp.Core.Services.Azure.Authentication;
+
+/// <summary>
+/// Tenant-scoped credential for explicit tenant switching scenarios.
+/// This class is isolated from <see cref="CustomChainedCredential"/> behavior so default auth
+/// flows are unchanged unless a tenant-specific credential is requested.
+/// </summary>
+internal class TenantAwareCredential(string tenantId, ILogger<CustomChainedCredential>? logger = null)
+    : CustomChainedCredential(tenantId, logger)
```
jongio: This inherits `CustomChainedCredential` but `GetToken`/`GetTokenAsync` bypass the base chain entirely when `_tenantId` is non-null (which is always the case, since the provider only creates this when tenantId isn't null). The inheritance is only used to access the static `CloudConfiguration` property.

Consider making this a standalone `TokenCredential` implementation, or better yet, parameterize `CustomChainedCredential.CreateBrowserCredential` to accept `useDefaultBrokerAccount` and `authRecord` so you don't need a subclass at all.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    private const string BrowserAuthenticationTimeoutEnvVarName = "AZURE_MCP_BROWSER_AUTH_TIMEOUT_SECONDS";
+    private const string ClientIdEnvVarName = "AZURE_MCP_CLIENT_ID";
+    private const string TokenCacheName = "azure-mcp-msal.cache";
```
jongio: These three constants (`BrowserAuthenticationTimeoutEnvVarName`, `ClientIdEnvVarName`, `TokenCacheName`) are identical copies of the ones in `CustomChainedCredential`. They're `private` in the base class so can't be accessed directly - but this is another signal that the subclass is duplicating rather than extending.

jongio (commented):
Thanks for tackling #1797 - isolating tenant-specific auth from the default chain is the right direction.

A few design and correctness issues to address before this can merge:

1. `TenantAwareCredential` inherits `CustomChainedCredential` but immediately bypasses the entire base chain in `GetToken`/`GetTokenAsync`. The inheritance isn't buying anything here except access to the static `CloudConfiguration` property. Consider either making this a standalone class or adding a parameterized factory method to the base class.

2. `CreateTenantScopedCredential` duplicates ~90% of `CustomChainedCredential.CreateBrowserCredential` (lines 221-258). The only meaningful differences are `UseDefaultBrokerAccount = false` and `AuthenticationRecord = null`. Refactoring `CreateBrowserCredential` to accept these as parameters would eliminate the duplication.

3. `SECURITY_HARDENED_ENTERPRISE_FORK.md` looks like fork-specific documentation. It shouldn't be part of the main repo PR.

4. The changelog entry uses section name `"Bugs Fixed"` but the template specifies `"Bug Fixes"` as the valid section name.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    private readonly string _tenantId = tenantId;
+    private readonly TokenCredential _tenantScopedCredential = CreateTenantScopedCredential(tenantId);
```
jongio: `_tenantScopedCredential` is constructed eagerly via the field initializer, which calls `WindowHandleProvider.GetWindowHandle()` at construction time. The base class uses a lazy pattern (credential is built on first `GetToken` call). This means the P/Invoke for the window handle runs inside the lock in `SingleIdentityTokenCredentialProvider`, blocking other threads. In headless environments where there's no window handle, this fails at construction rather than at first use.

Consider using `Lazy<TokenCredential>` or moving construction to `GetToken`/`GetTokenAsync`.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+    {
+        if (string.IsNullOrWhiteSpace(_tenantId))
+        {
```
jongio: This `string.IsNullOrWhiteSpace(_tenantId)` guard can never be true. `SingleIdentityTokenCredentialProvider` only creates `TenantAwareCredential` when `tenantId` is non-null (line 34 checks `tenantId is null` and returns the default credential). Same applies to `GetTokenAsync` below. This is dead code - either remove the fallback or document why it's kept as a defensive measure.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        return _tenantScopedCredential.GetTokenAsync(requestContext, cancellationToken);
+    }
+
+    private static TokenCredential CreateTenantScopedCredential(string tenantId)
+    {
+        // Explicit tenant switcher: avoid sticky-account behavior and stale auth record reuse.
+        // This enables a clean tenant-specific interactive auth path where MFA/2FA prompts can
+        // follow the external tenant policy.
```
jongio: This method is ~90% identical to `CustomChainedCredential.CreateBrowserCredential` (lines 221-258). The only differences: `UseDefaultBrokerAccount = false` and `AuthenticationRecord = null`. Consider refactoring `CreateBrowserCredential` to accept those two values as parameters, then call it from both places. That eliminates the duplication and keeps the browser credential logic in one spot.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
```diff
@@ -42,7 +42,7 @@ public Task<TokenCredential> GetTokenCredentialAsync(
             {
                 if (!_tenantSpecificCredentials.TryGetValue(tenantId, out tenantCredential))
                 {
-                    tenantCredential = new CustomChainedCredential(
+                    tenantCredential = new TenantAwareCredential(
                         tenantId,
                         _loggerFactory.CreateLogger<CustomChainedCredential>()
```
jongio: `CreateLogger<CustomChainedCredential>()` produces a logger with category `CustomChainedCredential`, but this is creating a `TenantAwareCredential`. Should be `CreateLogger<TenantAwareCredential>()` so log output correctly identifies the source.

servers/Fabric.Mcp.Server/SECURITY_HARDENED_ENTERPRISE_FORK.md
```diff
@@ -0,0 +1,48 @@
+# Security-Hardened Enterprise Fork (Fabric MCP)
```
jongio: This file describes fork-specific hardening rationale. It doesn't belong in a PR to the main repo. The relevant context for reviewers is already in the PR description and the code comments. Please remove this file.

servers/Fabric.Mcp.Server/changelog-entries/newventurecap-fix-tenant-switch-auth-1797.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Bugs Fixed"
```
jongio: Section should be `Bug Fixes` (not `Bugs Fixed`). Check the template at `servers/Fabric.Mcp.Server/changelog-entries/username-example-brief-description.yml` for valid section names.

newventurecap: Corrected to "Bug Fixes" per the Fabric server template.

servers/Fabric.Mcp.Server/SECURITY_HARDENED_ENTERPRISE_FORK.md
newventurecap: Agreed and Removed. The architectural context is now in the PR description and in the XML doc summary on CustomChainedCredential.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
newventurecap: TenantAwareCredential has been removed entirely (took the opportunity to rebase onto the current main so that we have a clean diff). The fix is now an isolateTenantAuth parameter on CustomChainedCredential itself — no subclass, no inheritance, no chain bypass. This file doesn't exist anymore on the PR.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
newventurecap: Removed with the class. The new CreateTenantIsolatedCredential method is private to CustomChainedCredential and reads the constants directly — no duplication.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
newventurecap: Same as before TenantAwareCredential removed and after rebase with main this is no longer the case, the isolateTenantAuth path in CreateCredential is only entered when tenantId is non-null and non-empty — the guard if (isolateTenantAuth && !string.IsNullOrEmpty(tenantId)) is the single entry point.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
newventurecap: TenantAwareCredential is gone now. isolateTenantAuth: true is now passed to CustomChainedCredential directly.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
newventurecap: Since rebasing, TenantAwareCredential.cs is removed from the PR. CustomChainedCredential already uses lazy initialization (_credential ??= CreateCredential(...)) — CreateTenantIsolatedCredential follows the same pattern, called only at first GetToken use

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/TenantAwareCredential.cs
newventurecap: Removed with the class. Rather than changing CreateBrowserCredential's signature, we added CreateTenantIsolatedCredential as a private method alongside it in CustomChainedCredential. The two methods have distinct, clearly named purposes. CreateBrowserCredential is untouched.

newventurecap:
@jongio Thank you for the thorough review and for the encouraging opening — it means a lot to know the direction is right.

All four points have been addressed in the updated commits. I also took the opportunity to rebase onto the current `main` so the diff is clean and there are no conflicts for you to resolve on merge — including picking up the `SafeTokenCredential` wrapper added in #2076 which we were missing in the original draft.

**1. `TenantAwareCredential` inheritance** — the class has been removed entirely. Rather than making it standalone or adding a factory method, I took the simpler path: `isolateTenantAuth` is now a constructor parameter on `CustomChainedCredential` itself. When `true`, `CreateCredential` takes an early-return path through a new private `CreateTenantIsolatedCredential` method, keeping everything in one class with no inheritance involved.

**2. Duplication of `CreateBrowserCredential`** — also resolved by the removal. `CreateTenantIsolatedCredential` lives alongside `CreateBrowserCredential` as a private method in `CustomChainedCredential`, with a clearly distinct name and purpose. Importantly, `CreateBrowserCredential` itself is untouched — no signature changes, no visibility changes — which keeps the diff minimal.

**3. `SECURITY_HARDENED_ENTERPRISE_FORK.md`** — removed. The relevant information for reviewers is in the PR description and in the XML doc on `CustomChainedCredential`.

**4. Changelog section name** — corrected to `"Bug Fixes"` per the Fabric server template. Apologies for the confusion — I had mistakenly cross-referenced the Azure server `CHANGELOG.md` which uses a slightly different convention. The Fabric template is the right source for this PR.

The net result is a much smaller diff than the original draft: two files modified, one test updated, one changelog entry — no new classes, no new files beyond the YAML. The commits are also split one-per-file to make the diff easier to review. Happy to make any further adjustments.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -138,6 +143,11 @@ private static TokenCredential CreateCredential(string? tenantId, ILogger<Custom
         }
 #endif
 
+        if (isolateTenantAuth && !string.IsNullOrEmpty(tenantId))
+        {
+            return CreateTenantIsolatedCredential(tenantId!);
+        }
```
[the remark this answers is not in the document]
newventurecap: Fixed. Added a prod mode guard to the isolateTenantAuth early-return — AZURE_TOKEN_CREDENTIALS=prod now bypasses the isolation path and falls through to the existing non-interactive chain as expected. A test (IsolateTenantAuth_RespectsProductionMode) covers this.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/SingleIdentityTokenCredentialProvider.cs
```diff
@@ -44,7 +44,8 @@ public Task<TokenCredential> GetTokenCredentialAsync(
                 {
                     tenantCredential = new CustomChainedCredential(
```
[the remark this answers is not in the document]
newventurecap: This is pre-existing code from upstream main — our change was only adding isolateTenantAuth: true inside the existing lock. Maybe worth addressing but outside of this PR.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
```diff
@@ -454,20 +454,36 @@ private static TokenCredential CreateCustomChainedCredential(bool forceBrowserFa
...
+    [Fact]
+    public void IsolateTenantAuth_CreatesCredentialSuccessfully()
+    {
+        // Act
+        var credential = CreateCustomChainedCredential(isolateTenantAuth: true);
+
+        // Assert
+        Assert.NotNull(credential);
+        Assert.IsAssignableFrom<TokenCredential>(credential);
+    }
```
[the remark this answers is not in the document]
newventurecap: Fixed. Replaced with two new tests that call the private CreateCredential via reflection with a non-null tenantId: IsolateTenantAuth_IgnoresCachedAuthRecord verifies the auth record env var is bypassed (invalid JSON does not throw), and IsolateTenantAuth_RespectsProductionMode verifies prod mode falls through correctly

jongio (commented):
Addresses my previous feedback well. Moving the isolation logic into CustomChainedCredential as a parameter instead of a subclass is the right call - cleaner, no misleading inheritance, and the lazy init pattern is preserved.

One remaining item inline about the duplication between the two Create* methods.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -260,6 +271,46 @@ private static TokenCredential CreateBrowserCredential(string? tenantId, Authent
         return new TimeoutTokenCredential(browserCredential, TimeSpan.FromSeconds(timeoutSeconds));
     }
 
+    private static TokenCredential CreateTenantIsolatedCredential(string tenantId)
```
jongio: This method shares ~80% of its body with `CreateBrowserCredential` (line 235). The meaningful differences are `UseDefaultBrokerAccount = false`, `AuthenticationRecord = null`, and the `SafeTokenCredential` wrapping.

Consider extracting a shared helper that accepts these as parameters, e.g.:

`csharp
private static InteractiveBrowserCredential CreateBrokerCredential(
    string tenantId, bool useDefaultBrokerAccount, AuthenticationRecord? authRecord)
`

Not blocking - both methods are private in the same class now, so this is easy to refactor later.

jongio (commented):
Addresses my previous feedback. The CreateBrokerCredential and CreateTimeoutCredential extraction cleanly eliminates the duplication I flagged between the two create methods. Both callers now delegate to the shared helpers with the right parameter differences. Looks good.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/CustomChainedCredential.cs
newventurecap: @jongio  - Thank you — really appreciate the detailed guidance throughout this review. The suggestions pushed the design to a much better place than where it started. Happy to address if questions come up during the merge process.

newventurecap:
@jongio @RickWinter - I noticed your [comment](https://github.com/microsoft/mcp/issues/1797#issuecomment-5049423704) on the original [issue](https://github.com/microsoft/mcp/issues/1797). This PR was deleted because Github-actions bot closed it, but since more evidence is now being sought from the OP, I will hold off before re-opening / modify anything.

PR #2071, opened 2026-03-16
Handle rate limiting, add runtime config and live tests for Functions toolset

manvkaur:
## What does this PR do?

Handle GitHub API rate limiting, add runtime configuration, and live tests for Azure Functions toolset.

### Key Fixes:

**Rate Limiting Handling:**
- Switch to raw GitHub URLs (~5000 req/hr) after fetching GitHub tree to minimize API usage (60 req/hr unauthenticated)
- Proper detection via `X-RateLimit-Remaining: 0` header (distinguishes from 403 permission errors)

**Runtime Configuration:**
- Language list now returns `runtimeVersions` with supported, preview, and default versions from CDN manifest

**Template Fetching (Tree API):**
- Switched from GitHub Contents API to Tree API to handle rate limiting
- Added 5MB size limit (`MaxTreeSizeBytes`)
- 12-hour cache duration for manifests and templates

**Live Tests:**
- 2 recorded tests for build pipeline (playback mode) - one per test class that fetches from CDN
- 18 tests marked `[LiveTestOnly]` - run only in Live mode
- Tests cover: language listing, template retrieval, runtime versions, error handling
- Cache-dependent tests cannot be reliably recorded due to non-deterministic test execution order

### Files Changed:
| Area | Changes |
|------|---------|
| Services | `FunctionsService`, `ManifestService` |
| Models | Removed redundant `[JsonPropertyName]` from 11 files |
| JSON Context | Consolidated into single `FunctionsJsonContext` |
| Tests | 115 unit tests, 20 live tests (2 recorded + 18 live-only) |

## GitHub issue number?
Related #1781

## Pre-merge Checklist
- [x] Required for All PRs
    - [x] Commit history is clean with descriptive messages
    - [x] Updated `servers/Azure.Mcp.Server/CHANGELOG.md`
- [x] For MCP tool changes:
    - [x] **One tool per PR**: This PR modifies the Functions toolset only
    - [ ] Updated documentation
    - [ ] Run `.\eng\scripts\Update-AzCommandsMetadata.ps1`


tools/Azure.Mcp.Tools.Functions/src/Services/ManifestService.cs
anannya03, line 85: If `json` exceeds more than the `MaxManifestSizeBytes`, the exception isn't caught here and it would bypass the fallback logic and propagate directly out of `FetchManifestAsync`. Is that the intended behavior?
Or do we want to add a catch block here, so that we can proceed to the fallback logic?

```
catch (InvalidOperationException ex)
{
    logger.LogError(ex, "Manifest response from {Url} exceeded size limit", url);
    return ManifestFetchResult.Failure(ex.Message);
}
```

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
anannya03: This method `ConvertToRawGitHubUrl` doesn't seem to be getting used anywhere. If this is not needed, we can clean this up and clear out the tests for this.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -13,21 +13,23 @@ namespace Azure.Mcp.Tools.Functions.Services;
...
     IHttpClientFactory httpClientFactory,
     ILanguageMetadataProvider languageMetadata,
     IManifestService manifestService,
+    ICacheService cacheService,
     ILogger<FunctionsService> logger) : IFunctionsService
 {
     private readonly ILanguageMetadataProvider _languageMetadata = languageMetadata ?? throw new ArgumentNullException(nameof(languageMetadata));
     private readonly IManifestService _manifestService = manifestService ?? throw new ArgumentNullException(nameof(manifestService));
+    private readonly ICacheService _cacheService = cacheService ?? throw new ArgumentNullException(nameof(cacheService));
+
```
anannya03, line 30: For consistency throughout the codebase and as an additional defense, let's also add null checks for httpClientFactory and logger-

`private readonly IHttpClientFactory _httpClientFactory = httpClientFactory ?? throw new ArgumentNullException(nameof(httpClientFactory));`
`private readonly ILogger _logger = logger ?? throw new ArgumentNullException(nameof(logger))`

And then replace all occurences of httpClientFactory and logger with _httpClientFactory and '_logger`.

tools/Azure.Mcp.Tools.Functions/src/Models/TemplateManifest.cs
```diff
@@ -25,4 +25,11 @@ public sealed class TemplateManifest
 
     [JsonPropertyName("templates")]
     public IReadOnlyList<TemplateManifestEntry> Templates { get; init; } = [];
+
+    /// <summary>
+    /// Runtime version information for each supported language.
+    /// Keys are language names (e.g., "Python", "JavaScript", "TypeScript", "Java", "CSharp", "PowerShell").
+    /// </summary>
+    [JsonPropertyName("runtimeVersions")]
```
anannya03: `[JsonPropertyName("<propertyName>")]` isn't needed for any of the properties in this file because it is just the camelcase version of the propertyName. This is already enforced via this line-
`[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase, DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull)]
`
in `FunctionsJsonContext.cs` file.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -347,15 +378,143 @@ private async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaCont
...
+            results.Add((item.Path, relativePath, item.Size));
+        }
+
+        return results;
+    }
+
+    /// <summary>
+    /// Fetches files by downloading the repository zipball.
     /// </summary>
     internal async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaArchiveAsync(
```
anannya03, line 511: I think we should either make this private as well like `FetchTreeFromGitHubAsync`. 
If we want to keep it as internal, we should add the valid URL check to make sure that in the future if there's any other class calling this method directly, there's a check in place for the `repositoryURL`. 

The validURL check I'm referring to- `GitHubUrlValidator.IsValidRepositoryUrl(entry.RepositoryUrl)`

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -277,68 +298,78 @@ internal async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesAsync(
...
                 continue;
             }
 
-            // Validate URL points to GitHub domain (SSRF prevention)
-            if (entry.DownloadUrl is null || !GitHubUrlValidator.IsValidGitHubUrl(entry.DownloadUrl))
-            {
-                logger.LogWarning("Skipping file {Name} - invalid or missing download URL", entry.Name);
-                continue;
-            }
+            var rawUrl = $"https://raw.githubusercontent.com/{repoPath}/{DefaultBranch}/{fullPath}";
```
anannya03: Just trying to understand, how does the `rawUrl` help us with the rate limiting?

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -347,15 +378,143 @@ private async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaCont
...
     }
 
     /// <summary>
-    /// Fetches files by downloading the repository as a zipball (efficient for root or large folders).
+    /// Fetches tree response from GitHub API.
+    /// </summary>
+    private async Task<GitHubTreeResponse> FetchTreeFromGitHubAsync(string treeUrl, string repoPath, CancellationToken cancellationToken)
+    {
+        using var client = httpClientFactory.CreateClient();
+        client.DefaultRequestHeaders.UserAgent.ParseAdd("Azure-MCP-Server/1.0");
```
anannya03: nit: I know we've added `Azure-MCP-Server/1.0` as the HTTP header's user agent throughout the code, but I'm wondering if there's a better way to add the version, because the 1.0 is not really correct.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -240,8 +246,7 @@ internal static string ConvertToRawGitHubUrl(string repositoryUrl, string folder
     }
 
     /// <summary>
-    /// Fetches all files from a template directory. Uses GitHub's zipball API for root/large folders,
-    /// or the Contents API for specific subdirectories.
+    /// Fetches all files from a template directory. Results are cached.
     /// </summary>
     internal async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesAsync(
```
anannya03: I'd make this either private and modify the testcases if needed or add the validURL check here as well to prevent future oversights.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -347,15 +378,143 @@ private async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaCont
...
+            logger.LogError(ex, "Failed to fetch tree from {Url}", treeUrl);
+            throw new InvalidOperationException(
+                $"Failed to fetch template files from GitHub. Check your network connection. Details: {ex.Message}", ex);
+        }
+
+        using (response)
+        {
+            // Check for rate limiting - verify via X-RateLimit-Remaining header
+            // HTTP 403 can also mean permission denied, so we check the header
+            if (response.StatusCode == System.Net.HttpStatusCode.TooManyRequests ||
```
anannya03: nit: let's move the full package (System.Net.HttpStatusCode.TooManyRequests) to the import statements at the top of the file.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -347,15 +378,143 @@ private async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaCont
...
+                (response.StatusCode == System.Net.HttpStatusCode.Forbidden && IsRateLimited(response)))
```
anannya03: same here as well

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
```diff
@@ -478,112 +625,21 @@ internal async Task<IReadOnlyList<ProjectTemplateFile>> FetchTemplateFilesViaArc
...
-        }
-
-        return allFiles;
-    }
 }
 
 /// <summary>
 /// AOT-safe JSON serialization context for CDN manifest and GitHub API deserialization.
 /// </summary>
 [JsonSerializable(typeof(TemplateManifest))]
```
anannya03: This whole block should be moved into the file `FunctionsJsonContext.cs`, because defining JsonSerializable is the purpose of that file. All of these should be present there. There're also a few duplicates between these and the ones mentioned in that file and should be cleaned up. 

`[JsonSerializable(typeof(TemplateManifest))]
[JsonSerializable(typeof(TemplateManifestEntry))]
[JsonSerializable(typeof(Dictionary<string, RuntimeVersionInfo>))]
[JsonSerializable(typeof(GitHubTreeResponse))]
[JsonSerializable(typeof(GitHubTreeItem))]
[JsonSerializable(typeof(List<GitHubTreeItem>))]
[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]`

servers/Azure.Mcp.Server/changelog-entries/1773697825677.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Features Added"
```
anannya03: This should be under the section "Other Changes".

tools/Azure.Mcp.Tools.Functions/tests/Azure.Mcp.Tools.Functions.LiveTests/Template/TemplateGetCommandLiveTests.cs
```diff
@@ -0,0 +1,322 @@
...
+        Assert.Equal("csharp", templateResult.FunctionTemplate.Language);
+    }
+
+    #endregion
+
+    #region Caching Tests
+
+    [Fact]
+    [LiveTestOnly]
+    public async Task ExecuteAsync_LanguageListThenTemplate_UsesSharedCache()
```
anannya03, line 215: I see that most of the Live Tests are marked as [LiveTestOnly]. Can we not have recorded tests for these to have better CI coverage?

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
manvkaur: I considered making this private, but kept it as internal to enable direct unit testing of edge case handling. The tests inject controlled mock data to verify specific behaviors, which wouldn't be possible through the public API alone.
Regarding URL validation: the existing ExtractGitHubRepoPath() call already validates the URL format and throws ArgumentException for invalid  inputs, so adding IsValidRepositoryUrl() beforehand would be redundant.

tools/Azure.Mcp.Tools.Functions/src/Services/FunctionsService.cs
manvkaur: Great question! The GitHub API (api.github.com) has a rate limit of 60 requests/hour for unauthenticated users. However, raw.githubusercontent.com  is served via CDN with a much higher limit (approx. 5000 requests/hour).

 Our approach: we make one API call to fetch the Tree (file listing), then use raw URLs for the actual file contents. This minimizes API usage while  allowing us to fetch many template files without hitting rate limits.

tools/Azure.Mcp.Tools.Functions/tests/Azure.Mcp.Tools.Functions.LiveTests/Template/TemplateGetCommandLiveTests.cs
manvkaur: The [LiveTestOnly] tests rely on an in-memory cached manifest populated by the first test in each class. xUnit runs tests in non-deterministic order, so in Playback mode a cache-dependent test might run first with an empty cache, try to call the CDN, but have no recorded response.

The CDN functionality is well covered:
   - 2 recorded live tests - validate real CDN integration and manifest structure
   - 118 unit tests - mock HTTP responses to test CDN fetching, caching, fallback logic, error handling, and edge cases
   
The remaining 18 live tests focus on query variations using cached data and are validated during live mode.

tools/Azure.Mcp.Tools.Functions/src/Models/LanguageInfo.cs
```diff
@@ -1,76 +1,59 @@
...
-    [JsonPropertyName("templateParameters")]
     public IReadOnlyList<TemplateParameter>? TemplateParameters { get; init; }
 
     /// <summary>
     /// Notes about why this language is recommended for its runtime.
     /// For example, "Recommended for Node.js runtime for type safety and better tooling."
     /// Null if not a recommended choice.
     /// </summary>
-    [JsonPropertyName("recommendationNotes")]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
anannya03, line 74: Do we not want this field to be included in the JSON output when it is null?

manvkaur: The WhenWritingNull behavior is preserved! The FunctionsJsonContext has DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull at the context  level, which applies to all types. The explicit attribute was redundant, so I removed it along with the other [JsonPropertyName] attributes.

I wanted to omit this when null to keep the JSON output clean and concise - this reduces payload size, token usage and provides a better experience for AI agents consuming these responses, as they don't need to parse or filter out null fields. For example, RecommendationNotes only applies to TypeScript (recommended over JavaScript), so omitting it for other languages avoids clutter. In future, if other languages have more notes, they can be added.

anannya03: > I wanted to omit this when null to keep the JSON output clean and concise - this reduces payload size, token usage and provides a better experience for AI agents consuming these responses, as they don't need to parse or filter out null fields. For example, RecommendationNotes only applies to TypeScript (recommended over JavaScript), so omitting it for other languages avoids clutter. In future, if other languages have more notes, they can be added.

I had typed my question wrong. I was trying to understand why we want to omit this. Makes sense. Thanks.

tools/Azure.Mcp.Tools.Functions/src/Services/Helpers/HttpClientHelper.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation.
```
anannya03: Let's revert this file as well, since we're now just going to remove version from the user agent and just keep it as `Azure-MCP-Server`.

anannya03:
@manvkaur, please run all the three mcp tools locally using MCP server to make sure that nothing has broken and there are no regressions.

manvkaur:
> please run all the three mcp tools locally using MCP server to make sure that nothing has broken and there are no regressions.

@anannya03 - Yes, I did test with random prompts as well, but here is the summary using E2E prompts you added in initial PR:

   | Prompt | Command | Result |
   |--------|---------|--------|
   | What languages does Azure Functions support? | `functions language list` | 6 languages returned |
   | What runtime versions are available? | `functions language list` | runtimeVersions for each language |
   | Set up a new Python project | `functions project get --language python` | initInstructions, projectFiles |
   | Generate TypeScript project files | `functions project get --language typescript` | With templateParameters |
   | Java with JDK 21 | `functions project get --language java` | javaVersion parameter |
   | C# triggers and bindings | `functions template get --language csharp` | 12+ templates listed |
   | Python templates | `functions template get --language python` | 25 templates listed |
   | Timer trigger C# | `functions template get --language csharp --template timer-trigger-csharp` | 6 files returned |
   | Cosmos DB trigger Java | `functions template get --language java --template cosmos-trigger-java` | 5 files returned |
   | MCP Tool trigger TypeScript | `functions template get --language typescript --template mcp-tool-trigger-typescript` | 8 files returned |

   **Test Summary:**
   - Unit Tests: 115 passed
   - Live Tests: 20 passed (Live mode runs all tests against real CDN/GitHub endpoints for additional validation)
   - Local CLI: All prompts working

anannya03:
Looks good. :)

PR #2124, opened 2026-03-19
Start a host when calling commands via CLI to capture telemetry

alzimmermsft:
## What does this PR do?

When processing a call to the MCP server executable directly the following is performed (rather than just accepting the command to call from the manually managed `ServiceCollection`):

Bind the manually maintained `ServiceCollection` with just the commands to start the MCP server. If the command is found invoke it as-is, as it will handle everything else for us. If the command isn't found, assume this is a direct tool call, rather than process start, bind everything up as if we're running in STDIO mode and run the command, capturing a `CommandExecuted` telemetry event.

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:


core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -103,7 +103,7 @@ private static void EnableAzureMonitor(this IServiceCollection services)
 #endif
 
         var enableOtlp = Environment.GetEnvironmentVariable("AZURE_MCP_ENABLE_OTLP_EXPORTER");
-        if (!string.IsNullOrEmpty(enableOtlp) && bool.TryParse(enableOtlp, out var shouldEnable) && shouldEnable)
+        if (true)
```
jongio: This replaces the `AZURE_MCP_ENABLE_OTLP_EXPORTER` env var check with unconditional `true`. Every user would get OTLP export enabled regardless of config - users without an OTLP collector would see failed connection attempts. The `enableOtlp` variable on the line above becomes dead code too. Should be reverted.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -64,7 +66,40 @@ private static async Task<int> Main(string[] args)
             var commandFactory = serviceProvider.GetRequiredService<ICommandFactory>();
             var rootCommand = commandFactory.RootCommand;
             var parseResult = rootCommand.Parse(args);
-            var status = await parseResult.InvokeAsync();
+            int status = 0;
+
+            if (parseResult.Errors.Count > 0)
```
jongio: `parseResult.Errors.Count > 0` doesn't distinguish "command not found in the ServerSetup subset" from genuine parse errors. If someone passes bad args to a recognized ServerSetup command (like `server start --invalid`), this would still spin up a full Host with all 50+ areas, re-parse, and fail again.

Consider checking whether the resolved command is a recognized ServerSetup command rather than using error count as a proxy.

core/Microsoft.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -327,6 +329,34 @@ private void CustomizeHelpOption(Command command)
...
+
+        var commandResult = parseResult.CommandResult;
+        var fullCommandName = commandResult.Command.Name;
+        while (commandResult.Parent is not null
+            && commandResult?.Parent is CommandResult
+            && (commandResult.Parent as CommandResult)?.Command.Name != RootCommand.Name)
+        {
+            commandResult = (commandResult.Parent as CommandResult)!;
+            fullCommandName = commandResult.Command.Name + "_" + fullCommandName;
+        }
```
jongio: The null-checking here mixes patterns in a way that's hard to follow - `commandResult.Parent is not null` then `commandResult?.Parent is CommandResult` (redundant null-conditional since `commandResult` can't be null after the prior access), then `as CommandResult` cast, then `!` null-forgiving inside the body. Can be simplified with pattern matching:

```suggestion
        while (commandResult.Parent is CommandResult parentResult
            && parentResult.Command.Name != RootCommand.Name)
        {
            fullCommandName = parentResult.Command.Name + "_" + fullCommandName;
            commandResult = parentResult;
        }
```

jongio (commented):
The telemetry tag injection for CLI commands fills the gap where `CommandFactoryToolLoader` already handles this for MCP server calls - good addition.

Two things to address before un-drafting:

1. **`if (true)` in `OpenTelemetryExtensions.cs`** - this unconditionally enables OTLP export for all users. Looks like a debug leftover that needs to be reverted (see inline).

2. **No tests** - `InjectToolAreaAndName`, the Host-creation fallback path, and the `ConfigureServices` overload all lack test coverage. The parse tree walking logic in particular warrants unit tests for edge cases (single-segment commands, deep nesting, commands at different depths).

Also, the 50+ line XML doc comment on `ConfigureServices` is now duplicated across both overloads. Since the original is now a one-liner delegate, consider trimming it to a brief note and keeping the full comment only on the implementation overload.

core/Microsoft.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -327,6 +329,34 @@ private void CustomizeHelpOption(Command command)
...
+
+        var index = fullCommandName.IndexOf('_');
+        activity.SetTag(TagName.ToolArea, index == -1 ? fullCommandName : fullCommandName.Substring(0, index))
```
jongio: Nit: `fullCommandName[..index]` is the more idiomatic C# range syntax.

core/Microsoft.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -327,6 +329,33 @@ private void CustomizeHelpOption(Command command)
         }
     }
 
+    /// <summary>
+    /// Injects tool area and name tags into the activity based on the command being executed. The full command name
+    /// is parsed to determine the tool area and name, which are then added as tags to the activity.
+    /// </summary>
+    /// <param name="activity">The activity to inject tool area and name tags into.</param>
+    /// <param name="parseResult">The parsing result for the command.</param>
+    private void InjectToolAreaAndName(Activity? activity, ParseResult parseResult)
```
jongio, line 522: No tests cover this method. It's the core of the PR, and the parent-walk logic has a few cases worth exercising: a multi-level command (`azmcp storage account list` should produce area=`storage`, name=`storage_account_list`), a root-adjacent single-segment command, and the `IndexOf('_') == -1` fallback. `ToolLoaderTelemetryTests` is the pattern to follow - it uses `ActivityListener` to capture tags and assert on them. The new cases would fit well next to the existing ones in `CommandFactoryTests`.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -50,23 +51,47 @@ private static async Task<int> Main(string[] args)
...
+                (extendedCommand.BaseCommand is ServiceStartCommand || extendedCommand.BaseCommand is PluginTelemetryCommand))
+            {
+                // One of the special commands that need to be handled differently.
+                status = await parseResult.InvokeAsync();
+            }
+            else
+            {
+                // Command wasn't one of the registered ServerSetup commands, so bind up a Host of all the services
+                // to run the command.
+                var builder = Host.CreateApplicationBuilder();
```
jongio, line 128: This whole branch is untested. At minimum something that proves invoking a non-server-start command via the CLI surface ends up with `ToolArea` / `ToolName` on the activity - that's the user-visible outcome the PR is buying. Hard to unit-test against `Main` directly, but factoring the host-building logic into a static helper (mirroring `SetupBasicInitializer`) would make it testable, or this could live as an end-to-end test that spawns the binary with a fake OTLP collector.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -50,23 +51,47 @@ private static async Task<int> Main(string[] args)
...
+            var serviceProvider = SetupBasicInitializer();
 
             var commandFactory = serviceProvider.GetRequiredService<ICommandFactory>();
             var rootCommand = commandFactory.RootCommand;
             var parseResult = rootCommand.Parse(args);
-            var status = await parseResult.InvokeAsync();
+            var command = parseResult.CommandResult.Command;
+            int status = 0;
+
+            if (command is ExtendedCommand extendedCommand &&
```
jongio, line 118: Hardcoding `ServiceStartCommand` and `PluginTelemetryCommand` here couples `Program.cs` to those two specific types. If a third 'doesn't need the full host' command shows up later (version dump, config print, etc.) this branch has to be touched and it's easy to miss. A marker interface on `IBaseCommand` (something like `IMinimalHostCommand`) or a `bool RequiresFullHost` property with a default would localize the decision to the command itself. Not a blocker, just something to consider while this is still in draft.

jongio (commented):
The Design update addressed my earlier inline comments well. The ExtendedCommand type check is a much stronger signal than the old `parseResult.Errors.Count > 0` heuristic.

Remaining concern is test coverage. `InjectToolAreaAndName` is the core mechanism added by this PR and it has zero unit tests, and the new Host-fallback branch in `Main` is also unexercised. Two small inline notes below.

servers/Azure.Mcp.Server/src/Program.cs
alzimmermsft: Given those are tools of extraordinary circumstances, I'm going to leave this as-is for now as this change should be more a moment in time thing until how the application is started is restructured.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -278,6 +301,43 @@ internal static void ConfigureServices(IServiceCollection services)
...
+            ActivatorUtilities.CreateInstance<ResourcePluginFileReferenceAllowlistProvider>(sp, thisAssembly, $"allowed-plugin-file-references.json"));
+        services.AddSingleton<IPluginSkillNameAllowlistProvider>(sp =>
+            ActivatorUtilities.CreateInstance<ResourcePluginSkillNameAllowlistProvider>(sp, thisAssembly, $"allowed-skill-names.json"));
+        services.AddSingleton<ICommandFactory, CommandFactory>();
+
+        foreach (IAreaSetup area in Areas.Where(area => area is ServerSetup))
+        {
+            services.AddSingleton(area);
+            area.ConfigureServices(services);
+        }
```
jongio, line 369: `PluginTelemetryCommand.ExecuteAsync` resolves `ICommandFactory` from `_serviceProvider` to validate `--tool-name`. But since this command runs in the lightweight path (basic initializer), that factory only has ServerSetup commands registered.

Result: `commandFactory.FindCommandByName("storage_account_list")` returns null, so legitimate tool names get rejected with a 403 "not a recognized azmcp command" response.

Two possible fixes:
- Have `PluginTelemetryCommand` also go through the Host path (it needs a telemetry host anyway via `CreateStdioHost`).
- Or defer tool-name validation until after the telemetry host is built, using that host's command factory instead.

servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -278,6 +301,43 @@ internal static void ConfigureServices(IServiceCollection services)
...
 
+    /// <summary>
+    /// Creates a very small service provider that allows for 'server start' and 'plugin telemetry' commands to be
+    /// executed without initializing the full service, which reduces startup time for these commands significantly.
+    /// <para>
+    /// This is a tightly bound service provider that manually configures only what is necessary for those two
+    /// commands. Any changes to the dependencies of those commands may require changes here.
+    /// </para>
+    /// </summary>
+    /// <returns>The basic initializer serivce provider.</returns>
```
jongio: Typo: "serivce provider" should be "service provider".

jongio (commented):
The merge commit looks clean - no conflicts in PR code.

My two open threads from last review still apply:

1. `InjectToolAreaAndName` has zero test coverage. The parent-walk + underscore-split logic is the core of this PR's CLI telemetry feature and warrants unit tests (multi-level command, single-segment, edge cases).

2. The Host-fallback branch (`else` path in `Main`) is also untested. A test that proves a non-server command ends up with `ToolArea`/`ToolName` on the activity would be the minimum bar.

Beyond that, there's a real issue with `PluginTelemetryCommand` validation in the lightweight path (see inline).

jongio (approved):
Rereviewed the 4 new commits since my last pass. All prior findings are addressed:

- PluginTelemetryCommand validation: fixed by creating the host before tool-name validation so ICommandFactory has all commands.
- InjectToolAreaAndName: the naive IndexOf('_') split is gone, replaced with list-based command tree walking. Much more reliable.
- SetupBasicInitializer: removed in favor of `ConfigureServices(services, areaFilter)` which selectively registers only the target area. Clean design.
- GetTargetAreaName has solid test coverage (4 branches, edge cases like `--`, `tools`, typos).

The parallel telemetry init and area-filtered DI are nice latency wins. MSAL exception handling in BaseCommand/GlobalCommand is a good addition.

One minor note for future: the Fabric/Template ExtendedCommand routing path (the `else` branch that creates a full Host) doesn't have dedicated tests yet. Not blocking.

PR #2188, opened 2026-03-24
Rename IlcOptimizationPreference to OptimizationPreference

jtschuster:
IlcOptimizationPreference is not used anywhere in the nativeaot msbuild files. OptimizationPreference is the correct property to set.


jongio (approved):
Reviewed the NativeAOT build config change. OptimizationPreference is the correct MSBuild property - IlcOptimizationPreference was a no-op. Clean fix. LGTM.

PR #2214, opened 2026-03-26
Fix elicitation for Github Copilot CLI

KarishmaGhiya:
update schema to use the TitledSingleSelectEnumSchema so that elicitation works on both Github Copilot cli and vscode


KarishmaGhiya:
Test run - https://dev.azure.com/azure-sdk/public/_build/results?buildId=6070084&view=results

PR #2261, opened 2026-03-27
Simplify command and parse result handling

alzimmermsft:
## What does this PR do?

Proof of concept for simplifying how `CommandResult` and `ParseResult` is handled with `Option` and `Option<T>`.

This changes the need to use `Option.Name` to retrieve the value when using the extension methods, instead the extension method handles using the name of the `Option.Name` to handle everything properly.

`Option.Name` is used for lookup as we have extension methods for `Options` which changes their required or optional behavior, but this creates a new instance which changes how lookup is handled by `CommandResult` and `ParseResult` as it's just looking up against a dictionary, and since the instance is different this could fail.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:


tools/Azure.Mcp.Tools.Compute/src/Commands/Vmss/VmssUpdateCommand.cs
```diff
@@ -64,8 +64,8 @@ protected override void RegisterOptions(Command command)
         command.Validators.Add(commandResult =>
         {
             // Custom validation: At least one update property must be specified
-            if (string.IsNullOrEmpty(commandResult.GetValueOrDefault<string>(ComputeOptionDefinitions.UpgradePolicy.Name)) &&
-                !commandResult.GetValueOrDefault<int?>(ComputeOptionDefinitions.Capacity.Name).HasValue &&
+            if (string.IsNullOrEmpty(commandResult.GetValueOrDefault(ComputeOptionDefinitions.UpgradePolicy)) &&
```
jongio: Only the first 2 conditions were migrated to the new pattern. Lines 69-73 (`VmSize`, `Overprovision`, `EnableAutoOsUpgrade`, `ScaleInPolicy`, `Tags`) still use `.Name`. Same gap in `BindOptions` (lines 84-91). Since this PoC demonstrates the new API, migrating the whole method makes the before/after comparison cleaner.

tools/Azure.Mcp.Tools.Compute/src/Commands/Vmss/VmssUpdateCommand.cs
```diff
@@ -64,8 +64,8 @@ protected override void RegisterOptions(Command command)
...
+                !commandResult.HasOptionResult(ComputeOptionDefinitions.Capacity) &&
```
jongio, line 67: `Capacity` is `Option<int>` but the old code queried `<int?>` - `FindOptionTByName<int?>` can't match `Option<int>`, so this check always evaluated as "capacity not provided" even when it was. Switching to `HasOptionResult` fixes this pre-existing bug. Worth calling out in the PR description.

Also note the semantic shift: `HasOptionResult` checks token presence while the old code checked `.HasValue` on the returned nullable. For `Option<int>` without a default these are equivalent, but be intentional about which semantic you want.

tools/Azure.Mcp.Tools.Monitor/src/Commands/WebTests/WebTestsCreateOrUpdateCommand.cs
```diff
@@ -101,25 +101,25 @@ protected override WebTestsCreateOrUpdateOptions BindOptions(ParseResult parseRe
...
-        options.IgnoreStatusCode = parseResult.CommandResult.GetValueWithoutDefault<bool?>(MonitorOptionDefinitions.WebTest.IgnoreStatusCode.Name);
-        options.ParseRequests = parseResult.CommandResult.GetValueWithoutDefault<bool?>(MonitorOptionDefinitions.WebTest.ParseRequests.Name);
-        options.RequestBody = parseResult.CommandResult.GetValueWithoutDefault<string>(MonitorOptionDefinitions.WebTest.RequestBody.Name);
-        options.RetryEnabled = parseResult.CommandResult.GetValueWithoutDefault<bool?>(MonitorOptionDefinitions.WebTest.RetryEnabled.Name);
-        options.SslCheck = parseResult.CommandResult.GetValueWithoutDefault<bool?>(MonitorOptionDefinitions.WebTest.SslCheck.Name);
-        options.SslLifetimeCheckInDays = parseResult.CommandResult.GetValueWithoutDefault<int?>(MonitorOptionDefinitions.WebTest.SslLifetimeCheckInDays.Name);
-        options.TimeoutInSeconds = parseResult.CommandResult.GetValueWithoutDefault<int?>(MonitorOptionDefinitions.WebTest.TimeoutInSeconds.Name);
+        options.AppInsightsComponentId = parseResult.CommandResult.GetValueWithoutDefault(MonitorOptionDefinitions.WebTest.AppInsightsComponentId);
```
jongio, line 103: `ResourceName` (line 102) and `ResourceGroup` (line 103) still use the old `.Name` pattern while everything below was migrated.

core/Microsoft.Mcp.Core/src/Extensions/CommandResultExtensions.cs
```diff
@@ -108,6 +107,21 @@ public static bool TryGetValue<T>(this CommandResult commandResult, string optio
...
+        var option = FindOptionTByName<T>(commandResult, optionName);
+        if (option is null)
+        {
+            return default;
+        }
+
+        return GetValueOrDefaultImpl(commandResult, option);
+    }
+
+    private static T? GetValueOrDefaultImpl<T>(this CommandResult commandResult, Option<T> option)
```
jongio: `this` on a private method is a no-op since it can't be discovered as an extension externally. Same applies to `GetValueWithoutDefaultImpl`. Drop the `this` modifier to make the intent clear.

jongio (commented):
Design direction is right - routing through `option.Name` to find the registered instance cleanly fixes the AsRequired/AsOptional identity mismatch. A few things to address across the surface:

`TryGetValue<T>(CommandResult, Option<T>, out T?)` (CommandResultExtensions.cs line 58) still calls `commandResult.GetResult(option)` directly. It hasn't been migrated to name-based routing and has the same instance-mismatch problem as the methods you've fixed.

This also needs at least one test demonstrating the fix works: register an option via `AsRequired()` on the command, then call `GetValueOrDefault()` with the original static definition. That's the scenario this PR exists for - show it works.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/CommandResultExtensionsTests.cs
```diff
@@ -503,4 +505,101 @@ public void GetValueWithoutDefault_WithStringOptionName_WithNonExistentOption_Re
         // Assert
         Assert.Null(result);
     }
+
+    [Theory]
+    [InlineData(new string[0], false, false)]
+    [InlineData(new string[0], true, false)]
+    [InlineData(new[] { "--name", "value" }, false, true)]
+    [InlineData(new[] { "--name", "value" }, true, true)]
+    public void HasOptionResult_TypeOptionWithVariousStringScenarios_ReturnsExpectedResult(string[] args, bool changedRequiredness, bool expected)
```
jongio: These tests use the same variable for both registration and lookup - `option = option.AsRequired()` then `HasOptionResult(option)` passes the same instance that was registered. The actual bug this PR fixes is when two different instances are involved. Something like:

```csharp
var original = new Option<string?>("--name");
var command = new Command("test") { original.AsRequired() };
var parseResult = command.Parse("--name value");
var result = parseResult.CommandResult.GetValueOrDefault(original);
Assert.Equal("value", result);
```

This also needs coverage for `GetValueOrDefault`, `TryGetValue`, and `GetValueWithoutDefault` - not just `HasOptionResult`.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/CommandResultExtensionsTests.cs
```diff
@@ -3,6 +3,8 @@
 
 using System.CommandLine;
 using Microsoft.Mcp.Core.Extensions;
+using Microsoft.Mcp.Core.Models.Option;
+using NSubstitute;
```
jongio: `NSubstitute` isn't used in any of the new tests - can drop this import.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/CommandResultExtensionsTests.cs
```diff
@@ -503,4 +505,101 @@ public void GetValueWithoutDefault_WithStringOptionName_WithNonExistentOption_Re
...
+        // Assert
+        Assert.Equal(expected, hasResult);
+    }
+
+    [Theory]
+    [InlineData(new string[0], false, false)]
+    [InlineData(new string[0], true, false)]
+    [InlineData(new[] { "--name", "value" }, false, true)]
+    [InlineData(new[] { "--name", "value" }, true, true)]
+    public void HasOptionResult_UntypeOptionWithVariousStringScenarios_ReturnsExpectedResult(string[] args, bool changedRequiredness, bool expected)
```
jongio: Typo: `UntypeOption` should be `UntypedOption` to match the companion test `UntypedOptionWithVariousBoolScenarios`.

core/Microsoft.Mcp.Core/src/Extensions/CommandResultExtensions.cs
```diff
@@ -8,7 +8,25 @@ namespace Microsoft.Mcp.Core.Extensions;
...
+    /// Checks if an option with the specified name has a result in the command result.
+    /// </summary>
+    /// <param name="commandResult">The command result to check.</param>
+    /// <param name="optionName">The name of the option to check for (including -- prefix).</param>
+    /// <returns>True if the option has a result, false otherwise.</returns>
+    public static bool HasOptionResult(this CommandResult commandResult, string optionName)
     {
+        // Find the option by name in the command
+        var option = commandResult.Command.Options
+            .FirstOrDefault(o => o.Name == optionName || o.Aliases.Contains(optionName));
```
jongio, line 26: Worth noting for the design discussion: every typed overload now does a `FirstOrDefault` scan through `commandResult.Command.Options`. Previously `GetResult(option)` did dictionary-style identity lookup. N is typically 5-20 so it isn't a real concern, but if this pattern gets adopted repo-wide (100+ commands), hot paths might want to consider caching the lookup.

jongio (commented):
Design direction is right - routing through `option.Name` to find the registered instance fixes the AsRequired/AsOptional identity mismatch. A few things to address:

1. The key test is still missing - see inline comment on the test file.
2. Unused `using NSubstitute` in the test file.
3. Minor test name typo (`UntypeOption` should be `UntypedOption`).
4. Every typed overload now does an O(N) `FirstOrDefault` scan. Fine for small N, but worth noting as a design trade-off.

The `HasAnyRetryOptions` simplification and tool command migrations look great.

core/Microsoft.Mcp.Core/src/Extensions/CommandResultExtensions.cs
alzimmermsft: Not the most worried about this as this shouldn't be a hot path. If this becomes a problem we may just want to look into removing the ability to `.AsRequired()` or `.AsOptional()`.

tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
```diff
@@ -307,7 +307,7 @@ public static class ComputeOptionDefinitions
 
     public const string CapacityName = "capacity";
 
-    public static readonly Option<int> Capacity = new($"--{CapacityName}")
+    public static readonly Option<int?> Capacity = new($"--{CapacityName}")
```
alzimmermsft, line 310: This fixes a bug where parsing was using `.Name` and `<int?>` for the type, while the Option was defined as a non-nullable int. Which when switching to the new design returned `0` instead of `null` which was causing testing failures.

Same applies for the other cases in this file adding `?`

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/CommandResultExtensionsTests.cs
```diff
@@ -3,190 +3,31 @@
...
-        var command = new Command("test") { option };
-        var parseResult = command.Parse("");
-
-        // Act
-        var result = parseResult.CommandResult.GetValueOrDefault(option);
-
-        // Assert
-        Assert.Null(result);
+        Assert.Equal(expected, result);
     }
```
jongio, line 31: The Copilot bot concern about generic `[Theory]` methods not working is incorrect - CI confirms xUnit handles type inference from `[InlineData]` fine. But the `[InlineData(new string[0], null)]` rows infer `T` as `object`, so those cases create `Option<object?>` - a type nobody uses in practice. The null-case rows pass but don't actually test any real option type like `string?` or `int?`.

Not a blocker since the typed rows (`"test"`, `42`, `true`, `1073741824L`) cover the real types. Just worth knowing the null row is a weaker assertion than it looks.

jongio (commented):
Noting one thing on the generic Theory approach.

jongio (commented):
The cross-instance test gap from my earlier review is still the main thing here. Two smaller items inline.

**CommandResultExtensionsTests.cs** - The `changedRequiredness` tests still use the same variable for registration and lookup - `option = option.AsRequired()` then both register and query `option`. This doesn't exercise the scenario this PR exists for: register `.AsRequired()` on the command, then look up with the original static definition.

Something like:

```csharp
var original = new Option<string?>("--name");
var command = new Command("test") { original.AsRequired() };
var parseResult = command.Parse("--name value");
Assert.Equal("value", parseResult.CommandResult.GetValueOrDefault(original));
```

Cover `GetValueOrDefault`, `TryGetValue`, `GetValueWithoutDefault`, and `HasOptionResult` with this divergent-instance pattern.

core/Microsoft.Mcp.Core/src/Extensions/CommandResultExtensions.cs
```diff
@@ -92,21 +104,22 @@ public static bool TryGetValue<T>(this CommandResult commandResult, Option<T> op
         return false;
     }
 
-    public static bool TryGetValue<T>(this CommandResult commandResult, string optionName, out T? value)
+    public static T? GetValueOrDefault<T>(this CommandResult commandResult, Option<T> option)
+        => GetValueOrDefault<T>(commandResult, option.Name);
+
```
jongio, line 113: `GetValueOrDefault<T>(Option<T>)` and `GetValueWithoutDefault<T>(Option<T>)` (line 156) previously threw `ArgumentNullException` for null options via `ThrowIfNull` in the method body. Now they dereference `option.Name` immediately, producing `NullReferenceException` instead. Low practical risk since options are static definitions, but it's a subtle API behavior change worth noting.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Extensions/CommandResultExtensionsTests.cs
```diff
@@ -285,222 +122,290 @@ public void GetValueWithoutDefault_WithMissingIntValue_ReturnsNull()
...
 
     [Fact]
-    public void GetValueWithoutDefault_WithMissingBoolValue_ReturnsNull()
+    public void GetValueWithoutDefault_WithStringOptionName_WithNonExistentOption_ReturnsNull()
     {
         // Arrange
-        var option = new Option<bool?>("--zone-redundant");
+        var option = new Option<string?>("--name");
         var command = new Command("test") { option };
         var parseResult = command.Parse("");
```
jongio, line 170: Typo: `TypeOption` should be `TypedOption` to match `TypedOptionWithVariousBoolScenarios`.

jongio (approved):
All prior feedback addressed. Cross-instance test coverage is in place across `HasOptionResult`, `TryGetValue`, `GetValueOrDefault`, and `GetValueWithoutDefault` for both typed and untyped paths. Null-arg behavior on the public `Option<T>` overloads is back to `ArgumentNullException`. Migrations are complete in `VmssUpdateCommand` and `WebTestsCreateOrUpdateCommand`. Design is sound - routing through `option.Name` cleanly fixes the `AsRequired()`/`AsOptional()` identity mismatch.

PR #2338, opened 2026-04-03
Add pre-flight version check for registry stdio servers

### LINKED ISSUE #1025: [BUG] azd min version check error could be clearer
### Describe the bug

azd version check tells user azd isn't install, but it's just not the right version

Image

### Expected behavior

tells user they need to upgrade

### Actual behavior

fails for missing azd

### Reproduction Steps

run #azd command with azmcp loaded

### Environment

_No response_

jongio:
## Problem

When `azd` is installed but at a version below the minimum (1.20.0), the error message says *"This tool may require dependencies that are not installed"* instead of telling the user their version is too old. This is confusing — the tool IS installed, just outdated.

Fixes #1025

## Solution

Add a generic pre-flight version check to `RegistryServerProvider` that runs before creating the stdio MCP client. Registry servers can now declare `minVersion` and `versionArgs` in `registry.json` to enable version checking.

**Error messages are now specific:**
- Not installed: `'azd' is not installed or not found in the system PATH.`
- Wrong version: `'azd' version 1.11.0 is installed, but version 1.20.0 or later is required.`

## Changes

| File | Change |
|------|--------|
| `RegistryServerInfo.cs` | Added `MinVersion` and `VersionArgs` properties |
| `RegistryServerProvider.cs` | Added `CheckCommandVersionAsync` with 30s timeout, concurrent stdout/stderr reading, proper process cleanup; added `ParseVersionFromOutput` with `GeneratedRegex` |
| `registry.json` | Added `minVersion: 1.20.0` and `versionArgs: ["version"]` for azd |
| `RegistryServerProviderTests.cs` | Added 7 tests covering: version parsing, command-not-found, version-too-old, version-sufficient, CreateClientAsync integration |

## Design Decisions

- **Generic, not azd-specific**: Any stdio registry server can use `minVersion`/`versionArgs`
- **Fail-open**: If version check encounters unknown errors or times out, it falls through to the normal connection attempt
- **30s hard timeout**: Prevents a hung version command from blocking indefinitely
- **Process safety**: Concurrent stdout+stderr reading (prevents pipe-buffer deadlocks), `Process.Kill(entireProcessTree: true)` on timeout, `Process.Start()` return value checked
- **AOT safe**: Uses `[GeneratedRegex]` for source-generated regex

## Testing

- 21 tests pass (14 existing + 7 new)
- 0 build warnings
- Format verified clean

## Invoking Livetests

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
@@ -17,14 +17,13 @@ namespace Microsoft.Mcp.Core.Areas.Server.Commands.Discovery;
 /// <param name="serverInfo">Configuration information for the server.</param>
 /// <param name="httpClientFactory">Factory for creating HTTP clients.</param>
 /// <param name="tokenCredentialProvider">The token credential provider for OAuth authentication.</param>
-public sealed partial class RegistryServerProvider(string id, RegistryServerInfo serverInfo, IHttpClientFactory httpClientFactory) : IMcpServerProvider
+public sealed class RegistryServerProvider(string id, RegistryServerInfo serverInfo, IHttpClientFactory httpClientFactory) : IMcpServerProvider
 {
     private readonly string _id = id;
     private readonly RegistryServerInfo _serverInfo = serverInfo;
     private readonly IHttpClientFactory _httpClientFactory = httpClientFactory;
 
-    [GeneratedRegex(@"(\d+\.\d+\.\d+)")]
-    private static partial Regex SemVerPattern();
+    private const string DefaultVersionPattern = @"(\d+\.\d+\.\d+)";
 
     /// <summary>
     /// Creates metadata that describes this registry-based server.
@@ -63,8 +62,15 @@ public sealed partial class RegistryServerProvider(string id, RegistryServerInfo
             !string.IsNullOrWhiteSpace(_serverInfo.MinVersion) &&
             !string.IsNullOrWhiteSpace(_serverInfo.Command))
         {
+            if (_serverInfo.VersionArgs == null || _serverInfo.VersionArgs.Count == 0)
+            {
+                throw new InvalidOperationException(
+                    $"Registry server '{_id}' specifies 'minVersion' but is missing 'versionArgs'. Both are required for version checking.");
+            }
+
             var versionError = await CheckCommandVersionAsync(
-                _serverInfo.Command, _serverInfo.VersionArgs, _serverInfo.MinVersion, cancellationToken);
+                _serverInfo.Command, _serverInfo.VersionArgs, _serverInfo.MinVersion,
+                _serverInfo.VersionPattern, cancellationToken);
 
             if (versionError != null)
             {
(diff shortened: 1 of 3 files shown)


core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -15,12 +17,15 @@ namespace Microsoft.Mcp.Core.Areas.Server.Commands.Discovery;
 /// <param name="serverInfo">Configuration information for the server.</param>
 /// <param name="httpClientFactory">Factory for creating HTTP clients.</param>
 /// <param name="tokenCredentialProvider">The token credential provider for OAuth authentication.</param>
-public sealed class RegistryServerProvider(string id, RegistryServerInfo serverInfo, IHttpClientFactory httpClientFactory) : IMcpServerProvider
+public sealed partial class RegistryServerProvider(string id, RegistryServerInfo serverInfo, IHttpClientFactory httpClientFactory) : IMcpServerProvider
```
alzimmermsft: Why did this need to become a `partial` class?

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -15,12 +17,15 @@ namespace Microsoft.Mcp.Core.Areas.Server.Commands.Discovery;
...
 {
     private readonly string _id = id;
     private readonly RegistryServerInfo _serverInfo = serverInfo;
     private readonly IHttpClientFactory _httpClientFactory = httpClientFactory;
 
+    [GeneratedRegex(@"(\d+\.\d+\.\d+)")]
+    private static partial Regex SemVerPattern();
```
alzimmermsft: This would be better defined as a `registry.json` property, we may need to deal with cases that don't use semver (or don't use it properly).

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -76,6 +104,107 @@ This tool may require dependencies that are not installed.
...
+            process.StartInfo = new ProcessStartInfo
+            {
+                FileName = command,
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                CreateNoWindow = true
+            };
+
+            var args = versionArgs ?? (IList<string>)["--version"];
```
alzimmermsft: Let's make it required for registry servers to include the specific arguments needed for this check. Better to fail ourselves on some improper configuration during development than for someone using the server to get a weird error message about an invalid argument to a tool.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
```diff
@@ -76,6 +104,107 @@ This tool may require dependencies that are not installed.
...
+            }
+
+            if (!process.Start())
+            {
+                return $"'{command}' failed to start.";
+            }
+
+            // Read both streams concurrently to avoid pipe-buffer deadlocks
+            var stdoutTask = process.StandardOutput.ReadToEndAsync(cancellationToken);
+            var stderrTask = process.StandardError.ReadToEndAsync(cancellationToken);
```
alzimmermsft: Should we instead start the process where it redirects error output to the standard output and we just need to read one stream.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
jongio: This was needed for `[GeneratedRegex]` source generation, but per your feedback below the version pattern now comes from registry config via `VersionPattern`, so `partial` has been removed.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
jongio: Agreed — added `VersionPattern` property to `RegistryServerInfo` (and the registry.json schema). Falls back to `(\d+\.\d+\.\d+)` when not specified.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
jongio: Agreed — `versionArgs` is now validated at runtime: if `minVersion` is set but `versionArgs` is missing, we throw `InvalidOperationException` with a clear config error. Added a test for this case.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/Discovery/RegistryServerProvider.cs
jongio: Agreed — removed `RedirectStandardError` entirely and now only read stdout. Simplified the stream handling significantly.

PR #2355, opened 2026-04-07
Added query parameterization

xiangyan99:
## What does this PR do?

Added query parameterization

`[Add additional context, screenshots, or information that helps reviewers]`

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
+++ b/core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
@@ -67,6 +67,19 @@ public static class SqlQueryParameterizer
         {
             if (query[i] == '\'')
             {
+                // Detect PostgreSQL E'...' escape string prefix
+                var isEscapeString = false;
+                if (i > 0 && (query[i - 1] == 'E' || query[i - 1] == 'e'))
+                {
+                    // Ensure E/e is a standalone prefix, not part of a longer identifier
+                    if (i - 1 == 0 || !(char.IsLetterOrDigit(query[i - 2]) || query[i - 2] == '_'))
+                    {
+                        isEscapeString = true;
+                        // Remove the E/e already appended to result
+                        result.Length--;
+                    }
+                }
+
                 var isTypedLiteral = IsTypedLiteralContext(query, i);
                 var literalStart = i;
                 var value = new StringBuilder();
@@ -74,9 +87,9 @@ public static class SqlQueryParameterizer
 
                 while (i < query.Length)
                 {
-                    if (dialect == SqlDialect.MySql && query[i] == '\\' && i + 1 < query.Length)
+                    if ((dialect == SqlDialect.MySql || isEscapeString) && query[i] == '\\' && i + 1 < query.Length)
                     {
-                        // MySQL backslash escape sequences
+                        // Backslash escape sequences (MySQL or PostgreSQL E'...')
                         var next = query[i + 1];
                         value.Append(next switch
                         {
@@ -112,6 +125,12 @@ public static class SqlQueryParameterizer
                 if (isTypedLiteral)
                 {
                     // Preserve original literal text for typed-literal contexts
+                    if (isEscapeString)
+                    {
+                        // Re-insert the stripped E/e prefix
+                        result.Append(query[literalStart - 1]);
+                    }
+
                     result.Append(query.AsSpan(literalStart, i - literalStart));
                 }
                 else
--- a/core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.UnitTests/Helpers/SqlQueryParameterizerTests.cs
+++ b/core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.UnitTests/Helpers/SqlQueryParameterizerTests.cs
@@ -176,4 +176,28 @@ public class SqlQueryParameterizerTests
         Assert.Equal("@p0", parameters[0].Name);
         Assert.Equal("Alice", parameters[0].Value);
     }
+
+    [Fact]
+    public void Parameterize_PostgresEscapeStringLiteral_StripsEPrefixAndDecodesBackslash()
+    {
+        var (query, parameters) = Parameterize(
+            "SELECT * FROM t WHERE col = E'\\n'",
+            SqlDialect.Standard);
+
+        Assert.Equal("SELECT * FROM t WHERE col = @p0", query);
+        Assert.Single(parameters);
+        Assert.Equal("\n", parameters[0].Value);  // actual newline, not backslash-n
+    }
+
+    [Fact]
+    public void Parameterize_PostgresEscapeStringLiteralLowercase_StripsEPrefixAndDecodesBackslash()
+    {
+        var (query, parameters) = Parameterize(
+            "SELECT * FROM t WHERE col = e'\\t'",
+            SqlDialect.Standard);
+
+        Assert.Equal("SELECT * FROM t WHERE col = @p0", query);
+        Assert.Single(parameters);
+        Assert.Equal("\t", parameters[0].Value);  // actual tab
+    }
 }


tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
```diff
@@ -287,6 +295,9 @@ public async Task<List<JsonElement>> QueryItems(
         return items;
     }
 
+    internal static (string Query, List<(string Name, string Value)> Parameters) ParameterizeStringLiterals(string query) =>
+        SqlQueryParameterizer.Parameterize(query);
```
alzimmermsft: Should Cosmos have its own thing that's not associated with SQL given it have its own querying language?

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// as-is because the SQL parser requires a literal token in those positions.
+    /// </summary>
+    /// <param name="query">The SQL query containing string literals.</param>
+    /// <param name="dialect">Which escape sequences to decode. Defaults to <see cref="SqlDialect.Standard"/>.</param>
+    /// <returns>
+    /// A tuple of the rewritten query and the list of extracted (name, value) pairs.
+    /// </returns>
+    public static (string Query, List<(string Name, string Value)> Parameters) Parameterize(
+        string query,
+        SqlDialect dialect = SqlDialect.Standard)
```
alzimmermsft: Let's remove this implicit param and force anyone calling into this to explicitly know what they need. This has been a bit of a problem with other areas like credentials where we have important implicit parameters that are forgotten to get passed as nothing is forcing correct behavior.

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+    private const int MaxQueryLength = 10000;
+
+    // Regex patterns for detecting boolean tautology injection.
+    // These catch patterns like: or 1==1, or 1=1, or true, or '1'=='1', etc.
+    private static readonly Regex TautologyPattern = RegexHelper.CreateRegex(
+        @"\bor\s+(\d+\s*==?\s*\d+|true|'[^']*'\s*==?\s*'[^']*')",
+        RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+    // Dangerous management commands that should never appear in user queries
+    private static readonly string[] DangerousCommands =
```
vukelich, line 28: Rather than maintain this list, can we leverage request properties to let the server-side query engine provide stronger safeguards than our bespoke code? If I'm on the right documentation, the [request properties](https://learn.microsoft.com/en-us/kusto/api/rest/request-properties) have options like `request_readonly` and `request_readonly_hardline` to prevent commands with side effects.

vukelich: There are a few that sound potentially useful.

tools/Azure.Mcp.Tools.Cosmos/src/Services/CosmosService.cs
xiangyan99: It is by design that we put some shared logic in the SqlQueryParameterizer.

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
xiangyan99: Fixed.

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// each one with a numbered parameter placeholder (@p0, @p1, …).
+    /// Literals that follow typed-literal keywords (e.g. DATE, INTERVAL) are preserved
+    /// as-is because the SQL parser requires a literal token in those positions.
+    /// </summary>
+    /// <param name="query">The SQL query containing string literals.</param>
+    /// <param name="dialect">Which escape sequences to decode. Defaults to <see cref="SqlDialect.Standard"/>.</param>
+    /// <returns>
+    /// A tuple of the rewritten query and the list of extracted (name, value) pairs.
+    /// </returns>
+    public static (string Query, List<(string Name, string Value)> Parameters) Parameterize(
```
vukelich, line 55: I'm trying to consider ways in which our functional changes to the user query can result in problems. The two off the top of my mind are:

1. Lots of string literals and potentially duplicates. Are there limits on the number of parameters to a query? Should we be de-duplicating repeated literals to avoid exploding the size of the query?
2. Are there ways in which this could make a query invalid? For example, are there places that string literals cannot be replaced by query parameters? What if a single quote is within a double quote? Is it valid to have a query that says `SELECT c["it's a property"] FROM c`?

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
xiangyan99: I will investigate later. It is out of the scope of "security fix for 2.0". :)

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
xiangyan99: Added the test to cover SELECT c["it's a property"] FROM c

xiangyan99: Regarding #1, will revisit what is the best approach to handle kql post 2.0.

conniey: Is it possible to use https://www.nuget.org/packages/Microsoft.Azure.Kusto.Language? this package seems like it does not look like it will break AOT compatibility and saves us from having to parse programmatically.

xiangyan99: Yes. This is definitely something we can investigate post 2.0.

conniey: ~~I think it's worth investigating now since we've had three back and forths already.~~

Mmm. Never mind.

tools/Azure.Mcp.Tools.Kusto/src/Services/KustoService.cs
```diff
@@ -34,6 +35,10 @@ public sealed class KustoService(
     /// <summary>
     /// Escapes a KQL identifier (e.g., table name) using bracket notation to prevent injection.
     /// </summary>
+    /// <summary>
+    /// Escapes a KQL identifier (e.g., table name) using bracket notation to prevent injection.
```
g2vinay: This xml block is a duplicate of above. 
We should remove the older one.

tools/Azure.Mcp.Tools.MySql/src/Services/MySqlService.cs
```diff
@@ -231,6 +231,9 @@ internal static void ValidateQuerySafety(string query)
         }
     }
```
g2vinay, line 232: This PR doesn't cover Azure.Mcp.Tools.Sql, is that intentional ?

xiangyan99: I don't see that we expose the query command for tools.sql.

g2vinay: Okay so no op there then.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.UnitTests/Helpers/SqlQueryParameterizerTests.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        var (query, parameters) = Parameterize(
+            "SELECT * FROM c WHERE c.name = 'Alice'",
+            SqlDialect.Standard);
+
+        Assert.Equal("SELECT * FROM c WHERE c.name = @p0", query);
+        Assert.Single(parameters);
+        Assert.Equal("@p0", parameters[0].Name);
+        Assert.Equal("Alice", parameters[0].Value);
+    }
```
g2vinay, line 178: Can we add these two edge cases as tests too ?
I suspect these might fail in today's paramterize logic.

```c#
  [Fact]
    public void Parameterize_PostgresEscapeStringLiteral_StripsEPrefixAndDecodesBackslash()
    {
        var (query, parameters) = Parameterize(
            "SELECT * FROM t WHERE col = E'\\n'",
            SqlDialect.Standard);

        Assert.Equal("SELECT * FROM t WHERE col = @p0", query);
        Assert.Single(parameters);
        Assert.Equal("\n", parameters[0].Value);  // actual newline, not backslash-n
    }

    [Fact]
    public void Parameterize_PostgresEscapeStringLiteralLowercase_StripsEPrefixAndDecodesBackslash()
    {
        var (query, parameters) = Parameterize(
            "SELECT * FROM t WHERE col = e'\\t'",
            SqlDialect.Standard);

        Assert.Equal("SELECT * FROM t WHERE col = @p0", query);
        Assert.Single(parameters);
        Assert.Equal("\t", parameters[0].Value);  // actual tab
    }

```

xiangyan99: Good catch! Added.

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
vukelich: Why is using `request_readonly` and `request_readonly_hardline` out of scope for 2.0? I'm not sure I follow.

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
```diff
@@ -0,0 +1,296 @@
...
+                        i++;
+                        break;
+                    }
+                    else
+                    {
+                        value.Append(query[i]);
+                        i++;
+                    }
+                }
+
```
jongio, line 124: When the inner `while` loop exits because `i >= query.Length` (unterminated string literal), execution falls through to the parameterization block silently. The result is a query with a dangling `@pN` placeholder whose value is whatever partial content was collected before EOF.

For a security-sensitive parser, silently consuming malformed input is risky - it could mask injection attempts or produce queries that fail in confusing ways downstream. Consider throwing an `ArgumentException` when the loop exits without finding a closing quote, similar to how `KqlQueryValidator` rejects malformed input.

core/Microsoft.Mcp.Core/src/Helpers/SqlQueryParameterizer.cs
```diff
@@ -0,0 +1,296 @@
...
+                        // Re-insert the stripped E/e prefix
+                        result.Append(query[literalStart - 1]);
+                    }
+
+                    result.Append(query.AsSpan(literalStart, i - literalStart));
+                }
+                else
+                {
+                    var paramName = $"@p{paramIndex++}";
+                    parameters.Add((paramName, value.ToString()));
```
jongio, line 139: The generated names `@p0`, `@p1`, ... will collide with pre-existing parameter references in the query. If someone passes `SELECT * FROM t WHERE x = @p0 AND y = 'foo'`, the output would contain two occurrences of `@p0` with different intended values.

Consider either scanning the query for existing `@pN` references and starting `paramIndex` above the max, or using a prefix that is unlikely to collide (e.g., `@__p0`).

core/Microsoft.Mcp.Core/src/Helpers/KqlSanitizer.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(identifier);
+        return $"['{identifier.Replace("'", "''", StringComparison.Ordinal)}']";
+    }
+
+    /// <summary>
+    /// Sanitizes KQL string literals by parsing each single-quoted literal,
+    /// normalizing doubled-quote escapes, and re-encoding with proper escaping.
+    /// This prevents injection through string literal breakout where a crafted
+    /// literal value could escape the quote context and inject KQL operators.
```
jongio, line 41: The xmldoc claims this method "prevents injection through string literal breakout", but re-parsing well-formed string literals and re-emitting them with doubled-quote escaping is a no-op for any query that is already syntactically valid. The actual injection defense is `KqlQueryValidator.ValidateQuerySafety` (which blocks tautologies, management commands, etc.).

Consider updating the doc to describe what this actually does: normalizes string literal escaping as a defense-in-depth measure. The current wording could give callers a false sense of security if they use `SanitizeStringLiterals` without `ValidateQuerySafety`.

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+/// </summary>
+internal static class KqlQueryValidator
+{
+    private const int MaxQueryLength = 10000;
+
+    // Regex patterns for detecting boolean tautology injection.
+    // These catch patterns like: or 1==1, or 1=1, or true, or '1'=='1', etc.
+    private static readonly Regex TautologyPattern = RegexHelper.CreateRegex(
+        @"\bor\s+(\d+\s*==?\s*\d+|true|'[^']*'\s*==?\s*'[^']*')",
+        RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
jongio, line 25: The `\d+` pattern matches integer tautologies (`1==1`) but misses float variants like `1.0==1.0` or `0.1==0.1`. An attacker who knows the validator exists could trivially bypass with decimal comparisons.

Consider `\d+\.?\d*` or similar to cover decimal numbers.

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+                $"Query length exceeds the maximum allowed limit of {MaxQueryLength:N0} characters.",
+                HttpStatusCode.BadRequest);
+        }
+
+        // Strip string literals before structural analysis to avoid false positives
+        var queryWithoutStrings = Regex.Replace(query, "'([^']|'')*'", "'str'", RegexOptions.None, RegexHelper.DefaultRegexTimeout);
+
+        // Detect KQL comments (// line comments)
+        if (queryWithoutStrings.Contains("//", StringComparison.Ordinal))
+        {
```
jongio, line 66: This checks for `//` line comments but KQL also supports `/* */` block comments. An attacker could use `/* comment */ .drop table T` to hide intent from human reviewers while the block comment is stripped by the Kusto engine before execution.

Consider also checking for `/*` in the string-stripped query.

tools/Azure.Mcp.Tools.Kusto/src/Validation/KqlQueryValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+                HttpStatusCode.BadRequest);
+        }
+
+        // Detect management/control commands
+        var lower = queryWithoutStrings.Trim().ToLowerInvariant();
+        foreach (var cmd in DangerousCommands)
+        {
+            if (lower.StartsWith(cmd, StringComparison.Ordinal) ||
+                lower.Contains($"| {cmd}", StringComparison.Ordinal) ||
+                lower.Contains($";{cmd}", StringComparison.Ordinal) ||
```
jongio, line 86: The pipe-command detection checks for `"| {cmd}"` (with a space), but KQL does not require whitespace between the pipe and the command name. `|.drop` is valid KQL and would bypass this check.

Consider matching `|` followed by optional whitespace: checking for both `"| {cmd}"` and `"|{cmd}"` patterns, or stripping whitespace after pipes before checking.

PR #2357, opened 2026-04-07
update to include query suites

g2vinay:


jongio (commented):
Heads up: this looks superseded by #2358, which already merged on 2026-04-07 and configured the same three suites using the singular CSV form:

```yaml
codeql:
  querySuite: 'sdl-required, active-ssirp, secure-future-initiative'
  compiled:
    enabled: true
```

This PR branched before that landed, so a 3-way merge would leave the `codeql:` block with both `querySuite` (singular string) and `querySuites` (plural list) side by side. That's ambiguous at best, and one of them will silently lose depending on how the 1ES template parser handles unknown/duplicate keys.

Two other things worth double-checking before reviving this:

1. The 1ES Pipeline Templates CodeQL schema field is `querySuite` (singular, CSV string), which is what #2358 adopted. I can't find `querySuites` (plural, list) in the documented schema. `Build Analyze` passing here doesn't prove it's wired up - unknown fields are typically accepted and ignored silently, so this change could be a no-op.
2. The PR body is empty, so the intent isn't clear. If the goal is readability (list over CSV), a follow-up that *replaces* the existing `querySuite:` line instead of adding a second key would be the cleaner path.

If #2358 already covers the intent, this one can probably be closed.

PR #2380, opened 2026-04-09
Fix the New Build Info Script to handle edge case release versions

g2vinay:
Updates the New Build Info script to fetch all released versions and not just latest.
In the event we ship next beta series before GA'ing the current beta train, the current logic breaks.

### DIFF
--- a/eng/scripts/New-BuildInfo.ps1
+++ b/eng/scripts/New-BuildInfo.ps1
@@ -117,9 +117,9 @@ function Get-LatestMarketplaceVersion {
                     )
                 }
             )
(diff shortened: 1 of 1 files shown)


anannya03 (approved):

KarishmaGhiya (approved):

jongio (approved):
Flag arithmetic verified - the old value (914) had IncludeLatestVersionOnly (0x200) set, which was the root cause. New value (387) correctly uses IncludeVersions (0x1) instead. Dropping IncludeVersionProperties (0x10) is fine since the code only reads version strings. Clean fix.

PR #2391, opened 2026-04-13
Fix subscription ID tests when logged in with Azure CLI

alzimmermsft:
## What does this PR do?

Fixes testing pain point where some tests fail if you're signed into Azure CLI due to `subscription` falling back to the logged in Azure CLI subscription if it's available.

Tests targeting `subscription` testing now use the Azure CLI subscription if it's set, otherwise fallback to previous handling with using the injected / omitted `AZURE_SUBSCRIPTION_ID` environment variable.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Helpers/CommandHelperTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Helpers/CommandHelperTests.cs
@@ -13,28 +13,28 @@ public class CommandHelperTests
     public void GetSubscription_EmptySubscriptionParameter_ReturnsEnvironmentValue()
     {
         // Arrange
-        EnvironmentHelpers.SetAzureSubscriptionId("env-subs");
+        var subscription = EnvironmentHelpers.SetAzureSubscriptionId("env-subs");
         var parseResult = GetParseResult(["--subscription", ""]);
 
         // Act
         var actual = CommandHelper.GetSubscription(parseResult);
 
         // Assert
-        Assert.Equal("env-subs", actual);
+        Assert.Equal(subscription, actual);
     }
 
     [Fact]
     public void GetSubscription_MissingSubscriptionParameter_ReturnsEnvironmentValue()
     {
         // Arrange
-        EnvironmentHelpers.SetAzureSubscriptionId("env-subs");
+        var subscription = EnvironmentHelpers.SetAzureSubscriptionId("env-subs");
         var parseResult = GetParseResult([]);
 
         // Act
         var actual = CommandHelper.GetSubscription(parseResult);
 
         // Assert
-        Assert.Equal("env-subs", actual);
+        Assert.Equal(subscription, actual);
     }
 
     [Fact]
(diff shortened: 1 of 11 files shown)


core/Microsoft.Mcp.Core/src/Helpers/EnvironmentVariableHelpers.cs
```diff
@@ -8,36 +8,47 @@ public static class EnvironmentHelpers
...
     {
+        var currentSubscription = CommandHelper.GetProfileSubscription();
+        if (!string.IsNullOrEmpty(currentSubscription))
+        {
+            return currentSubscription;
+        }
+
         Environment.SetEnvironmentVariable(AzureSubscriptionIdEnvironmentVariable, subscriptionId);
+        return subscriptionId;
     }
```
jongio: A method named `Set` that conditionally doesn't set is surprising. When CLI is logged in, this returns the CLI subscription without writing the env var at all - so `GetAzureSubscriptionId()` after this call still returns the old value. The Set/Get contract is broken.

Consider keeping this as a pure setter and moving the `resolve effective subscription` logic to a separate test helper or using the existing `CommandHelper.GetDefaultSubscription()` in test assertions.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Helpers/CommandHelperTests.cs
```diff
@@ -55,54 +55,56 @@ public void GetSubscription_ValidSubscriptionParameter_ReturnsParameterValue()
...
         // Arrange
+        var subscription = CommandHelper.GetProfileSubscription();
         var parseResult = GetParseResult(["--subscription", "Some default name"]);
 
         // Act
         var actual = CommandHelper.GetSubscription(parseResult);
 
         // Assert
-        Assert.Equal("Some default name", actual);
+        Assert.Equal(subscription ?? "Some default name", actual);
```
jongio: This assertion silently tests different behavior depending on machine state. When CLI is logged in: `subscription` is the CLI profile ID, and the `??` never fires - you're comparing two values from the same Lazy cache. When CLI isn't logged in: `subscription` is null and you compare `Some default name` to `Some default name`.

These are entirely different behaviors hidden behind `??`. The test name says `ReturnsParameterValue` but when CLI is logged in, the parameter value isn't returned at all.

jongio (changes requested):
CI is red - the signature change on `SetAzureSubscriptionId` breaks `RegistryListCommandTests.cs:57` which passes `null`. Beyond that compile fix, I think the approach here has a deeper issue worth discussing.

The core problem: `SetAzureSubscriptionId` now conditionally doesn't set anything and returns a value unrelated to its input. On a CLI-logged-in machine, calling `Set("env-subs")` silently no-ops and returns the CLI profile subscription. Set/Get isn't round-trippable, and callers can't tell whether their value was actually written. That's a correctness hazard for any future caller.

The test assertions also become tautological when CLI is logged in - both sides of `Assert.Equal(subscription, actual)` resolve through the same `s_profileDefault.Value` cache, so the tests prove cache consistency rather than subscription resolution behavior.

A simpler alternative: keep `SetAzureSubscriptionId` as a pure setter (always writes the env var), and have tests derive the expected value via `CommandHelper.GetDefaultSubscription()`. That way the env var IS written (no hidden no-op), the resolution precedence is tested for real, and the assertions are meaningful regardless of whether CLI is logged in.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Helpers/CommandHelperTests.cs
alzimmermsft: Yeah, this is a bit confusing but so is the logic of the running code in this case. For a long time, the only fallback for `--subscription` not being passed was having the `AZURE_SUBSCRIPTION_ID` environment configuration set, but then support for Azure CLI being logged in was added which gets checked before `AZURE_SUBSCRIPTION_ID` (and this is the correct behavior).

This makes any tests using `EnvironmentHelper.SetAzureSubscriptionId` liable to fail if the running environment ran `az login` at any point and that log in is still considered valid.

jongio (approved):
The iteration looks good - all my previous concerns are addressed. Moving `SetAzureSubscriptionId` to `TestEnvironment` as a pure setter and using `GetDefaultSubscription()` / `GetProfileSubscription()` to derive expected values is the right approach. Clean separation between production and test code.

One small nit for a future pass: the `ContainerAppListCommandTests.cs` finally block could add an `else { TestEnvironment.ClearAzureSubscriptionId(); }` for defensive cleanup when the original value was null. Not blocking - the env var is already cleared from the try block.

PR #2396, opened 2026-04-14
[DO NOT MERGE} Test - PR

g2vinay:


jongio (commented):
Script logic looks sound. The new branch matches the existing $vsixVersion / $vsixIsPrerelease assignment pattern above, and scoping the exception to exactly 1.0.0 (stable, no prerelease) is a safe default. A couple of things to sort before this can merge:

1. Merge conflicts against main need resolving (mergeStateStatus is DIRTY).
2. The .csproj bump to 1.0.0 is a real release change sitting inside a PR tagged "DO NOT MERGE - Test". If the bump isn't meant to ship as part of this PR, split it out; if it is, update the title to reflect that.

One thing to flag for later: the guard is major/minor/patch-literal on 1.0.0. If a future 2.0.0 first-GA lands with no marketplace history for 2.0.X, the script will error. The inline comment acknowledges this as intentional, so it's fine as-is, but worth noting if the team later wants to generalize to any X.0.0 stable first-publish.

PR #2399, opened 2026-04-14
Add base class for command unit tests

alzimmermsft:
## What does this PR do?

This PR adds a base class for command unit tests to provide shared functionality that can be used by all command unit tests. The base class binds up a `ServiceCollection` and `ServiceProvider` to provide a richer testing experience that more closely aligns with how production runtime works. This leverages dependency injection used in runtime and should allow for future changes to the runtime to more easily integrate into all unit tests by providing a central location for configuration.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:


core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
```diff
@@ -0,0 +1,59 @@
...
+using Xunit;
+
+namespace Microsoft.Mcp.Tests.Client;
+
+/// <summary>
+/// Base class for command unit tests, providing common setup for testing command execution and validation logic.
+/// </summary>
+/// <typeparam name="ToolCommand">The tool command.</typeparam>
+/// <typeparam name="ToolService">The tool service.</typeparam>
+public abstract class CommandUnitTestsBase<ToolCommand, ToolService>
```
jongio: `ServiceProvider` implements `IDisposable`, and xUnit creates a new class instance per test, so each test leaks one. This codebase already uses the `IDisposable` pattern on test base classes (e.g., `ToolLoaderTelemetryTests`, `NamespaceToolLoaderTests`). Consider making this class implement `IDisposable` to dispose `ServiceProvider`.

jongio (commented):
Good refactoring - the DI-based base class is a solid improvement that eliminates a lot of boilerplate and better mirrors production runtime. A couple things to consider below.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
jongio: Nit: `BaseCommand<TOptions>` uses the standard `T` prefix for type parameters. For consistency, consider renaming `ToolCommand`/`ToolService` to `TCommand`/`TService`.

AGENTS.md
```diff
@@ -310,6 +311,10 @@ All commands must include comprehensive unit tests:
 [Fact] public void BindOptions_BindsOptionsCorrectly()
 ```
 
+Command unit tests should extend `CommandUnitTestsBase<ToolCommand, ToolService>` where
```
jongio: Typo: `being test` - should be `being tested`

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Group/UnitTests/ResourceGroupListCommandTests.cs
```diff
@@ -1,48 +1,22 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.CommandLine;
 using System.Net;
 using System.Text.Json;
```
jongio: This file still uses the manual `CommandDefinition.Parse` + `Command.ExecuteAsync` pattern in all tests. The other migrated files (e.g., `ResourceListCommandTests`, `AccountGetCommandTests`) use `ExecuteCommandAsync`. Worth making consistent?

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        Service = Substitute.For<TService>();
+        Logger = Substitute.For<ILogger<TCommand>>();
+        var serviceCollection = new ServiceCollection()
+            .AddSingleton(Logger)
+            .AddSingleton(Service)
+            .AddSingleton<TCommand>();
+        extensions?.Invoke(serviceCollection);
+        ServiceProvider = serviceCollection.BuildServiceProvider();
+        Command = ServiceProvider.GetRequiredService<TCommand>();
```
conniey, line 47: Can we assume that all commands will only need a logger and service as a dependency? Seems very restrictive. What if we also wanted to collect some telemetry in this command

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    protected Task<CommandResponse> ExecuteCommandAsync(string args)
+        => Command.ExecuteAsync(Context, CommandDefinition.Parse(args), TestContext.Current.CancellationToken);
+
+    protected T? ConvertResponse<T>(CommandResponse response, JsonTypeInfo<T> jsonTypeInfo)
+        => JsonSerializer.Deserialize(JsonSerializer.Serialize(response.Results), jsonTypeInfo);
+
+    public void Dispose()
+    {
+        GC.SuppressFinalize(this);
```
conniey, line 64: I would put this as the end after we ensure disposals are called...  But this doesn't implement the dispose pattern I am used to seeing.  Maybe it's outdated? https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
```diff
@@ -0,0 +1,65 @@
...
+        CommandDefinition = Command.GetCommand();
+    }
+
+    protected Task<CommandResponse> ExecuteCommandAsync(params string[] args)
+        => Command.ExecuteAsync(Context, CommandDefinition.Parse(args), TestContext.Current.CancellationToken);
+
+    protected Task<CommandResponse> ExecuteCommandAsync(string args)
+        => Command.ExecuteAsync(Context, CommandDefinition.Parse(args), TestContext.Current.CancellationToken);
+
+    protected T? ConvertResponse<T>(CommandResponse response, JsonTypeInfo<T> jsonTypeInfo)
```
conniey: Nit naming to use Deserialize rather than Convert

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
alzimmermsft: Nope, can't make that assumption. But so far for the tools I've migrated tests for that's all that is needed.

It's also why the constructor takes an action callback for injecting more dependencies.

In the future we can keep adding into this to support additional common scenarios such as telemetry testing, etc.

jongio (commented):
No additional issues beyond what's already been flagged. The DI-based base class is clean - the `extensions` callback provides a good extensibility point, and the migration preserves test behavior correctly across all 15 migrated files. Ready to merge once existing nits are addressed.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
```
conniey, line 1: Although this class does simplify the initialization of field variables, I don't feel like it has helped simplify how we write tests.  The lines have changed because we now reference a Property instead of a member variable. 

The initialization isn't complex, we aren't performing any special configuration or computation that is inherent to `Command` classes.  We don't need a ServiceProvider either to write these tests.

alzimmermsft: I think it's okay for this initial round of work to not change how we test much and it only providing minimal improvement over each test managing initialization. This is more for unifying our testing so it's easier to make more complex in the future.

Ex, every test owning its own bespoke initialization means if we want to add some generalized validation on telemetry for all tests, we'd need to update 100+ test files. With this base class we could handle all that in one location.

Or a more in the near future example, the work to dependency inject the service into tool commands rather than them using `CommandContext` to retrieve those from the `ServiceProvider` requires updates to every test using that tool command class. This removes the need to do that.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/CommandUnitTestsBase.cs
jongio: Adding some specifics on what's still missing after the IDisposable update:

- `Dispose(bool disposing)` is currently `public`, but the pattern you linked wants it `protected virtual`. Exposing it publicly lets callers invoke the "core" dispose directly and skip `GC.SuppressFinalize`, defeating the point of having the two-method split.
- The `disposing` parameter is never read. Per the pattern, managed resources like `ServiceProvider` should only be cleaned up when `disposing == true` (managed state is unsafe to touch from a finalizer).

Since this class has no unmanaged resources and no finalizer, the full pattern is arguably overkill here. Two options that would both be clean:

1. Simplify to a single `Dispose()`:
```csharp
public virtual void Dispose()
{
    if (_disposed) return;
    ServiceProvider.Dispose();
    _disposed = true;
    GC.SuppressFinalize(this);
}
```
2. Keep the two-method form but make it standards-compliant:
```csharp
protected virtual void Dispose(bool disposing)
{
    if (_disposed) return;
    if (disposing) ServiceProvider.Dispose();
    _disposed = true;
}
```

jongio (approved):
Approving. The DI-based base class is clean, the `extensions` callback handles the extensibility conniey raised, and the migration preserves test behavior across all 15 files. The dispose-pattern thread is a nit - there's no finalizer and no unmanaged resources, so the non-standard shape doesn't actually cause a bug. Worth tidying up as follow-up, but not blocking.

PR #2116, opened 2026-03-19
Add Azure Backup MCP ToolSet - Drop1

shrja-ms:
## Summary

Adds the **Azure Backup** MCP toolset with 15 commands covering Recovery Services vaults, Data Protection backup vaults, backup policies, protected items, jobs, recovery points, governance, and disaster recovery operations.

### New Toolset: `Azure.Mcp.Tools.AzureBackup`

**15 commands across 7 command groups:**

| Group | Command | Description |
|-------|---------|-------------|
| Vault | `azurebackup_vault_get` | List/get Recovery Services and Backup Vaults |
| Vault | `azurebackup_vault_create` | Create a Recovery Services vault |
| Vault | `azurebackup_vault_update` | Update vault properties (redundancy, soft delete, immutability) |
| Policy | `azurebackup_policy_get` | List/get backup policies for a vault |
| Policy | `azurebackup_policy_create` | Create backup policies for VMs, SQL, SAP HANA, Blobs, Disks, PostgreSQL |
| Protected Item | `azurebackup_protecteditem_get` | List/get protected (backed up) items |
| Protected Item | `azurebackup_protecteditem_protect` | Enable backup protection for a resource |
| Protectable Item | `azurebackup_protectableitem_list` | List resources available for backup |
| Backup | `azurebackup_backup_status` | Check backup status for an Azure resource |
| Job | `azurebackup_job_get` | List/get backup and restore jobs |
| Recovery Point | `azurebackup_recoverypoint_get` | List/get recovery points for a protected item |
| Governance | `azurebackup_governance_find-unprotected` | Find unprotected Azure resources |
| Governance | `azurebackup_governance_immutability` | Check vault immutability configuration |
| Governance | `azurebackup_governance_soft-delete` | Check vault soft delete configuration |
| DR | `azurebackup_dr_enablecrr` | Enable cross-region restore on a vault |

### Architecture

- **Dual-stack support**: Both Recovery Services (RSV) and Data Protection (DPP) vault types
- **Datasource registry pattern**: Extensible `RsvDatasourceRegistry` and `DppDatasourceRegistry` for workload-specific operations
- **Service layer**: `AzureBackupService` orchestrates RSV and DPP operations through `RsvBackupOperations` and `DppBackupOperations`
- **AOT-safe**: All models registered in `AzureBackupJsonContext`

### Test Coverage

- **99 unit tests** covering all 15 commands (success, error, input validation, deserialization)
- **6 live integration tests** (vault get/list, policy list, governance checks, job list)
[...]

PR #2417, opened 2026-04-16
Bump Azure.Core, Azure.ResourceManager, System.ClientModel for Azure Backup SDK compatibility

shrja-ms:
## Summary

Bumps three centrally-managed packages that are **required transitive dependencies** for the Azure Backup SDK packages used in #2116:

| Package | Old Version | New Version | Reason |
|---------|-------------|-------------|--------|
| Azure.Core | 1.50.0 | 1.51.1 | Required by RecoveryServicesBackup 1.3.1, RecoveryServices 1.2.0, DataProtectionBackup 1.7.1 |
| Azure.ResourceManager | 1.13.2 | 1.14.0 | Required by RecoveryServicesBackup 1.3.1, RecoveryServices 1.2.0, DataProtectionBackup 1.7.1 |
| System.ClientModel | 1.8.0 | 1.9.0 | Transitive dependency of the above |

All three are **stable GA minor/patch bumps** with backward compatibility guarantees.

## Why a separate PR?

Per review feedback on #2116, these core package bumps affect all toolsets in the repo, not just Azure Backup. Splitting them out allows independent validation across all toolsets before the Azure Backup toolset lands.

## Validation

- Full server build (\dotnet build servers/Azure.Mcp.Server/src\) passes with 0 warnings, 0 errors.

## Invoking Livetests

Copilot submitted PRs are not trustworthy by default. Users with \write\ access to the repo need to validate the contents of this PR before leaving a comment with the text \/azp run mcp - pullrequest - live\. This will trigger the necessary livetest workflows to complete required validation.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,18 +22,6 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902</WarningsNotAsErrors>
   </PropertyGroup>
 
-  <!--
-    Suppress known vulnerability advisories for System.Security.Cryptography.Xml where no patched
-    version is available for net10.0. All 10.x versions are affected. Remove these suppressions
-    once a patched version is released.
-    See: https://github.com/advisories/GHSA-37gx-xxp4-5rgx
-    See: https://github.com/advisories/GHSA-w3x6-4m5h-cxqf
-  -->
-  <ItemGroup>
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-37gx-xxp4-5rgx" />
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-w3x6-4m5h-cxqf" />
-  </ItemGroup>
-
   <PropertyGroup Condition="'$(BuildNative)' == 'true'">
     <DefineConstants>$(DefineConstants);BUILD_NATIVE</DefineConstants>
   </PropertyGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,6 +22,18 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902</WarningsNotAsErrors>
   </PropertyGroup>
 
+  <!--
+    Suppress known vulnerability advisories for System.Security.Cryptography.Xml where no patched
+    version is available for net10.0. All 10.x versions are affected. Remove these suppressions
+    once a patched version is released.
+    See: https://github.com/advisories/GHSA-37gx-xxp4-5rgx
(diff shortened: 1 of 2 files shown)


jongio (approved):
Clean, focused dependency bump. Three stable GA minor/patch version bumps (Azure.Core 1.50.0->1.51.1, Azure.ResourceManager 1.13.2->1.14.0, System.ClientModel 1.8.0->1.9.0) for Azure Backup SDK compatibility. CI passes across all 12 build targets.

The NuGetAuditSuppress entries for GHSA-37gx-xxp4-5rgx and GHSA-w3x6-4m5h-cxqf are well-documented with removal conditions. Patched versions (10.0.6) may now exist upstream for System.Security.Cryptography.Xml, but since it's a transitive dependency locked by SDK 10.0.103, the suppressions are still the right call until the SDK ships with the fix.

Good call splitting this from #2116 for independent validation.

Directory.Build.props
```diff
@@ -22,6 +22,18 @@
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902</WarningsNotAsErrors>
   </PropertyGroup>
 
+  <!--
+    Suppress known vulnerability advisories for System.Security.Cryptography.Xml where no patched
+    version is available for net10.0. All 10.x versions are affected. Remove these suppressions
```
jongio: The comment here says "no patched version is available for net10.0" and "All 10.x versions are affected" - both are inaccurate after the merge with upstream. GHSA-37gx-xxp4-5rgx and GHSA-w3x6-4m5h-cxqf both list `10.0.6` as the patched version (affected range is `>=10.0.0, <=10.0.5`), and this PR's merge commit (`d86cece`) pins `System.Security.Cryptography.Xml` to `10.0.6` in `Directory.Packages.props` plus adds an explicit `<PackageReference>` to it in `Microsoft.Mcp.Core.csproj`.

Two paths forward:

1. Drop both suppressions and this comment block. Since the package is pinned to the patched version, the audit won't flag it, and keeping the suppressions would mask the signal if a future transitive path ever regressed to an affected version.
2. Keep them as belt-and-suspenders but rewrite the comment to be accurate: "Pinned to 10.0.6 (patched). Suppressions retained as defense-in-depth against transitive regressions."

I'd lean toward option 1.

jongio (approved):
Re-approving after the merge with main. One non-blocking follow-up below on the `NuGetAuditSuppress` block - after the merge it's pointing at the patched version and the justification comment is no longer accurate. Happy to have this addressed here or in a follow-up. CI still in progress; assuming the build targets stay green.

anannya03 (approved):

PR #2421, opened 2026-04-16
Fix Az Compute Tools Update issue.

g2vinay:
The `azmcp compute vmss update` command returned an HTTP 400 error when invoked with only `--capacity`, `--overprovision`, or `--enable-auto-os-upgrade` as the sole update parameter. This happened because the three option definitions were typed as non-nullable (`Option<int>`, `Option<bool>`) instead of nullable (`Option<int?>`, `Option<bool?>`).

When a non-nullable option is not supplied, the System.CommandLine parser resolves it to its default value (`0` for `int`, `false` for `bool`) rather than `null`. The command's "at least one update property required" validation then evaluated these default values as if the user had passed them, but the downstream service call sent `0` or `false` to ARM — causing a 400 from the API in some cases, and silently no-op'ing the update in others.

*Fix**
Changed the three affected option definitions in `ComputeOptionDefinitions.cs` to use nullable types:

### DIFF
--- a/servers/Azure.Mcp.Server/changelog-entries/1776357848029.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1776357848029.yaml
@@ -1,0 +1,3 @@
+changes:
+  - section: "Bugs Fixed"
+    description: "Fixed compute_vmss_update tool returning HTTP 400 when only --capacity, --overprovision, or --enable-auto-os-upgrade is provided by correcting option type definitions to use nullable types"
--- a/tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
+++ b/tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
@@ -307,7 +307,7 @@ public static class ComputeOptionDefinitions
 
     public const string CapacityName = "capacity";
 
-    public static readonly Option<int> Capacity = new($"--{CapacityName}")
+    public static readonly Option<int?> Capacity = new($"--{CapacityName}")
     {
         Description = "Number of VM instances (capacity) in the scale set",
         Required = false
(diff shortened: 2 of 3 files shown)


alzimmermsft (approved):
Would another fix be in `Validators` to use `HasOptions` instead of trying to get it or the default?

alzimmermsft:
May also conflict with my PR https://github.com/microsoft/mcp/pull/2261

But this can merge first and I'll just rebase later

jongio (commented):
Fix is correct and the root cause is well-identified. With non-nullable `Option<bool>` / `Option<int>`, the parser produced `default(T)` (false/0) when the flag was omitted, so the validator's `.HasValue` checks saw non-null values and the service sent `Overprovision=false` to Azure on updates the user didn't intend to touch. Switching to `Option<bool?>` / `Option<int?>` lines the option declarations up with the binding code (`GetValueOrDefault<bool?>`, `GetValueOrDefault<int?>`) and the `VmssUpdateOptions` properties, which were already nullable.

One follow-up worth a separate PR: the same type mismatch exists on the create side.

- `ComputeOptionDefinitions.InstanceCount` is `Option<int>` but `VmssCreateCommand.BindOptions` reads it as `GetValueOrDefault<int?>`. When `--instance-count` is omitted, the bound value is `0`, not `null`, so `effectiveInstanceCount = instanceCount ?? 2` in `ComputeService` resolves to `0` and the VMSS is created with capacity 0.
- Same pattern for `OsDiskSizeGb` (`Option<int>`) consumed as `int?` by both `VmCreateCommand` and `VmssCreateCommand`. The service guards `DiskSizeGB = effectiveOsDiskSizeGb` on the assumption null means "let Azure pick", but it currently can't receive null.

No action needed here, this PR's scope is the reported update bug, but flagging for a follow-up since it's the same class of defect.

Issue #201, opened 2025-09-03
Allow Resource Group Option to be reused in child commands.


alzimmermsft:
Right now we have resource group option in MonitorOptionDefinitions as OptionalResourceGroup, and it is also in OptionDefinitions.  The reason they did that in MonitorOptionsDefinitions is because it isn't required.  Setting IsRequired to false broke other scenarios.  It is a static variable so shared amongst many commands, which breaks tests.  So we need a design that will allow us to define parameters once and have commands register them in RegisterOptions and also indicate if the option is required or not for their particular command without impacting other commands or tests.  Build something that is idiomatic to C#.

Update CHANGELOG.md with your fixes.  Make sure `dotnet test` from the root passes.  Run `dotnet format` before committing.

alzimmermsft:
Copied from https://github.com/Azure/azure-mcp/issues/617

joshfree:
@jongio is already working on some refactoring here

jongio:
#452

jongio:
Fixed with #452

PR #2428, opened 2026-04-16
Fix VS Code links in changelog

anannya03:
## What does this PR do?
This PR fixes incorrect VS Code links in the changelog to ensure they point to the correct locations.

### DIFF
--- a/servers/Azure.Mcp.Server/CHANGELOG.md
+++ b/servers/Azure.Mcp.Server/CHANGELOG.md
@@ -16,7 +16,7 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Bugs Fixed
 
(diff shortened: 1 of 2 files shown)


jongio (commented):
Anchor targets resolve against actual headings in this file (`## 2.0.1 (2025-10-29) (pre-release)` and `## 2.0.40 (2026-04-07) (pre-release)`), so the link fix is correct.

Worth reinforcing the bot's observation about display text: the visible link text still reads `2.0.0-beta.1` / `2.0.0-beta.40`, but neither string is a heading in this file. A click on `2.0.0-beta.1` lands on `## 2.0.1` and that mapping isn't documented anywhere in the page, which is going to confuse readers scanning the release history.

Heads up: `main` carries the same display-text mismatch. The anchors there already point at `#201-...` and `#2040-...` but the link text still reads `2.0.0-beta.1` / `2.0.0-beta.40`. If you adopt the bot's suggestion here, worth porting the same change to `main` in a follow-up so the two branches don't diverge again.

PR #2430, opened 2026-04-17
Run Update-Solution script in Build stage

anannya03:
## What does this PR do?

Adds a -Verify flag to Update-Solution.ps1 that checks whether .slnx solution files are up to date without overwriting them. Integrates this verification into Analyze-Code.ps1 so the CI pipeline catches stale .slnx files before merge.

### Changes 

`Update-Solution.ps1`
- Added `-Verify` switch parameter that generates solution files to temp paths and compares against the committed versions
- Tracks stale files via `$script:stale` and `$script:staleFiles` and prints a summary with remediation instructions on failure
- Added `$PSNativeCommandUseErrorActionPreference = $true` so dotnet failures surface as terminating errors automatically
- Temp files are cleaned up in finally blocks to avoid leaving artifacts on failure

`Analyze-Code.ps1`
- Added a "solution files up to date" check that calls `Update-Solution.ps1 -All -Verify`

## GitHub issue number?
https://github.com/microsoft/mcp/issues/2466

## Pre-merge Checklist
- [x] Required for All PRs

### DIFF
--- a/eng/scripts/Analyze-Code.ps1
+++ b/eng/scripts/Analyze-Code.ps1
@@ -5,20 +5,23 @@
 
 Push-Location $RepoRoot
 try {
+    $hasErrors = $false
+
     Write-Host "Checking if solution files are up to date."
-    & "$PSScriptRoot/Update-Solution.ps1" -All
-    $slnxDiff = git diff --name-only -- "*.slnx"
-    if ($slnxDiff) {
-        Write-Host "❌ Solution files are out of date. The following files need updating:"
-        Write-Host $slnxDiff
-        Write-Host "Please run './eng/scripts/Update-Solution.ps1 -All' and commit the changes."
-        git checkout -- "*.slnx"  # restore originals so other checks aren't affected
+    try {
+        & "$PSScriptRoot/Update-Solution.ps1" -All -Verify
+    } catch {
+        Write-Host "❌ Solution update failed: $_"
         $hasErrors = $true
-    } else {
-        Write-Host "✅ Solution files are up-to-date."
+    }
+    if ($LASTEXITCODE -ne 0) {
+        $hasErrors = $true
+    } 
+    if (-not $hasErrors) {
+        Write-Host "✅ Solution files are up to date."
     }
 
-    Write-Host "Running dotnet format to check for formatting issues..."
+    # Write-Host "Running dotnet format to check for formatting issues..."
     $solutionFile = "$RepoRoot/Microsoft.Mcp.slnx"
 
     # Excluding diagnostics IL2026 and IL3050 due to known issues with source generator
--- a/eng/scripts/Analyze-Code.ps1
+++ b/eng/scripts/Analyze-Code.ps1
@@ -5,6 +5,19 @@
 
 Push-Location $RepoRoot
 try {
+    Write-Host "Checking if solution files are up to date."
+    & "$PSScriptRoot/Update-Solution.ps1" -All
+    $slnxDiff = git diff --name-only -- "*.slnx"
+    if ($slnxDiff) {
+        Write-Host "❌ Solution files are out of date. The following files need updating:"
+        Write-Host $slnxDiff
+        Write-Host "Please run './eng/scripts/Update-Solution.ps1 -All' and commit the changes."
+        git checkout -- "*.slnx"  # restore originals so other checks aren't affected
+        $hasErrors = $true
+    } else {
+        Write-Host "✅ Solution files are up-to-date."
+    }
+
     Write-Host "Running dotnet format to check for formatting issues..."
     $solutionFile = "$RepoRoot/Microsoft.Mcp.slnx"
 
--- a/eng/scripts/Analyze-Code.ps1
+++ b/eng/scripts/Analyze-Code.ps1
@@ -21,68 +21,68 @@ try {
         Write-Host "✅ Solution files are up to date."
     }
 
-    # Write-Host "Running dotnet format to check for formatting issues..."
-    # $solutionFile = "$RepoRoot/Microsoft.Mcp.slnx"
+    Write-Host "Running dotnet format to check for formatting issues..."
+    $solutionFile = "$RepoRoot/Microsoft.Mcp.slnx"
 
-    # # Excluding diagnostics IL2026 and IL3050 due to known issues with source generator
-    # # Can be removed when https://github.com/dotnet/sdk/issues/45054 is resolved
-    # dotnet format $solutionFile --verify-no-changes --exclude-diagnostics IL2026 IL3050
+    # Excluding diagnostics IL2026 and IL3050 due to known issues with source generator
+    # Can be removed when https://github.com/dotnet/sdk/issues/45054 is resolved
+    dotnet format $solutionFile --verify-no-changes --exclude-diagnostics IL2026 IL3050
 
-    # # Run dotnet format
-    # if ($LASTEXITCODE) {
-    #     Write-Host "❌ dotnet format detected formatting issues."
-    #     Write-Host "Please run 'dotnet format `"$solutionFile`"' to fix the issues and then try committing again."
-    #     $hasErrors = $true
-    # } else {
-    #     Write-Host "✅ dotnet format did not detect any formatting issues."
-    # }
+    # Run dotnet format
+    if ($LASTEXITCODE) {
+        Write-Host "❌ dotnet format detected formatting issues."
+        Write-Host "Please run 'dotnet format `"$solutionFile`"' to fix the issues and then try committing again."
+        $hasErrors = $true
+    } else {
+        Write-Host "✅ dotnet format did not detect any formatting issues."
+    }
 
-    # # Run cspell spell check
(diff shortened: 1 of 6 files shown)


eng/scripts/Analyze-Code.ps1
```diff
@@ -5,6 +5,19 @@
 
 Push-Location $RepoRoot
 try {
+    Write-Host "Checking if solution files are up to date."
+    & "$PSScriptRoot/Update-Solution.ps1" -All
```
jongio: `Update-Solution.ps1` starts by `Remove-Item`ing every committed `.slnx` and regenerating via `dotnet new sln`. That means running `Analyze-Code.ps1` locally silently destroys any uncommitted `.slnx` work a developer has in their working tree, which is exactly the scenario this check is trying to catch.

The `git checkout -- "*.slnx"` on line 15 doesn't rescue it either. By the time we hit that line, Update-Solution has already overwritten the WIP.

Two options worth considering:

1. Gate the whole block on `$env:TF_BUILD` so it only runs in CI (following the pattern the cspell block uses below).
2. Before running Update-Solution, bail out if the working tree already has dirty `.slnx` files:

```powershell
if (git status --porcelain -- "*.slnx") {
    Write-Host "Skipping solution freshness check: uncommitted .slnx changes present."
} else {
    & "$PSScriptRoot/Update-Solution.ps1" -All
    ...
}
```

eng/scripts/Analyze-Code.ps1
```diff
@@ -5,6 +5,19 @@
...
+    $slnxDiff = git diff --name-only -- "*.slnx"
```
jongio: `git diff --name-only` only reports modified tracked files. If a PR adds a new server directory and Update-Solution creates a new `.slnx` for it, that file is untracked and this check won't flag it.

`git status --porcelain -- "*.slnx"` covers both modified and untracked paths:

```powershell
$slnxDiff = git status --porcelain -- "*.slnx"
```

eng/scripts/Analyze-Code.ps1
jongio: Separate concern on the same line: no exit code check on the Update-Solution invocation. `$ErrorActionPreference = "Stop"` inside that script only traps cmdlet errors. Native exit codes from `dotnet new sln` / `dotnet sln add` don't throw. If one of those fails, this script proceeds to the diff check and may produce a passing result by coincidence (or a misleading "out of date" message when the real problem is that the regeneration failed).

```powershell
& "$PSScriptRoot/Update-Solution.ps1" -All
if ($LASTEXITCODE) {
    Write-Host "Update-Solution.ps1 failed with exit code $LASTEXITCODE."
    $hasErrors = $true
}
```

jongio (commented):
A few concerns with how this check interacts with local developer workflow and edge cases. Details inline.

One meta point: the PR body is unfilled template placeholders. Please add a short description and link the issue this addresses.

eng/scripts/Update-Solution.ps1
```diff
@@ -73,20 +104,54 @@ function Update-RootSolution {
...
-    Write-Host "Adding all projects to root solution" -ForegroundColor Cyan
-    dotnet sln $slnFile add $allProjects
+    $tempSlnFile = ".temp.root.slnx"
+    Remove-Item -Path $tempSlnFile -Force -ErrorAction SilentlyContinue
+
+    try {
+        Write-Host "Creating new root solution file" -ForegroundColor Cyan
+        dotnet new sln -n ".temp.root" --format slnx
+
+        $allProjects = Get-ChildItem -Path $RepoRoot -Filter "*.csproj" -Recurse
```
jongio: The verify check compares the freshly-generated `.slnx` byte-for-byte against the committed copy, so it needs the regeneration to be deterministic across machines. `Get-ChildItem -Recurse` enumeration order isn't guaranteed: NTFS on Windows is typically alphabetical, ext4/APFS are not. The server-project enumeration on line 44 has `Sort-Object -Property FullName`, but this `-Recurse` for root projects, and the test-project enumeration on line 61, don't.

If the committed `Microsoft.Mcp.slnx` was generated on Windows and CI runs on Linux (or vice versa), the verify step can fail even when nothing is actually stale, and the remediation message would send the author on a wild goose chase.

Suggest sorting the outputs before handing them to `dotnet sln add`:

```powershell
$allProjects = Get-ChildItem -Path $RepoRoot -Filter "*.csproj" -Recurse | Sort-Object -Property FullName
```

and the same treatment on line 61 for test projects.

jongio (commented):
The pivot to a `-Verify` mode cleanly addresses the concerns from my earlier review (destructive regeneration, new untracked .slnx files, and the missing exit-code check). Copilot's inline notes on temp-file placement and the dead `$targetFile` in `Update-RootSolution` are all worth picking up.

One new concern about the verify mechanic itself, inline.

eng/scripts/Update-Solution.ps1
```diff
@@ -21,50 +23,77 @@ function Update-Solution {
...
+            $serverRelativePath = Resolve-Path $fullPath -Relative -RelativeBasePath $serverDirectory
+            $contents = $contents.Replace($match.Value, " Path=`"$($serverRelativePath.Replace('\', '/'))`"")
+        }
+        if ($Verify) {
+            $originalFile = "$serverDirectory/$serverName.slnx"
+            if (-not (Test-Path $originalFile)) {
+                Write-Host "❌ $serverName.slnx does not exist."
+                $script:stale = $true
+                $script:staleFiles += "$serverName.slnx"
+            } elseif ((Get-Content $originalFile -Raw) -ne $contents) {
```
jongio: The byte-for-byte compare here is sensitive to line endings, encoding, and trailing newlines, and a few things conspire to make that fragile across platforms:

1. `.gitattributes` pins `*.{sln,slnx} text=auto eol=crlf`, so the committed file in the working tree is always CRLF, even on Linux and macOS.
2. `dotnet new sln` writes the temp `.slnx` with LF on non-Windows, so `$contents` carries LF there.
3. The non-verify write path on line 90 uses `Set-Content -NoNewline`, which strips any trailing newline from `$contents`. The root path on line 139 uses `Move-Item`, which preserves whatever `dotnet new sln` wrote. So the per-server committed file and the regenerated `$contents` won't even agree on trailing-newline handling on a single machine.

CI is green today because the Analyze stage runs on Windows. The first time someone runs `./eng/scripts/Analyze-Code.ps1` on Mac or Linux, they'll get a false "out of date" report and the remediation message sends them on a wild goose chase, which is exactly the failure mode my earlier review flagged for the sort-order case.

A few options:

- Normalize before comparing:
  ```powershell
  function Normalize-Slnx($s) { ($s -replace "`r`n", "`n").TrimEnd("`n") }
  if ((Normalize-Slnx (Get-Content $originalFile -Raw)) -ne (Normalize-Slnx $contents)) { ... }
  ```
- Or unify the write path so root and per-server both go through the same code (e.g., always `Set-Content` with explicit `-Encoding utf8` and consistent newline handling), then the read-back compare lines up.
- Or sidestep the encoding question entirely: write regenerated content to temp files and compare via `git hash-object --path=<committed-path> <temp>` against `git hash-object <committed>`. Git applies the same `eol=crlf` filter to both inputs, so the hashes match iff the logical content matches.

Same concern applies to the root comparison on line 131.

jongio (commented):
The pivot to in-memory comparison plus the deterministic sorting cleanly resolves the destructive-regen, untracked-file, exit-code, and cross-platform sort concerns from my earlier review. One remaining concern about the compare itself, inline.

eng/scripts/Update-Solution.ps1
anannya03: I went ahead with Normalizing before comparison. 
I also added the utf-8 encoding for uniformity. Another change that I made in my recent commit was to have unique file names by adding a GUID to the temp file name, so that if we ever have parallel runs in CI, it won't result in a race condition. As part of this change, I have also modified .gitignore file.

jongio (approved):
Remaining concerns from previous rounds are addressed: normalization for line endings, Sort-Object for deterministic enumeration, GUID temp names for collision avoidance. Error propagation to Analyze-Code.ps1 works correctly via both exception path and exit code path. No new issues found.

eng/scripts/Update-Solution.ps1
jongio: Addressed by Sort-Object on all project collections.

eng/scripts/Update-Solution.ps1
jongio: Addressed by Normalize-Solution function.

eng/scripts/Analyze-Code.ps1
jongio: Moot - the approach changed to in-memory comparison.

vcolin7 (approved):
After addressing Jon's and Copilot's comments, the PR looks good to me. One thing I'd like us to do in a separate one is to update the documentation around adding new tools to tell users and agents to run this check locally instead of having us rely on it failing on CI to catch any errors :)

PR #2441, opened 2026-04-20
Add protecteditem undelete command to Azure Backup toolset

shrja-ms:
## Summary

Adds a new `azurebackup protecteditem undelete` command that restores soft-deleted backup items to an active protection state for both RSV (Recovery Services vault) and DPP (Data Protection) vault types.

### New Tool: `protecteditem undelete`

**RSV implementation:**
- Finds soft-deleted protected item by datasource ID (ARM resource ID or in-guest workload protectable item name)
- Disambiguates prefix matches when multiple file shares share the same storage account
- Supports in-guest workload names (SQL/HANA) with required `--container`
- When `--container` is provided with ARM ID, uses direct lookup via `RsvNamingHelper` to bypass listing
- Sets `IsRehydrate = true` on the matched item's existing properties (preserving PolicyId, ContainerName, etc.)
- Returns `Accepted` status with job ID from response headers

**DPP implementation:**
- Lists soft-deleted instances via `GetDeletedDataProtectionBackupInstances`
- Matches by datasource ID and calls `UndeleteAsync`
- Extracts job ID from `Azure-AsyncOperation` header (consistent with other DPP LROs)
- Returns `Accepted` status with job ID (null when unavailable, with operation ID in message)

### Files Added (3)
| File | Description |
|------|-------------|
| `ProtectedItemUndeleteCommand.cs` | Sealed command with error handling for 403/404/409 scenarios |
| `ProtectedItemUndeleteOptions.cs` | Options class extending `BaseAzureBackupOptions` with `DatasourceId` and `Container` |
| `ProtectedItemUndeleteCommandTests.cs` | 14 unit tests |

### Files Modified
| File | Description |
|------|-------------|
| `IAzureBackupService.cs` | Added `UndeleteProtectedItemAsync` to service contract |
| `IRsvBackupOperations.cs` | Added RSV undelete interface method |
| `IDppBackupOperations.cs` | Added DPP undelete interface method |
| `AzureBackupService.cs` | Routes undelete to RSV/DPP based on resolved vault type |
| `RsvBackupOperations.cs` | RSV implementation with `IsRehydrate=true`, prefix disambiguation, workload support |
| `DppBackupOperations.cs` | DPP implementation via `DeletedDataProtectionBackupInstanceResource.UndeleteAsync` |
| `AzureBackupSetup.cs` | Registers the new command in DI + command group |
| `AzureBackupJsonContext.cs` | AOT serialization context for `ProtectedItemUndeleteCommandResult` |
[...]

PR #2452, opened 2026-04-21
Add azurebackup policy update command to Azure Backup toolset

shrja-ms:
## Summary

Adds a new `azurebackup policy update` command that modifies an existing RSV backup policy's schedule time and daily retention settings. Supports IaasVM, SQL/SAP HANA (VmWorkload), and FileShare policy types. DPP vaults return a clear error since they do not support policy update.

### New Tool: `azurebackup policy update`

RSV implementation:
- Fetches existing policy via `GetAsync`, mutates schedule/retention in-place on the existing properties object (preserving all other fields like PolicyId, WorkloadType, etc.)
- Handles IaasVM (`IaasVmProtectionPolicy`), VmWorkload SQL/HANA (`VmWorkloadProtectionPolicy` Full sub-policy), and FileShare (`FileShareProtectionPolicy`)
- Calls `CreateOrUpdateAsync` with the modified policy data
- Returns `Succeeded` with no-op message when no changes specified

DPP implementation:
- Returns `InvalidOperationException` with clear message that DPP does not support policy update

### Files Added (3)

| File | Purpose |
|------|---------|
| PolicyUpdateCommand.cs | Sealed command with error handling for 403/404/DPP-not-supported |
| PolicyUpdateOptions.cs | Options class with Policy, ScheduleTime, DailyRetentionDays |
| PolicyUpdateCommandTests.cs | 12 unit tests (success, errors, validation, deserialization, partial updates) |

### Files Modified (13)

| File | Change |
|------|--------|
| IAzureBackupService.cs | Added `UpdatePolicyAsync` to service contract |
| AzureBackupService.cs | Routes update to RSV; throws `InvalidOperationException` for DPP |
| IRsvBackupOperations.cs | Added RSV update interface method |
| RsvBackupOperations.cs | RSV implementation: fetch-modify-save pattern with `UpdatePolicyScheduleAndRetention` helper |
| AzureBackupSetup.cs | Registers `PolicyUpdateCommand` in DI + policy command group |
| AzureBackupJsonContext.cs | AOT serialization context for `PolicyUpdateCommandResult` |
| AzureBackupSetupTests.cs | Verifies update command in policy group |
| AzureBackupCommandTests.cs | 6 live tests (IaasVM retention, IaasVM schedule, SQL retention, SQL schedule+retention, no-changes, DPP rejection) |
| assets.json | Updated recording tag |
| consolidated-tools.json | Added `azurebackup_policy_update` to `update_azure_backup_settings` |
| azmcp-commands.md | Command reference + metadata |
| e2eTestPrompts.md | 2 e2e test prompts |
| changelog entry | Features Added |

## Test Results

- 343 unit tests: All pass (12 new policy update tests)
- 51 live tests (playback): All pass (6 new: IaasVM x2, SQL x2, no-changes, DPP rejection)
- Consolidated tool tests: 5/5 pass
- dotnet format: Clean
- ToolDescriptionEvaluator: `azurebackup_policy_update` ranks #1 for both test prompts (0.664, 0.689)

## Invoking Livetests


shrja-ms:
Addressed all 3 Copilot review comments in the latest push:

**Comment 1** (scheduleTime/dailyRetentionDays silently treated as null):
Fixed. Non-empty values that fail parsing now throw \ArgumentException\ with clear messages (e.g., \Invalid schedule time 'abc'. Provide a valid time in UTC HH:mm format\). Retention days <= 0 also rejected. Added \GetStatusCode\ override so \ArgumentException\ returns 400 BadRequest. Added 2 new unit tests verifying this.

**Comment 2** (all InvalidOperationException mapped to DPP message):
Fixed. Narrowed mapping to only match when \x.Message\ contains \DPP\. Other \InvalidOperationException\s now surface their original message. Added 1 new unit test verifying non-DPP exceptions surface correctly.

**Comment 3** (missing default case in UpdatePolicyScheduleAndRetention):
Fixed. Added \default\ case that throws \InvalidOperationException\ for unsupported policy subtypes, preventing false-positive success responses.

**Additional** (cross-checked with PR #2441 feedback):
- In-place mutation pattern already correct (no new property instances that strip fields)
- Added policy update prompt example to README.md Azure Backup section
- Validated with \Process-PackageReadMe.ps1 -Command validate\

All 346 unit tests pass (15 policy update tests now). All 49 live tests pass in playback. Format clean.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/RsvBackupOperations.cs
```diff
@@ -724,6 +724,126 @@ public async Task<OperationResult> CreatePolicyAsync(
...
+
+        await policyCollection.CreateOrUpdateAsync(WaitUntil.Completed, policyName, policyData, cancellationToken);
+
+        return new OperationResult("Succeeded", null, $"Policy '{policyName}' updated in vault '{vaultName}'.");
+    }
+
+    private static void UpdatePolicyScheduleAndRetention(BackupGenericProtectionPolicy policyProperties, DateTimeOffset? newScheduleTime, int? newRetentionDays)
+    {
+        switch (policyProperties)
+        {
```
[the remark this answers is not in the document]
jongio: > Copilot earlier:
> "tracking whether a schedule/retention update was actually applied and failing if not, to avoid false-positive success responses"

The default case landed, but the second half is still outstanding. Inside the supported branches the inner pattern matches silently no-op:

- `IaasVmProtectionPolicy`: schedule update only fires when `vmPolicy.SchedulePolicy is SimpleSchedulePolicy`. Enhanced/V2 IaaS VM policies (created by the Azure portal or `New-AzRecoveryServicesBackupProtectionPolicy` with enhanced policy) use `SimpleSchedulePolicyV2` for hourly / enhanced daily schedules. For those, the pattern match silently fails, nothing is mutated, and `CreateOrUpdateAsync` runs against an unchanged policy while the caller still reports `Succeeded`. Same shape for retention: `LongTermRetentionPolicy` is matched, `SimpleRetentionPolicy` is silently skipped.
- `VmWorkloadProtectionPolicy`: if no sub-policy has `PolicyType.ToString() == "Full"` (legitimately possible for Differential/Log-only configurations), the loop completes without applying anything but the command still returns success.

Cheapest fix matching what you already did for the outer default: track whether each requested update (schedule, retention) actually landed, and throw `InvalidOperationException` with a precise message when an input was provided but no branch handled it. That also makes the command description ("Updates the backup schedule time and daily retention days for VM, SQL, SAP HANA, and file share workload policies") accurate without qualifications.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+        RequestFailedException reqEx => reqEx.Message,
+        InvalidOperationException ioEx when ioEx.Message.Contains("DPP", StringComparison.OrdinalIgnoreCase) =>
+            "Update is only supported for RSV (Recovery Services vault) policies. DPP policies do not support update.",
+        InvalidOperationException ioEx => ioEx.Message,
+        _ => base.GetErrorMessage(ex)
+    };
+
+    protected override HttpStatusCode GetStatusCode(Exception ex) => ex switch
+    {
+        ArgumentException => HttpStatusCode.BadRequest,
```
jongio, line 105: The DPP-not-supported case in `AzureBackupService.UpdatePolicyAsync` (line 290) throws `InvalidOperationException`, which falls through to `base.GetStatusCode(ex)` and maps to 500 InternalServerError. This is a user-input error (wrong vault type), not a server fault, so it should be 400 BadRequest.

The convention in this toolset uses `ArgumentException` for the same "wrong vault type" condition: `AzureBackupService.cs:303` (protectable item discovery), `VaultTypeResolver.cs:21,26,39`, and similar. Switching the DPP throw to `ArgumentException` lets it flow through the existing `ArgumentException => HttpStatusCode.BadRequest` arm and removes the need for the fragile `ioEx.Message.Contains("DPP")` string match in `GetErrorMessage`.

Same applies to the `InvalidOperationException`s thrown from `RsvBackupOperations.UpdatePolicyAsync` (line 747 unsupported properties type, line 860 unsupported policy variant) and any new throws added when addressing my earlier comment about silent no-op branches. Those are user-input errors too.

jongio: Still open after the latest push. `GetStatusCode` has no `InvalidOperationException` case, so both the DPP-not-supported error and the unsupported-policy-type errors from `UpdatePolicyScheduleAndRetention` return 500.

```suggestion
    protected override HttpStatusCode GetStatusCode(Exception ex) => ex switch
    {
        ArgumentException => HttpStatusCode.BadRequest,
        InvalidOperationException => HttpStatusCode.BadRequest,
        RequestFailedException reqEx => (HttpStatusCode)reqEx.Status,
        _ => base.GetStatusCode(ex)
    };
```

The `HandlesDppNotSupportedError` test assertion also needs to change from `InternalServerError` to `BadRequest`.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/RsvBackupOperations.cs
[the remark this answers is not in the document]
jongio: The `scheduleApplied`/`retentionApplied` tracking is a good fix for the `default` case, but there's a specific gap in the IaasVM and FileShare branches.

In both, `scheduleApplied = true` is set inside the retention sync block:

```csharp
// IaasVM branch, inside the retention block
if (newScheduleTime is not null)
{
    vmRetention.DailySchedule.RetentionTimes.Clear();
    vmRetention.DailySchedule.RetentionTimes.Add(scheduleRunTime);
    scheduleApplied = true; // <-- problem line
}
```

If `SchedulePolicy` isn't `SimpleSchedulePolicy` (e.g., `SimpleSchedulePolicyV2` for enhanced policies), Block 1 skips and `scheduleApplied` stays false. But Block 2 enters the retention path and sets `scheduleApplied = true` from the retention time sync alone. The command reports success even though the actual schedule policy wasn't updated.

The VmWorkload branch doesn't have this issue - its retention block correctly doesn't set `scheduleApplied`.

Fix: remove `scheduleApplied = true` from the retention time sync in IaasVM and FileShare. Only the schedule policy update block should set that flag.

jongio (commented):
Two items from my previous review are still open after the latest push. Replied on both threads with specifics.

1. `GetStatusCode` still has no `InvalidOperationException` mapping - DPP and unsupported-policy-type errors return 500 instead of 400. The unit test asserts the wrong status code.

2. `scheduleApplied` flag is set in the wrong place in the IaasVM and FileShare branches - the retention sync block sets it even when the actual schedule policy wasn't updated. VmWorkload branch is correct and can be used as reference.

anannya03:
Hi @shrja-ms ,
There's a live test failure for Azure Backup at the Test Resource step. Could you also please take a look into it? 

CC: @jongio

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
shrja-ms: Fixed in commit 9c34fd88. Added `InvalidOperationException => HttpStatusCode.BadRequest` to `GetStatusCode`, exactly as suggested. Updated both `HandlesDppNotSupportedError` and `HandlesNonDppInvalidOperationException` test assertions from `InternalServerError` to `BadRequest`. All 361 unit tests pass.

shrja-ms: Fixed in commit 9c34fd88. Added `InvalidOperationException => HttpStatusCode.BadRequest` to `GetStatusCode`, exactly as suggested. Updated both `HandlesDppNotSupportedError` and `HandlesNonDppInvalidOperationException` test assertions from `InternalServerError` to `BadRequest`. All 361 unit tests pass.

jongio (commented):
Addresses my previous feedback cleanly:

1. GetStatusCode now maps InvalidOperationException to BadRequest - both DPP-not-supported and unsupported-policy-type errors return 400 instead of 500.
2. Removed the misplaced scheduleApplied = true from the IaasVM and FileShare retention sync blocks. Schedule tracking is now scoped correctly.
3. Test assertions updated to match.
4. 	est-resources-post.ps1 key casing fix should resolve the live test failure.

One action item: live tests need to be re-triggered to verify the DeploymentOutputs key casing fix works in CI.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/RsvBackupOperations.cs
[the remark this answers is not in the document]
jongio: Fixed in the latest commit. scheduleApplied = true is no longer set inside the retention sync blocks. Verified all three branches (IaasVM, FileShare, VmWorkload) track schedule and retention flags independently.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
jongio: Confirmed - looks correct. Both the DPP case and the general unsupported-policy-type case map to 400 now. Tests are aligned.

shrja-ms:
Hi @anannya03 - the live test failure at the Test Resource step was caused by a case mismatch in test-resources-post.ps1: the script was reading DeploymentOutputs (PascalCase) but the pipeline sets deploymentOutputs (camelCase). Fixed in commit 9c34fd88. All review threads from both Copilot and @jongio are now resolved. Could someone please re-trigger live tests to verify the fix?

jongio (commented):
Addresses all my previous feedback. The four items I flagged are fixed and the threads are resolved:

- GetStatusCode maps InvalidOperationException to BadRequest (both DPP and unsupported-policy-type cases)
- scheduleApplied/retentionApplied tracking prevents false-positive success
- scheduleApplied no longer set inside the retention sync blocks for IaasVM and FileShare
- Default case throws for unsupported policy types

One blocker before merge: there are merge conflicts on this branch that need rebasing.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        var options = BindOptions(parseResult);
+
+        try
+        {
+            var result = await _azureBackupService.UpdatePolicyAsync(
+                options.Vault!,
+                options.ResourceGroup!,
+                options.Subscription!,
```
jongio, line 66: Missing telemetry tags. Both `VaultUpdateCommand` and `PolicyCreateCommand` call telemetry helpers after `BindOptions` (`AddVaultTags` / `AddVaultAndWorkloadTags`). Add `AzureBackupTelemetryTags.AddVaultTags(context.Activity, options.VaultType);` after line 58 to keep observability consistent across the command set.

jongio (commented):
Clean implementation following the established AzureBackup command patterns. Error handling is solid - `GetStatusCode` maps both `ArgumentException` and `InvalidOperationException` to `BadRequest`, and `GetErrorMessage` has specific messages for DPP rejection vs generic IOEx.

The `scheduleApplied`/`retentionApplied` tracking correctly catches unsupported schedule/retention types at the end of `UpdatePolicyScheduleAndRetention`. Tests cover the key paths including DPP rejection, invalid inputs, schedule-only and retention-only updates.

One item to add before merge - see inline comment.

jongio (commented):
All prior feedback addressed. The rebase pulled in #2505 (telemetry tags for all AzureBackup commands), so PolicyUpdateCommand now includes the `AddVaultTags` call I flagged. The `scheduleApplied`/`retentionApplied` tracking, input validation, error handling, and test coverage all look solid. No new issues.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft Corporation.
```
anannya03, line 1: Please rebase with main - there's a design change merged recently, which needs to be followed by all the new commands. All the CommandMetadata is now added at the top of the class definition.

https://teams.microsoft.com/l/message/19:Jwo_T5Rls-7KVRFup69ij9R4CoBCqtVuV3_TC7JwfRM1@thread.tacv2/1777504437594?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=256e3336-e328-4683-a24a-7a36c8a123fc&parentMessageId=1777504437594&teamName=Azure%20MCP%20Server&channelName=General&createdTime=1777504437594

tools/Azure.Mcp.Tools.AzureBackup/tests/Azure.Mcp.Tools.AzureBackup.LiveTests/assets.json
```diff
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
```
anannya03, line 2: rebase this as well because of the other PR changes that got in. You'd probably have to re-record the live tests pertaining to this PR and commit the updated assets.json

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Policy/PolicyUpdateCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+using Azure.Mcp.Tools.AzureBackup.Services;
+using Microsoft.Extensions.Logging;
+using Microsoft.Mcp.Core.Commands;
+using Microsoft.Mcp.Core.Extensions;
+using Microsoft.Mcp.Core.Models.Command;
+using Microsoft.Mcp.Core.Models.Option;
+
+namespace Azure.Mcp.Tools.AzureBackup.Commands.Policy;
+
+public sealed class PolicyUpdateCommand(ILogger<PolicyUpdateCommand> logger, IAzureBackupService azureBackupService) : BaseAzureBackupCommand<PolicyUpdateOptions>()
```
anannya03, line 28:
```
[CommandMetadata(
    Id = "a3f7d2c1-9e84-4b6a-8d3c-5f1e7a2b9c04",
    Name = "update",
    Title = "Update Backup Policy",
    Description = "Modifies an existing RSV backup policy. Updates the backup schedule time and daily retention days for VM, SQL, SAP HANA, and file share workload policies. The named policy must already exist in the vault.",
    Destructive = true,
    Idempotent = true,
    OpenWorld = false,
    ReadOnly = false,
    Secret = false,
    LocalRequired = false)]
public sealed class PolicyUpdateCommand(ILogger<PolicyUpdateCommand> logger, IAzureBackupService azureBackupService) : BaseAzureBackupCommand<PolicyUpdateOptions>()
{
    private readonly ILogger<PolicyUpdateCommand> _logger = logger;
    private readonly IAzureBackupService _azureBackupService = azureBackupService;
```

This should now be modified like this

anannya03: https://github.com/microsoft/mcp/pull/2452#discussion_r3165675843

Please refer to the teams announcement link I shared in the above comment, to see the modified design.

jongio (commented):
Reviewed the rebase changes (HEAD moved from accefe2 to 6bb3a0a). Two new commits since my last review:

- CommandMetadata attribute applied to PolicyUpdateCommand (per anannya03's request) - values look correct: Destructive=true, Idempotent=true, OpenWorld=false, ReadOnly=false.
- Live tests re-recorded after rebase, retention test updated to increase rather than decrease (handles immutable vault constraints).

All prior feedback remains addressed. No new issues found.

tools/Azure.Mcp.Tools.AzureBackup/src/AzureBackupSetup.cs
```diff
@@ -72,10 +73,11 @@ and Backup vaults (DPP/Data Protection). Supports vault management, protected it
...
         vault.AddCommand<VaultUpdateCommand>(serviceProvider);
 
-        var policy = new CommandGroup("policy", "Backup policy operations - Get policy details or list all policies, and create policies.");
+        var policy = new CommandGroup("policy", "Backup policy operations – Get policy details or list all policies, create, and update policies.");
         azureBackup.AddSubGroup(policy);
-        policy.AddCommand<PolicyGetCommand>(serviceProvider);
-        policy.AddCommand<PolicyCreateCommand>(serviceProvider);
+        RegisterCommand<PolicyGetCommand>(serviceProvider, policy);
+        RegisterCommand<PolicyCreateCommand>(serviceProvider, policy);
+        RegisterCommand<PolicyUpdateCommand>(serviceProvider, policy);
```
jongio, line 80: The merge of main deleted the private `RegisterCommand` helper but these three calls weren't migrated. This is what's breaking all builds.

```suggestion
        policy.AddCommand<PolicyGetCommand>(serviceProvider);
        policy.AddCommand<PolicyCreateCommand>(serviceProvider);
        policy.AddCommand<PolicyUpdateCommand>(serviceProvider);
```

Also update line 76's description from `Backup policy operations - Get policy details...` (replace the em-dash with a hyphen to match the other groups).

jongio (commented):
The merge of main brought in the `AddCommand<T>` API migration (#2538/#2539/#2540), but the `policy` group wasn't updated to match. The private `RegisterCommand` helper was deleted, so the three `RegisterCommand<>` calls on lines 78-80 won't compile - that's why all build checks are failing.

Also, the policy group description on line 76 still has an em-dash while all other descriptions were updated to hyphens.

tools/Azure.Mcp.Tools.AzureBackup/src/AzureBackupSetup.cs
anannya03: Resolving this as it's been fixed now

anannya03 (approved):

PR #2464, opened 2026-04-22
Add rollback-on-failure for VM and VMSS creation

haagha:
## Summary

When VM or VMSS provisioning fails partway through, orphaned network resources (NSG, VNet, Public IP, NIC) are now cleaned up automatically. Previously, failed creation attempts left these resources behind, causing cross-region mismatch errors on retry and requiring manual cleanup.

### New: `CreatedResourceTracker`
- Tracks resources created during multi-step operations
- Deletes tracked resources in reverse creation order on failure
- Best-effort rollback: logs errors, handles 404 (already deleted), does not mask the original exception

### Modified: `ComputeService.cs`
- **VM path**: `CreateOrGetNetworkResourcesAsync` now returns a `(ResourceIdentifier NicId, CreatedResourceTracker Tracker)` tuple, tracking newly created NSG, VNet, PIP, and NIC. `CreateVmAsync` wraps the VM PUT in try/catch that triggers rollback on failure.
- **VMSS path**: `CreateOrGetVmssNetworkResourcesAsync` similarly returns a tracker for the newly created VNet. `CreateVmssAsync` wraps the VMSS PUT with rollback.
- Pre-existing resources (found via GET) are **not** tracked or deleted during rollback.

### New: Unit Tests
- 8 tests for `CreatedResourceTracker`: tracking, reverse-order deletion, error continuation, 404 handling

## Rollback Behavior

| Scenario | Behavior |
|----------|----------|
| VM PUT fails, all network resources newly created | NSG, VNet, PIP, NIC deleted in reverse order |
| VM PUT fails, NSG/VNet pre-existing | Only PIP and NIC deleted |
| Network resource creation fails midway | Already-created resources rolled back |
| Rollback delete fails | Warning logged, continues deleting remaining resources |
| Resource already deleted (404) | Treated as success, logged at debug level |

## Testing

- All 8 `CreatedResourceTracker` unit tests pass
- 4 pre-existing test failures in `ExecuteAsync_ValidatesInputCorrectly` (confirmed on `upstream/main` — not caused by this PR)

## Invoking Livetests


jongio (commented):
Clean approach to tracking and rolling back orphaned network resources. The tracker class is well-isolated and the reverse-order deletion is the right call. Two things worth addressing before this leaves draft:

1. The try/catch in `CreateVmAsync` wraps `GetVmIpAddressesAsync` alongside the VM PUT. If the VM creates successfully but IP retrieval fails, rollback will try to delete the NIC the VM is actively using - that's worse than just surfacing the IP-read error. Same issue isn't present in the VMSS path (no equivalent post-creation read).

2. All rollback calls pass the original `cancellationToken`. If the caller cancelled the operation, every `DeleteAsync` in rollback will throw `OperationCanceledException` (caught as a generic warning), and no resources actually get cleaned up - the exact scenario this PR exists to fix.

tools/Azure.Mcp.Tools.Compute/src/Services/ComputeService.cs
```diff
@@ -201,33 +202,42 @@ public async Task<VmCreateResult> CreateVmAsync(
...
+                Location: createdVm.Data.Location.Name,
+                VmSize: createdVm.Data.HardwareProfile?.VmSize?.ToString(),
+                ProvisioningState: createdVm.Data.ProvisioningState,
+                OsType: effectiveOsType,
+                PublicIpAddress: publicIp,
+                PrivateIpAddress: privateIp,
+                Zones: createdVm.Data.Zones?.ToList(),
+                Tags: createdVm.Data.Tags as IReadOnlyDictionary<string, string>);
+        }
+        catch
```
jongio, line 229: This try/catch wraps both the VM PUT and `GetVmIpAddressesAsync`. If the VM creates successfully but the IP address read fails (transient network error, timeout), rollback deletes the NIC/PIP that the running VM depends on.

Consider narrowing the try to just the VM creation, then reading IPs outside it:

```csharp
VirtualMachineResource createdVm;
try
{
    var vmCollection = resourceGroupResource.GetVirtualMachines();
    var vmOperation = await vmCollection.CreateOrUpdateAsync(...);
    await WaitForLroCompletionAsync(vmOperation, cancellationToken);
    createdVm = vmOperation.Value;
}
catch
{
    _logger.LogWarning("VM creation failed...");
    await networkTracker.RollbackAsync(armClient, cancellationToken);
    throw;
}

var (publicIp, privateIp) = await GetVmIpAddressesAsync(...);
return new(...);
```

tools/Azure.Mcp.Tools.Compute/src/Services/CreatedResourceTracker.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Delete all tracked resources in reverse creation order.
+    /// Best-effort: logs errors but does not throw.
+    /// </summary>
+    public async Task RollbackAsync(ArmClient armClient, CancellationToken cancellationToken)
+    {
+        for (var i = _createdResources.Count - 1; i >= 0; i--)
+        {
+            var id = _createdResources[i];
+            try
+            {
```
jongio, line 74: If the original operation was cancelled, this `cancellationToken` is already in a cancelled state. Each `DeleteAsync` call will throw `OperationCanceledException`, get caught below as a generic warning, and nothing actually gets deleted.

Consider using `CancellationToken.None` here (or a fresh linked token with a timeout) so rollback can complete even when the caller cancelled:

```csharp
public async Task RollbackAsync(ArmClient armClient, CancellationToken cancellationToken = default)
```

haagha: added a fresh cancellation token

jongio (approved):
Addresses my previous feedback. Both the try/catch scoping in `CreateVmAsync` and the cancellation token handling in `RollbackAsync` look correct now. The 5-minute timeout for rollback cleanup is a reasonable bound. No new concerns.

tools/Azure.Mcp.Tools.Compute/src/Services/ComputeService.cs
haagha: moved the IP reads out of the try catch

jongio (approved):
Live test for the rollback path looks good. Uses a non-existent marketplace URN to force failure at the VM PUT step after network resources are provisioned, confirming the cleanup triggers correctly.

tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.Tests/ComputeCommandTests.cs
```diff
@@ -308,6 +308,59 @@ public async Task Should_create_windows_vm_with_password_auth()
...
+        // originated from the VM PUT (RequestFailedException) — confirming the
+        // pipeline reached the VM-creation step *after* provisioning network resources,
+        // which is the path that triggers rollback.
+        var results = root.GetProperty("results");
+        var type = results.GetProperty("type").GetString();
+        Assert.Equal("RequestFailedException", type);
+
+        var innerMessage = results.GetProperty("message").GetString();
+        Assert.Contains("PlatformImageNotFound", innerMessage);
+    }
```
audreyttt, line 362: Does this actually test that the resources get deleted? Seems like it just checks that the error is thrown.

tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.Tests/CreatedResourceTrackerTests.cs
```diff
@@ -0,0 +1,225 @@
...
+                        status: 400,
+                        message: "Nic reserved for another VM",
+                        errorCode: "NicReservedForAnotherVm",
+                        innerException: null);
+                }
+                return Substitute.For<ArmOperation>();
+            });
+        armClient.GetGenericResource(nicId).Returns(mockResource);
+
+        await tracker.RollbackAsync(armClient);
```
g2vinay, line 220: This test will take ~15 seconds in CI

RollbackAsync creates its own internal CancellationTokenSource and calls Task.Delay(delays[0], rollbackToken) = Task.Delay(TimeSpan.FromSeconds(15), ...). The test exercises this path with a real delay,  every CI run pays a 15-second tax on a single unit test.

The fix we can consider is inject delays via TimeProvider (available since .NET 8 / the Azure SDK baseline) or accept an IEnumerable<TimeSpan> delays parameter with a default:

```net
internal sealed class CreatedResourceTracker(ILogger logger, TimeProvider? timeProvider = null)
{
    private readonly TimeProvider _timeProvider = timeProvider ?? TimeProvider.System;
    // ...
    await Task.Delay(delays[attempt], _timeProvider, cancellationToken);
}
```

In the test, consider passing FakeTimeProvider (from Microsoft.Extensions.TimeProvider.Testing). This also unlocks the ability to test the full retry schedule without wall-clock delays.

tools/Azure.Mcp.Tools.Compute/src/Services/CreatedResourceTracker.cs
```diff
@@ -0,0 +1,116 @@
...
+                logger.LogWarning(ex, "Failed to roll back resource {ResourceId}. Manual cleanup required.", id);
+                return;
+            }
+        }
+    }
+
+    private static bool IsRetriableRollbackError(RequestFailedException ex) =>
+        ex.Status == 400 && ex.ErrorCode is
+            "NicReservedForAnotherVm" or
+            "InUseSubnetCannotBeDeleted" or
```
g2vinay, line 113: Should this include `InUsePublicIPAddressCannotBeDeleted` ?

The rollback order  NIC -> PIP -> VNet -> NSG. When the NIC delete LRO completes on the client side but Azure's control plane hasn't fully disassociated it yet, a PIP delete issued immediately after can return 400 InUsePublicIPAddressCannotBeDeleted.

jongio (approved):
Reviewed the new commits (background rollback, retry logic for NIC reservation conflicts, NIC GET-first pattern). Architecture looks solid - fire-and-forget for rollback is the right call given the NIC reservation hold time. Re-approving.

PR #2468, opened 2026-04-22
Propose fix for `Run unit tests` intermittent timeout

scbedd:
This looks sus on linux especially.


alzimmermsft (approved):

jongio (commented):
The fix matches the established pattern from `ServerStartCommandTests.cs` - `Kill(entireProcessTree: true)` to clean up child processes plus try/catch for the `HasExited`/`Kill()` race. Clean change.

One thing: `ConsolidatedModeTests.cs` has the same bare `process.Kill()` at line 77 and could hit the same intermittent timeout. `TestProxy.cs` (line 465) has a similar pattern too. Worth a follow-up to align them.

PR #2481, opened 2026-04-23
Fix CODEOWNERS typo: kenchennt → chenkennt

diberry:
The SignalR tools section references @kenchennt but the correct GitHub handle for Ken Chen is @chenkennt (verified via OSPO portal).


jongio (approved):
Confirmed `@chenkennt` is the right handle (Ken Chen) and `@kenchennt` doesn't exist (404). The two failing CI checks (Build Analyze, Build linux_arm64) aren't related to CODEOWNERS edits.

Issue #2341, opened 2026-04-03
[ONBOARD] Azure Compute - Addional VM Lifecycle, Extension, and List operations


audreyttt:
### Service / Tool Name

Azure.Compute.VM

### Contacts

Audrey Toney (audreytoney), Haider Agha (haider.agha), Bila Akpan (saakpan)

### Intended Agent Scenarios

   Enable AI agents to manage Azure VMs and VMSS through natural language commands. This onboarding focuses on
   the most common VM lifecycle commands, VM extension commands, and common list commands for VM/VMSS from the ComputeRP API.

## MCP Command Summary 

| Category | Commands |
|--------|----------|
| **VM Lifecycle** | azmcp-compute-vm-start<br>azmcp-compute-vm-stop<br>azmcp-compute-vm-deallocate<br>azmcp-compute-vm-restart |
| **VM/VMSS Discovery & List** | azmcp-compute-vmss-instance-view<br>azmcp-compute-vm-list-usage<br>azmcp-compute-vm-list-skus<br>azmcp-compute-vm-list-sizes<br>azmcp-compute-vm-ip-list |
| **VM Extensions** | azmcp-compute-vm-extension-get<br>azmcp-compute-vm-extension-set<br>azmcp-compute-vm-extension-delete |
| **VMSS Extensions** | azmcp-compute-vmss-extension-get<br>azmcp-compute-vmss-extension-set<br>azmcp-compute-vmss-extension-delete |
| **VM Disk Ops** | azmcp-compute-vm-disk-attach<br>azmcp-compute-vm-disk-detach |
 
### **azmcp-compute-vm-start**

Start a stopped or deallocated virtual machine.

**Equivalent CLI:** `az vm start`

**Required Parameters**
- `--name` (string): Virtual machine name
- `--resource-group` (string): Resource group name
- `--subscription` (string): Subscription ID or name

**Optional Parameters**
- `--no-wait` (bool): Return immediately without waiting for completion

### **azmcp-compute-vm-stop**

Gracefully stop a virtual machine by requesting a guest OS shutdown **and** deallocating compute resources.

This operation is preferred for **normal lifecycle management**, where the guest OS should be given an opportunity to shut down cleanly.

**Equivalent CLI:** `az vm stop`

**Tool Metadata**
- `Destructive: true`
- `Idempotent: true`
- `ReadOnly: false`

**Required Parameters**
- `--name` (string): Virtual machine name
- `--resource-group` (string): Resource group name
[...]

PR #2498, opened 2026-04-24
Add Compute VM power-state command

audreyttt:
## What does this PR do?
Onboarding issue: https://github.com/microsoft/mcp/issues/2341
Adds command azmcp compute vm power-state
CLI documentation: https://learn.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest

### `azmcp compute vm power-state`
Start, stop, deallocate, or restart an Azure VM via --state.
        Equivalent to 'az vm start/stop/deallocate/restart'.

### Parameters:
| Parameter | Required | Description |
|-----------|----------|-------------|
| `--subscription` | Yes | Azure subscription ID |
| `--resource-group`, `-g` | Yes | Resource group name |
| `--vm-name` | Yes | Name of the virtual machine |
| `--power-action` | Yes | Power action to apply (not the current power state): `start`, `stop`, `deallocate`, or `restart` |
| `--no-wait` | No | Return immediately without waiting for the operation to complete |
| `--skip-shutdown` | No | Skip OS shutdown when stopping (force power off). Only valid with `--power-action stop` |


tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.LiveTests/ComputeCommandTests.cs
```diff
@@ -32,6 +40,11 @@ public class ComputeCommandTests(ITestOutputHelper output, TestProxyFixture fixt
         })
     ];
 
+    public override CustomDefaultMatcher? TestMatcher => new()
```
jongio: This disables request body comparison for every test in the class, not just the new power-state tests. If a future recording update changes request bodies for disk, VM CRUD, or VMSS tests, this matcher won't catch it.

Can this be scoped to just the power-state tests? If it genuinely needs to be global, a comment explaining why would help.

servers/Azure.Mcp.Server/src/Resources/consolidated-tools.json
```diff
@@ -1564,7 +1564,7 @@
     },
     {
       "name": "update_azure_compute_vm_and_vmss",
-      "description": "Update Azure Virtual Machines (VMs) and Virtual Machine Scale Sets (VMSS). Modify VM properties like size, tags, license type, boot diagnostics, and user data. Update VMSS capacity, upgrade policy, overprovision settings, auto OS upgrade, and scale-in policies. Uses PATCH semantics - only specified properties are updated; unspecified properties remain unchanged.",
+      "description": "Update Azure Virtual Machines (VMs) and Virtual Machine Scale Sets (VMSS). Modify VM properties like size, tags, license type, boot diagnostics, and user data. Update VMSS capacity, upgrade policy, overprovision settings, auto OS upgrade, and scale-in policies. Change VM power state (start, stop, deallocate, restart). Uses PATCH semantics - only specified properties are updated; unspecified properties remain unchanged.",
```
jongio: Power-state operations (start/stop/deallocate/restart) are POST operations, not PATCH. The group description still says `Uses PATCH semantics` which is inaccurate for these tools.

Consider creating a separate tool group for power-state ops, or updating the description to reflect both update (PATCH) and power-state (POST) semantics.

tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.LiveTests/ComputeCommandTests.cs
```diff
@@ -594,6 +607,177 @@ await CallToolAsync(
...
+
+        var state = powerState.GetProperty("state");
+        Assert.Equal("stop", state.GetString());
+
+        var completed = powerState.GetProperty("completed");
+        Assert.True(completed.GetBoolean());
+    }
+
+    [Fact]
+public async Task Should_deallocate_vm()
```
jongio: From `Should_deallocate_vm` onward, all remaining test methods are at column 0 instead of the 4-space indent the rest of the file uses. Same issue in `ComputeOptionDefinitions.cs` where `// Power state options` is at column 0. Running `dotnet format` should fix both.

servers/Azure.Mcp.Server/changelog-entries/1776373272516.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Features Added"
+    description: "Added compute vm power-state tool for managing VM power state (start/stop/restart/deallocate)"
```
jongio: Missing trailing newline at end of file.

jongio (commented):
Solid implementation that follows the existing VM command patterns. A few items to address - see inline comments.

jongio (commented):
The consolidated-tools.json description update looks good - it correctly separates PATCH semantics (for vm/vmss update) from operation-based power-state actions. That addresses my earlier concern.

Three items from my prior review are still open:

1. **Formatting**: `dotnet format` still needs to run - `Should_deallocate_vm` through `Should_fail_power_state_for_nonexistent_vm` in `ComputeCommandTests.cs` are at column 0 instead of 4-space indent, and `// Power state options` in `ComputeOptionDefinitions.cs` is also at column 0. This is likely why Build Analyze is failing.

2. **`CompareBodies = false`**: Still class-wide. Can this be scoped to just the power-state tests?

3. **Changelog**: `1776373272516.yaml` is missing a trailing newline.

Also, the PR description still references `azmcp compute disk delete` instead of `azmcp compute vm power-state`.

servers/Azure.Mcp.Server/docs/azmcp-commands.md
```diff
@@ -1256,6 +1076,84 @@ azmcp compute vm delete --subscription "my-subscription" \
...
+**Command Behavior:**
+- Changes the power state of a virtual machine. Equivalent to `az vm start/stop/deallocate/restart`.
+- **start**: Powers on a stopped or deallocated VM.
+- **stop**: Shuts down the OS and powers off the VM (VM is still allocated and billing continues). Use `--skip-shutdown` to force power off without OS shutdown.
+- **deallocate**: Shuts down the OS, powers off the VM, and releases compute resources (billing stops for compute).
+- **restart**: Restarts the VM (equivalent to stop then start).
+- **With `--no-wait`**: Returns immediately after initiating the operation without waiting for completion.
+
+**Returns:**
+- VM name, ID, resource group, requested state, completion status, and a status message.
```
alzimmermsft: If `--no-wait` is specified is there a URI that can be checked for status of the VM state change that should be returned?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmPowerStateCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+using Microsoft.Mcp.Core.Models.Command;
+using Microsoft.Mcp.Core.Models.Option;
+
+namespace Azure.Mcp.Tools.Compute.Commands.Vm;
+
+public sealed class VmPowerStateCommand(ILogger<VmPowerStateCommand> logger)
+    : BaseComputeCommand<VmPowerStateOptions>(true)
+{
+    private const string CommandTitle = "Change Virtual Machine Power State";
+    private static readonly string[] s_validStates = ["start", "stop", "deallocate", "restart"];
```
alzimmermsft: Should this be turned into a case insensitive `HashSet` to simplify validation logic?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmPowerStateCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+    private readonly ILogger<VmPowerStateCommand> _logger = logger;
+
+    public override string Id => "a7c1e4b2-9d3f-4e8a-b5c6-2f1d3e4a5b6c";
+
+    public override string Name => "power-state";
+
+    public override string Description =>
+        """
+        Start, stop, deallocate, or restart an Azure VM via --state.
```
alzimmermsft:
```suggestion
        Start, stop, deallocate, or restart an Azure VM.
```

I think we can drop `via --state` as we send back the parameters for the MCP tool to the agent and it should be able to figure it out.

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmPowerStateCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+    public override async Task<CommandResponse> ExecuteAsync(CommandContext context, ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        if (!Validate(parseResult.CommandResult, context.Response).IsValid)
+        {
+            return context.Response;
+        }
+
+        var options = BindOptions(parseResult);
+
+        var computeService = context.GetService<IComputeService>();
```
alzimmermsft: Use dependency injection in the tool constructor for `IComputeService`, like this:

https://github.com/microsoft/mcp/blob/main/tools/Azure.Mcp.Tools.Storage/src/Commands/Blob/BlobGetCommand.cs#L17

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmPowerStateCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+                options.State!,
+                options.NoWait,
+                options.SkipShutdown,
+                options.Tenant,
+                options.RetryPolicy,
+                cancellationToken);
+
+            context.Response.Results = ResponseResult.Create(
+                new VmPowerStateCommandResult(result),
+                ComputeJsonContext.Default.VmPowerStateCommandResult);
```
alzimmermsft, line 118:
```suggestion
            context.Response.Results = ResponseResult.Create(
                new(result),
                ComputeJsonContext.Default.VmPowerStateCommandResult);
```

Code style nit

tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
```diff
@@ -335,6 +335,29 @@ public static class ComputeOptionDefinitions
...
 
+// Power state options
+    public const string StateName = "state";
+    public const string NoWaitName = "no-wait";
+    public const string SkipShutdownName = "skip-shutdown";
+
+    public static readonly Option<string> State = new($"--{StateName}")
+    {
+        Description = "The power state change to apply to the VM. Accepted values: start, stop, deallocate, restart.",
+        Required = false
```
alzimmermsft: Will there be other tools that use this option in the future optionally? If not, or not planned soon, mark this as `Required = true` and skip using `.AsRequired()` when registering the option.

tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
```diff
@@ -335,6 +335,29 @@ public static class ComputeOptionDefinitions
...
+
+    public static readonly Option<bool> NoWait = new($"--{NoWaitName}")
+    {
+        Description = "Return immediately without waiting for the operation to complete.",
+        Required = false
+    };
+
+    public static readonly Option<bool> SkipShutdown = new($"--{SkipShutdownName}")
+    {
+        Description = "Skip the graceful guest OS shutdown and force power off. Only compatible with the 'stop' state.",
```
alzimmermsft:
```suggestion
        Description = "Skip the graceful OS shutdown and force power off. Only compatible with the 'stop' state.",
```

Does `guest` need to be specified? I think just mentioning graceful OS shutdown should be sufficient?

tools/Azure.Mcp.Tools.Compute/src/Services/ComputeService.cs
```diff
@@ -1121,6 +1121,56 @@ public async Task<bool> DeleteVmAsync(
...
+                break;
+            case "deallocate":
+                await vmResource.DeallocateAsync(waitUntil, cancellationToken: cancellationToken);
+                break;
+            case "restart":
+                await vmResource.RestartAsync(waitUntil, cancellationToken);
+                break;
+            default:
+                throw new ArgumentException($"Invalid state '{state}'. Accepted values: start, stop, deallocate, restart.", nameof(state));
+        }
```
alzimmermsft: We'll want to hook into LRO poll interval injection for playback testing with `WaitForLroCompletionAsync` in `BaseAzureService` if `--no-wait` is false. Do all the power state change methods return an `Operation<T>` here? If so, this would be a better choice:


```suggestion
        ArmOperation operation;
        switch (state.ToLowerInvariant())
        {
            case "start":
                operation = await vmResource.PowerOnAsync(WaitUntil.Started, cancellationToken);
                break;
            case "stop":
                operation = await vmResource.PowerOffAsync(WaitUntil.Started, skipShutdown, cancellationToken);
                break;
            case "deallocate":
                operation = await vmResource.DeallocateAsync(WaitUntil.Started, cancellationToken: cancellationToken);
                break;
            case "restart":
                operation = await vmResource.RestartAsync(WaitUntil.Started, cancellationToken);
                break;
            default:
                throw new ArgumentException($"Invalid state '{state}'. Accepted values: start, stop, deallocate, restart.", nameof(state));
        }

        if (!noWait)
        {
            await WaitForLroCompletionAsync(operation, cancellationToken);
        }
```

This will start all the operations and if `--no-wait` was false it will then wait until the LRO completes (success or failure).

tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.UnitTests/Vm/VmPowerStateCommandTests.cs
```diff
@@ -0,0 +1,406 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Mcp.Core.Models.Command;
+using NSubstitute;
+using NSubstitute.ExceptionExtensions;
+using Xunit;
+
+namespace Azure.Mcp.Tools.Compute.UnitTests.Vm;
+
+public class VmPowerStateCommandTests
```
alzimmermsft: Another more recent design change, unit tests should extend `CommandUnitTestsBase<TCommand, TService>`. This simplifies a lot of the substitution and execution code for unit tests.

Should be able to use the other Compute unit test classes as an example:

https://github.com/microsoft/mcp/blob/main/tools/Azure.Mcp.Tools.Compute/tests/Azure.Mcp.Tools.Compute.UnitTests/Vm/VmUpdateCommandTests.cs

servers/Azure.Mcp.Server/docs/azmcp-commands.md
audreyttt: Added a status URI to the response

KarishmaGhiya (approved):

KarishmaGhiya:
The previous feedback has been addressed and the changes look good! 
The implementation is solid, well-tested, and follows existing patterns.
One suggestion: VmUpdateCommand.cs had its "Do not use this..." disambiguation line removed, but since compute_vm_power-state is grouped under update_azure_compute_vm_and_vmss in consolidated-tools.json, prompts like "restart my VM" could misroute to vm update. Consider
adding to the VmUpdateCommand description:
 "Do not use this to change VM power state (start, stop, deallocate, restart); use VM power-state instead."

jongio (commented):
Addresses all my previous feedback. DI refactoring, status URI for no-wait, scoped `CompareBodies`, and formatting all look correct. No new issues.

g2vinay (changes requested):
Run The Tool Description Evaluator and share the latest results here, post feedback being addressed.

audreyttt:
Results: ### Success Rate Metrics

**Top Choice Success:** 100.0% (97/97 tests)  

#### Confidence Level Distribution

**💪 Very High Confidence (≥0.8):** 0.0% (0/97 tests)  
**🎯 High Confidence (≥0.7):** 2.1% (2/97 tests)  
**✅ Good Confidence (≥0.6):** 32.0% (31/97 tests)  
**👍 Fair Confidence (≥0.5):** 81.4% (79/97 tests)  
**👌 Acceptable Confidence (≥0.4):** 100.0% (97/97 tests)  
**❌ Low Confidence (<0.4):** 0.0% (0/97 tests)  

#### Top Choice + Confidence Combinations

**💪 Top Choice + Very High Confidence (≥0.8):** 0.0% (0/97 tests)  
**🎯 Top Choice + High Confidence (≥0.7):** 2.1% (2/97 tests)  
**✅ Top Choice + Good Confidence (≥0.6):** 32.0% (31/97 tests)  
**👍 Top Choice + Fair Confidence (≥0.5):** 81.4% (79/97 tests)  
**👌 Top Choice + Acceptable Confidence (≥0.4):** 100.0% (97/97 tests)

KarishmaGhiya:
Looks good to me!! Requested feedback has been addressed.

KarishmaGhiya (approved):

servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
```diff
@@ -277,7 +285,7 @@ This file contains prompts used for end-to-end testing to ensure each tool is in
 | compute_vmss_update | Add tags to scale set <vmss-name> in resource group <resource-group-name> |
 | compute_vmss_delete | Delete scale set <vmss-name> in resource group <resource-group-name> |
 | compute_vmss_delete | Remove VMSS <vmss-name> from resource group <resource-group-name> |
-| compute_vmss_delete | Destroy virtual machine scale set <vmss-name> in resource group <resource-group-name> |
+| compute_vmss_delete | Delete and destroy virtual machine scale set <vmss-name> in resource group <resource-group-name> |
```
haagha: again the intention is to evaluate accuracy for different acceptable verbs - please remove delete from this as another prompt has delete already

servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
```diff
@@ -260,9 +260,17 @@ This file contains prompts used for end-to-end testing to ensure each tool is in
...
+| compute_vm_power-state | Deallocate VM <vm-name> in resource group <resource-group-name> to release compute resources while keeping the VM |
+| compute_vm_power-state | Restart VM <vm-name> in resource group <resource-group-name> |
+| compute_vm_power-state | Stop VM <vm-name> in resource group <resource-group-name> and skip the OS shutdown |
+| compute_vm_power-state | Start VM <vm-name> in resource group <resource-group-name> without waiting for completion |
+| compute_vm_power-state | Power off and shut down VM <vm-name> in resource group <resource-group-name> |
+| compute_vm_power-state | Deallocate and power off VM <vm-name> to stop billing for compute resources while preserving the VM |
 | compute_vmss_create | Create a virtual machine scale set named <vmss-name> in resource group <resource-group-name> |
 | compute_vmss_create | Create a VMSS with 3 instances in <resource-group-name> |
-| compute_vmss_create | Deploy a scale set with Rolling upgrade policy and 5 instances |
+| compute_vmss_create | Create and deploy a new virtual machine scale set with Rolling upgrade policy and 5 instances |
```
haagha: the intention is to evaluate accuracy for different acceptable verbs - we should have evals running on Create and deploy separately. can you remove create from this?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmCreateCommand.cs
```diff
@@ -19,15 +19,15 @@ namespace Azure.Mcp.Tools.Compute.Commands.Vm;
...
+        Create, deploy, or provision a single Azure Virtual Machine (VM) with its OS disk.
         Use this to launch a new Linux or Windows VM with SSH key or password authentication.
         Automatically creates networking resources (VNet, subnet, NSG, NIC, public IP) when not specified.
         Equivalent to 'az vm create'. Defaults to Standard_D2s_v5 VM size when not specified.
         The --image option is required and has no default; if the user does not specify an image, ask them which image to use
         (an alias such as 'Ubuntu2404' or 'Win2022Datacenter', a marketplace URN like 'publisher:offer:sku:version',
         or a shared gallery image ID starting with '/sharedGalleries/').
         For Linux VMs with SSH, read the user's public key file (e.g., ~/.ssh/id_rsa.pub) and pass its content.
-        Do not use this for creating Virtual Machine Scale Sets with multiple identical instances (use VMSS create instead).
+        Do not use this for creating standalone managed disks (use disk create) or Virtual Machine Scale Sets (use VMSS create).
```
haagha: why would this get confused with standalone managed disks?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmDeleteCommand.cs
```diff
@@ -17,13 +17,13 @@ namespace Azure.Mcp.Tools.Compute.Commands.Vm;
...
-        Delete, remove, or destroy an Azure Virtual Machine (VM).
-        Use this to permanently remove a VM that is no longer needed.
+        Delete, remove, or destroy an Azure Virtual Machine (VM) permanently.
+        Use this only when the VM is no longer needed and should be permanently removed.
         Equivalent to 'az vm delete'. This operation is irreversible and the VM data will be lost.
         Use --force-deletion to force delete the VM even if it is in a running or failed state
         (passes forceDeletion=true to the Azure API).
         Associated resources like disks, NICs, and public IPs are NOT automatically deleted.
-        Do not use this to delete Virtual Machine Scale Sets (use VMSS delete instead).
+        Do not use this to stop or pause a VM; use the VM power-state command instead.
```
haagha, line 26: can you add VMSS deletion back in? unless this was purposefully ommited

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmGetCommand.cs
```diff
@@ -17,7 +17,7 @@ namespace Azure.Mcp.Tools.Compute.Commands.Vm;
     Id = "c1a8b3e5-4f2d-4a6e-8c7b-9d2e3f4a5b6c",
     Name = "get",
     Title = "Get Virtual Machine(s)",
-    Description = "List or get Azure Virtual Machine (VM) configuration and properties in a resource group. By default, returns VM details including name, location, size, provisioning state, and OS type. When retrieving a specific VM with --vm-name and --instance-view, the response also includes power state (running/stopped/deallocated). Use this tool to retrieve VM configuration details.",
+    Description = "List all Azure Virtual Machines (VMs) in a subscription, or query VMs in a specific resource group. Show all VMs or retrieve a specific VM by name. Returns read-only VM details including name, location, VM size, provisioning state, OS type, and network interfaces. Use --instance-view to query and check the current runtime status and power state of a VM along with provisioning state. This is a read-only inspection and inventory tool for viewing VM configuration, properties, and runtime status.",
```
haagha: does this improve VM get evals? I would like to add back the term GET in first sentence

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmUpdateCommand.cs
```diff
@@ -18,10 +18,10 @@ namespace Azure.Mcp.Tools.Compute.Commands.Vm;
     Name = "update",
     Title = "Update Virtual Machine",
     Description = """
-        Update, modify, or reconfigure an existing Azure Virtual Machine (VM).
-        Use this to resize a VM, update tags, configure boot diagnostics, or change user data.
+        Update, modify, or reconfigure an existing Azure Virtual Machine (VM) configuration.
+        Use this to add or change tags on a VM, resize a VM to a different size, enable or configure boot diagnostics, or update user data.
         Equivalent to 'az vm update'. The VM may need to be deallocated before resizing to certain sizes.
-        Do not use this to create a new VM (use VM create) or to update Virtual Machine Scale Sets (use VMSS update).
+        Do not use this to change VM power state (start, stop, deallocate, restart); use VM power-state instead.
```
haagha, line 24: please add back the instruction to not use this for VMSS

tools/Azure.Mcp.Tools.Compute/src/Commands/Vmss/VmssDeleteCommand.cs
```diff
@@ -17,12 +17,12 @@ namespace Azure.Mcp.Tools.Compute.Commands.Vmss;
     Name = "delete",
     Title = "Delete Virtual Machine Scale Set",
     Description = """
-        Delete, remove, or destroy an Azure Virtual Machine Scale Set (VMSS) and all its VM instances.
+        Delete, remove, destroy, or tear down an Azure Virtual Machine Scale Set (VMSS) and all its VM instances.
```
haagha: is tear down part of azure nomenclature? sounds funny

tools/Azure.Mcp.Tools.Compute/src/Models/VmPowerStateResult.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Text.Json.Serialization;
+
+namespace Azure.Mcp.Tools.Compute.Models;
+
+public sealed record VmPowerStateResult(
+    [property: JsonPropertyName("name")] string Name,
+    [property: JsonPropertyName("id")] string? Id,
+    [property: JsonPropertyName("resourceGroup")] string ResourceGroup,
+    [property: JsonPropertyName("state")] string State,
```
haagha: which state is this? The preoperation state or post?

tools/Azure.Mcp.Tools.Compute/src/Commands/Vm/VmPowerStateCommand.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Azure.Mcp.Tools.Compute.Commands.Vm;
+
+public sealed class VmPowerStateCommand(ILogger<VmPowerStateCommand> logger, IComputeService computeService)
+    : BaseComputeCommand<VmPowerStateOptions>(true)
+{
+    private const string CommandTitle = "Change Virtual Machine Power State";
+    private static readonly HashSet<string> s_validStates = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "start", "stop", "deallocate", "restart"
```
haagha, line 22: these are not power states - these are operation names to modify power states.

tools/Azure.Mcp.Tools.Compute/src/Options/Vm/VmPowerStateOptions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Mcp.Tools.Compute.Options.Vm;
+
+public class VmPowerStateOptions : BaseComputeOptions
+{
+    public string? VmName { get; set; }
+
+    public string? State { get; set; }
```
haagha: this should be called powerstateOperation or something similar. start/stop/restart are not power states and shouldn't be confused with actual powerstates:

Power States:
Creating | Virtual machine is allocating resources. | Not Billed*
Starting | Virtual machine is powering up. | Billed
Running | Virtual machine is fully up. This state is the standard working state. | Billed
Stopping | This state is transitional between running and stopped. | Billed
Stopped | The virtual machine is allocated on a host but not running. Also called PoweredOff state or Stopped (Allocated). This state can be result of invoking the PowerOff API operation or invoking shutdown from within the guest OS. The Stopped state might also be observed briefly during VM creation or while starting a VM from Stopped (Deallocated) state. | Billed
Deallocating | This state is transitional between Running and Deallocated. | Not billed*
Deallocated | The virtual machine has released the lease on the underlying hardware. If the machine is powered off it is shown as Stopped (Deallocated). If it has entered hibernation it is shown as Hibernated (Deallocated)

tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
```diff
@@ -335,6 +335,29 @@ public static class ComputeOptionDefinitions
         Required = false
     };
 
+    // Power state options
+    public const string StateName = "state";
```
haagha: please change this to powerOperation? powerAction or something similar

servers/Azure.Mcp.Server/docs/azmcp-commands.md
```diff
@@ -1423,6 +1423,84 @@ azmcp compute vm delete --subscription "my-subscription" \
 | `--vm-name` | Yes | Name of the virtual machine to delete |
 | `--force-deletion` | No | Force delete the VM even if running or failed (Azure API forceDeletion) |
 
+```bash
+# Change Virtual Machine Power State
+# ✅ Destructive | ✅ Idempotent | ❌ OpenWorld | ❌ ReadOnly | ❌ Secret | ❌ LocalRequired
+azmcp compute vm power-state --subscription <subscription> \
+                             --resource-group <resource-group> \
+                             --vm-name <vm-name> \
+                             --state <start|stop|deallocate|restart> \
```
haagha: please modify --state to represent an action and not a state. --power-state-action or --power-action or something similar

haagha (changes requested):
Please change the parameter option --state to power-state-action or power-action or something similar. This needs to be not confused with actual powersates which are:
Creating 
Starting 
Running 
Stopping 
Stopped also called PoweredOff state or Stopped (Allocated). 
Deallocating 
Deallocated . If the machine is powered off it is shown as Stopped (Deallocated). If it has entered hibernation it is shown as Hibernated (Deallocated)

Similarly the test's need to get the VM with --instance-view and check for powerStates after executing the operation

tools/Azure.Mcp.Tools.Compute/src/Models/VmPowerStateResult.cs
audreyttt: the pre-operation state. dropping it because it is not useful in the result

tools/Azure.Mcp.Tools.Compute/src/Commands/Vmss/VmssDeleteCommand.cs
audreyttt: you're right, removing

haagha (approved):

tools/Azure.Mcp.Tools.Compute/src/Options/ComputeOptionDefinitions.cs
```diff
@@ -335,6 +335,29 @@ public static class ComputeOptionDefinitions
...
 
+    // Power state options
+    public const string PowerActionName = "power-action";
+    public const string NoWaitName = "no-wait";
+    public const string SkipShutdownName = "skip-shutdown";
+
+    public static readonly Option<string> PowerAction = new($"--{PowerActionName}")
+    {
+        Description = "The power action to apply to the VM (not the current power state). Accepted values: start, stop, deallocate, restart.",
+        Required = true
```
g2vinay: ComputeOptionDefinitions.PowerAction has Required = true baked into the static definition, violating the composable .AsRequired() pattern required by AGENTS.md. Every other option in that file uses [Required = false] + .AsRequired()  at the call site.

tools/Azure.Mcp.Tools.Compute/src/Models/VmPowerStateResult.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json.Serialization;
+
+namespace Azure.Mcp.Tools.Compute.Models;
+
+public sealed record VmPowerStateResult(
```
g2vinay, line 8: VmPowerStateResult omits powerAction field, making --no-wait responses ambiguous
Can it include powerAction field ?

audreyttt: Added

g2vinay (approved):

PR #2504, opened 2026-04-27
AzureBackup: PolicyCreate overhaul — schedule/retention options, builders, validators, 55 live tests

shrja-ms:
## Summary

Complete overhaul of `azmcp azurebackup policy create` with full schedule, retention, and archive tier support across all RSV and DPP workloads.

### What changed

**New options (25+):** Schedule frequency/days/times, hourly intervals, daily/weekly/monthly/yearly retention with relative and absolute formats, archive tier mode/days, SQL Full/Differential/Log sub-policies, SAPHANA incremental/snapshot, VM Enhanced/smart-tier/instant-RP, DPP vault-tier copy, backup mode (vaulted/continuous), PITR retention, AKS scoping flags, policy tags.

**PolicyCreateValidator** — Pre-flight validation that converts opaque service 400s into actionable client-side messages:
- Workload-family exclusivity (RSV-only vs DPP-only flags)
- SQL: Full/Diff day overlap, Diff single-day-only, Log retention ≥7 days, Log < Diff retention
- DPP: Archive tier rejected for all DPP datasources (none support ArchiveStore today)
- AzureDisk: Yearly retention rejected (manifest allows Daily/Weekly/Monthly only)
- Required-input combinations (weekly needs days-of-week, hourly needs all three inputs, etc.)
- Continuous DPP workloads reject schedule/retention/archive flags

**RsvPolicyBuilder** — Builds RSV policies for VM, SQL/SAPHANA/SAPASE, FileShare:
- Weekly multi-day Full + single-day Differential for SQL
- Weekly retention days-of-week auto-aligns to schedule days when omitted
- Monthly/yearly relative format (week-of-month) for Weekly schedules
- Archive tier (TieringPolicy) on VM Enhanced and SQL Full sub-policies
- Hourly Enhanced V2 with instant-RP and snapshot consistency
- Smart-tier (TierRecommended) for VM

**DppPolicyBuilder** — Builds DPP policies for Disk, Blob, ADLS, AKS, PostgreSQL, CosmosDB, ElasticSAN:
- Vault-tier copy with OS→VS lifecycle for AzureDisk
- Vaulted mode for Blob/ADLS (VaultStore-only, no AzureBackupRule)
- Continuous mode with PITR retention for Blob/ADLS
- Per-datasource profiles via DppDatasourceRegistry

**PolicyCreateRequest** — Clean POCO decoupling command options from builder internals.

### Test coverage

- **425 unit tests** (validator shape rules, builder behaviors, datasource registry)
- **55 live tests passing**, 3 skipped (private preview features: smart-tier, vaulted Blob, vaulted ADLS)
- Started at 51 passed / 7 skipped → resolved all 4 originally-deferred tests:
  - DPP Disk vault-tier multi-tier (dropped unsupported archive/yearly per manifest)
  - SQL Full+Diff+Log (Weekly multi-day Full, single Diff, retention constraints via Kusto log analysis)
  - SQL Weekly+Archive (workload type fix, relative monthly format, log retention)
  - VM Weekly+Archive (relative monthly/yearly format, days alignment, Enhanced sub-type)

### Files changed

28 files, +4663 / -210 lines across options, validators, builders, tests, docs, and changelog.

## Invoking Livetests


tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+            "TIERAFTER" => new CopyOnExpirySetting(),
+            _ when hasDays => new CustomCopySetting { Duration = TimeSpan.FromDays(afterDays) },
+            _ => new CopyOnExpirySetting(),
+        };
+
+        lifeCycle.TargetDataStoreCopySettings.Add(
+            new TargetCopySetting(copySetting, new DataStoreInfoBase(DataStoreType.ArchiveStore, "DataStoreInfoBase")));
+    }
+
+    private static void AppendVaultTierCopyIfRequested(SourceLifeCycle lifeCycle, PolicyCreateRequest request)
```
jongio: This no-op stub discards both parameters and the comment says vault-tier copy is handled by `BuildVaultTierCopyRetentionRule`. The single call site at line 109 can just be removed along with this method - dead code that'll confuse future readers.
Suggestion: delete the method and the call at line 109. If you want to preserve the design note, a comment at the former call site (`// Vault-tier copy emitted as a separate retention rule`) is cleaner.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+        }
+        value = 0;
+        return false;
+    }
+
+    /// <summary>
+    /// Normalizes RSV-style frequency names (Daily/Weekly) to ISO-8601 durations for DPP repeating intervals.
+    /// If the value is already an ISO duration (starts with P or PT), it is passed through unchanged.
+    /// </summary>
+    private static string NormalizeDppScheduleInterval(string frequency) => frequency.Trim().ToUpperInvariant() switch
```
jongio, line 295: The fallback arm passes the raw string through assuming ISO-8601, but there's no validation that it actually is a valid duration. A user typo (`--schedule-frequency Monthly` instead of `P1M`) silently produces an invalid repeating interval that only fails at the service with an opaque error.

Consider either: (a) adding a validator rule for DPP `--schedule-frequency` against a known set (`Daily`, `Weekly`, `PT4H`, `PT6H`, `PT8H`, `PT12H`, `P1D`, `P1W`, `P2W`, `P1M`), or (b) adding a regex check here for ISO-8601 duration format (`^P(\d+[YMWD])?(T(\d+[HMS])?)?$`).

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/RsvPolicyBuilder.cs
```diff
@@ -0,0 +1,850 @@
...
+            "V2" or "ENHANCED" => IaasVmPolicyType.V2,
+            _ => new IaasVmPolicyType(text!),
+        };
+    }
+
+    private static ScheduleRunType ParseScheduleRunType(string? text, bool defaultDaily)
+    {
+        if (string.IsNullOrWhiteSpace(text))
+        {
+            return defaultDaily ? ScheduleRunType.Daily : ScheduleRunType.Daily;
```
jongio: Both branches of this ternary return `ScheduleRunType.Daily` - the `defaultDaily` parameter has no effect here. Same issue on line 805. Looks like a copy-paste artifact.

```suggestion
return ScheduleRunType.Daily;
```

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/RsvPolicyBuilder.cs
```diff
@@ -0,0 +1,850 @@
...
+    private static BackupDayOfWeek NormalizeDayOfWeek(string text) => text.ToUpperInvariant() switch
+    {
+        "MON" or "MONDAY" => BackupDayOfWeek.Monday,
+        "TUE" or "TUES" or "TUESDAY" => BackupDayOfWeek.Tuesday,
+        "WED" or "WEDNESDAY" => BackupDayOfWeek.Wednesday,
+        "THU" or "THUR" or "THURS" or "THURSDAY" => BackupDayOfWeek.Thursday,
+        "FRI" or "FRIDAY" => BackupDayOfWeek.Friday,
+        "SAT" or "SATURDAY" => BackupDayOfWeek.Saturday,
+        "SUN" or "SUNDAY" => BackupDayOfWeek.Sunday,
+        _ => BackupDayOfWeek.Sunday,
```
jongio: Silently defaulting to Sunday for unrecognized input is risky. If a user passes `--schedule-days-of-week "Foobar"`, they get a Sunday schedule with no warning. The same pattern applies to `NormalizeMonthOfYear` (defaults to January, line 735) and `ParseWeeksOfMonth` (defaults to First).

Consider throwing an `ArgumentException` for unrecognized values here and catching it in the validator to produce a clear error message - or validate day/month names in `PolicyCreateValidator` before they reach the builder.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/RsvPolicyBuilder.cs
```diff
@@ -0,0 +1,850 @@
...
+    {
+        var list = new List<BackupDayOfWeek>();
+        if (string.IsNullOrWhiteSpace(csv))
+        {
+            return list;
+        }
+
+        foreach (var raw in csv!.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries))
+        {
+            var normalized = raw.Length switch
```
jongio: Both arms of this switch do the same thing (`Capitalize(raw)`). Can simplify to just `var normalized = Capitalize(raw);`.

jongio (commented):
Thorough overhaul - the validator/builder/request layering is well-designed and the test suite is solid for the scope. A few items worth addressing before merge:

**Silent defaults for invalid input** - NormalizeDayOfWeek, NormalizeMonthOfYear, and ParseWeeksOfMonth in RsvPolicyBuilder silently default to Sunday/January/First when given unrecognized strings. If a user types --schedule-days-of-week "Foobar", they'll get a Sunday schedule with no warning. Consider validating known values in the validator (or at least logging a warning).

**DPP schedule-frequency passthrough** - NormalizeDppScheduleInterval normalizes "Daily" and "Weekly" but passes anything else verbatim into the repeating interval string. A typo like --schedule-frequency "BADVALUE" becomes R/.../BADVALUE and only fails at the Azure API with an opaque error. A validator rule for known DPP frequencies would give a much better error message.

**Test coverage gaps** - The validator has solid coverage overall, but a few rules lack dedicated unit tests: SQL Full/Diff day-overlap detection (HasDayOverlap), SQL Log retention constraints (min 7 days, must be < diff retention), archive-tier-on-DPP rejection (Rule A.8), and yearly-on-Disk rejection (Rule A.9). These are the most complex validation paths and would benefit from targeted tests.

**No SAPASE builder tests** - SAPASE has unique behavior (incremental allowed, SQL compression forbidden) but zero dedicated builder unit tests. The live tests don't cover SAPASE either.

**Parse helpers untested** - ParseScheduleTimes, ParseDaysOfWeek, ParseMonthsOfYear, ParseWeeksOfMonth, ParseDaysOfMonth have non-trivial normalization (abbreviations, comma splitting, case handling) with no dedicated tests. If the normalization logic changes, regressions won't be caught.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -24,12 +24,76 @@ public static class AzureBackupOptionDefinitions
...
+    public const string SnapshotInstantRpRetentionDaysName = "snapshot-instant-rp-retention-days";
+    public const string SnapshotInstantRpResourceGroupName = "snapshot-instant-rp-resource-group";
+    // DPP Disk vault tier copy
+    public const string EnableVaultTierCopyName = "enable-vault-tier-copy";
+    public const string VaultTierCopyAfterDaysName = "vault-tier-copy-after-days";
+    // DPP Blob/ADLS backup mode (Continuous vs Vaulted)
+    public const string BackupModeName = "backup-mode";
+    // DPP PITR retention for continuous Blob/ADLS
+    public const string PitrRetentionDaysName = "pitr-retention-days";
+    // DPP policy-level tags
```
jongio: This comment says `DPP policy-level tags` but the option is RSV-only (the validator at `PolicyCreateValidator` line ~290 rejects it for DPP, and the option description says `RSV only`). Should read `// RSV policy-level tags`.

jongio (commented):
Addresses all my previous feedback - the redundant ternary, silent day/month/week defaults, dead no-op stub, redundant switch, option comment, and DPP schedule-frequency passthrough are all fixed. The new `ArgumentException` throws for bad input and the `IsKnownDppScheduleFrequency` validator are the right approach. Test additions (+230 validator, +155 builder) fill the gaps I called out.

The `BuildRsvProtectResultAsync` consolidation and DPP `WaitUntil.Completed` change both look solid. Removing the 180s container polling loop for VM protection (in favor of the direct PUT approach that `az backup` uses) is a good call.

One pervasive issue: mojibake characters (`ΓÇö`, `Γ£à`, `ΓÜá`, `≡ƒö¼`) appear ~45 times across 9 files. These are garbled Unicode (em dashes, checkmarks, warning signs, wrench emoji) likely from an editor encoding mismatch. They render incorrectly in IDE tooltips, `--help` output, and generated docs. A find/replace across the AzureBackup directory should fix all of them. Files affected: `PolicyCreateValidator.cs` (14), `AzureBackupCommandTests.cs` (12), `AzureBackupOptionDefinitions.cs` (6), `ProtectedItemProtectCommandTests.cs` (5), `DppPolicyBuilder.cs` (3), `ProtectResult.cs` (2), `DppBackupOperations.cs` (1), `RsvPolicyBuilder.cs` (1), `PolicyCreateValidatorTests.cs` (1).

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/PolicyCreateValidator.cs
```diff
@@ -0,0 +1,689 @@
...
+    /// </summary>
+    public static PolicyValidationResult Validate(PolicyCreateOptions options)
+    {
+        ArgumentNullException.ThrowIfNull(options);
+
+        var issues = new List<PolicyValidationIssue>();
+        var workload = (options.WorkloadType ?? string.Empty).Trim();
+        var family = ClassifyWorkload(workload);
+
+        // Rule D: CosmosDB pass-through — no special validator action; fall through to common rules.
```
jongio: `ΓÇö` here (and ~13 more times in this file) is mojibake for an em dash. The UTF-8 bytes are `CE 93 C3 87 C3 B6` instead of `E2 80 94`. Replace with a plain hyphen `-` or rewrite. Same issue across 8 other files in this PR (~45 total instances).

servers/Azure.Mcp.Server/docs/azmcp-commands.md
```diff
@@ -792,15 +792,69 @@ azmcp azurebackup vault update --subscription <subscription> \
 
 ```bash
 # Creates a backup policy for a specified workload type with schedule and retention rules.
+# Supports daily/weekly/hourly schedules, multi-tier retention (daily/weekly/monthly/yearly), and archive
+# tiering for Recovery Services vaults (RSV). Backup vaults (DPP) do not currently support archive tiering.
 # ✅ Destructive | ❌ Idempotent | ❌ OpenWorld | ❌ ReadOnly | ❌ Secret | ❌ LocalRequired
 azmcp azurebackup policy create --subscription <subscription> \
```
alzimmermsft, line 859: The number of parameters to this tool concerns me on how well it'll work, as all the descriptions will likely eat up a huge chunk of the agent's context window. Is this a case where this should be broken down to a multi-tool flow where subsequent tools are more targeted?

For example, this handles what looks to be the creation for 5+ different backup concepts, so if I wanted to interact with RSV-VM I need to parse through a ton of options that don't work for my use case.

shrja-ms: From Azure backup perspective today it is a single SDK command too for each DPP and RSV. What is the guidance on complicated commands in general?

jongio (commented):
Mojibake fix looks good - all ~45 garbled characters across 18 files replaced with clean ASCII. This addresses my last remaining comment.

All findings from both my previous reviews have been addressed. No new issues in the incremental diff.

servers/Azure.Mcp.Server/CHANGELOG.md
```diff
@@ -6,10 +6,37 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Features Added
 
+- Overhauled `azurebackup policy create` with comprehensive schedule and retention support across both Recovery Services vaults (RSV) and Backup vaults (DPP). Added 38 new flags including:
```
anannya03: These are a lot of changelog entries- do we need all of these?
Also it'd help to use the .yaml format to add changelog entries to prevent encountering merge conflicts, as they aren't fun to resolve.

servers/Azure.Mcp.Server/docs/azmcp-commands.md
anannya03: I agree with Alan- these are ~50 parameters. My assumption is that the agent _might be_ able to handle it, but it'll be context heavy.

servers/Azure.Mcp.Server/CHANGELOG.md
shrja-ms: Switched to YAML changelog entries via New-ChangelogEntry.ps1 to avoid merge conflicts. Removed all direct CHANGELOG.md edits and created 2 YAML entry files (features + bugs fixed). Also rebased on upstream/main and squashed into one clean commit.

shrja-ms: Switched to YAML changelog entries to avoid merge conflicts. Rebased and squashed.

jongio (commented):
PR was rebased and squashed on main. Source code is clean - all prior findings remain addressed in the squashed commit. Changelog entries and e2e test prompts look good.

One remaining issue: rchitecture.md still has 32 mojibake characters (same encoding problem as the source files that were fixed in the previous commit). The feature support matrix section has garbled UTF-8 throughout.

tools/Azure.Mcp.Tools.AzureBackup/docs/architecture.md
```diff
@@ -304,6 +304,47 @@ The toolset exposes **15 commands** organized in **9 command groups**:
 | Azure Data Lake Storage | `AzureDataLakeStorage` | adls, datalake | `.../storageAccounts/blobServices` | Operational | Continuous |
 | Azure Cosmos DB | `CosmosDB` | cosmosdb, cosmos | `Microsoft.DocumentDB/databaseAccounts` | Operational | Continuous |
 
+### `policy create` — Feature Support Matrix
```
jongio: This new section has ~32 mojibake characters (same issue fixed in source files). The garbled sequences: `ΓÇö` (em dash), `Γ£à` (checkmark), `ΓÜá` (warning), `≡ƒö¼` (magnifying glass), `ΓåÆ` (arrow). Run the same find/replace here that cleaned up the .cs files.

shrja-ms: Fixed. Replaced 112 mojibake chars in architecture.md (em dashes, arrows, checkmarks, warnings, bullets). Box-drawing chars for directory trees are preserved.

shrja-ms: Fixed 112 mojibake chars in architecture.md. Box-drawing chars preserved.

jongio (commented):
Fifth pass (incremental). The architecture.md mojibake I flagged last round is fully cleaned up - zero garbled characters remain. All prior findings from my earlier reviews have been addressed. Nothing new to flag.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ProtectedItem/ProtectedItemProtectCommand.cs
```diff
@@ -86,6 +86,11 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
                 options.VaultType,
                 options.Container,
                 options.DatasourceType,
+                options.AksIncludedNamespaces,
```
jongio, line 94: The 5 AKS options (`AksSnapshotResourceGroup`, `AksIncludedNamespaces`, `AksExcludedNamespaces`, `AksLabelSelectors`, `AksIncludeClusterScopeResources`) are passed here but will always be `null`. They were removed from `PolicyCreateCommand.RegisterOptions()`/`BindOptions()` but never added to this command's `RegisterOptions` or `BindOptions`.

Add to `RegisterOptions`:

```csharp
command.Options.Add(AzureBackupOptionDefinitions.AksSnapshotResourceGroup);
command.Options.Add(AzureBackupOptionDefinitions.AksIncludedNamespaces);
command.Options.Add(AzureBackupOptionDefinitions.AksExcludedNamespaces);
command.Options.Add(AzureBackupOptionDefinitions.AksLabelSelectors);
command.Options.Add(AzureBackupOptionDefinitions.AksIncludeClusterScopeResources);
```

And add to `BindOptions`:

```csharp
options.AksSnapshotResourceGroup = parseResult.GetValueOrDefault<string>(AzureBackupOptionDefinitions.AksSnapshotResourceGroup.Name);
options.AksIncludedNamespaces = parseResult.GetValueOrDefault<string>(AzureBackupOptionDefinitions.AksIncludedNamespaces.Name);
options.AksExcludedNamespaces = parseResult.GetValueOrDefault<string>(AzureBackupOptionDefinitions.AksExcludedNamespaces.Name);
options.AksLabelSelectors = parseResult.GetValueOrDefault<string>(AzureBackupOptionDefinitions.AksLabelSelectors.Name);
options.AksIncludeClusterScopeResources = parseResult.GetValueOrDefault<string>(AzureBackupOptionDefinitions.AksIncludeClusterScopeResources.Name);
```

jongio (commented):
The AKS parameter move from `policy create` to `protecteditem protect` is architecturally correct - these are per-backup-instance settings, not policy-level. However, the options registration and binding didn't follow. The 5 AKS options on `ProtectedItemProtectOptions` will always be `null` because they're never registered as CLI options and never bound from the parse result.

jongio (commented):
AKS option registration and binding in `ProtectedItemProtectCommand` is now correct - all 5 options are registered, bound, and passed through to the service. My prior finding is addressed. The refactor cleanly separates policy-level settings from per-backup-instance settings.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> ScheduleFrequency = new($"--{ScheduleFrequencyName}")
+    {
+        Description = "Backup schedule frequency. RSV vaults accept 'Daily', 'Weekly', or 'Hourly'. DPP (Backup) vaults accept ISO 8601 intervals: 'PT4H', 'PT6H', 'PT8H', 'PT12H', 'P1D', 'P1W', 'P2W', or 'P1M'.",
+        Required = false
+    };
+
+    public static readonly Option<string> ScheduleTimes = new($"--{ScheduleTimesName}")
```
alzimmermsft, line 255: Does the service API use a comma-separated list of times for this?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+
+    public static readonly Option<string> WeeklyRetentionWeeks = new($"--{WeeklyRetentionWeeksName}")
+    {
+        Description = "Number of weeks to keep weekly recovery points. Pair with --weekly-retention-days-of-week.",
+        Required = false
+    };
+
+    public static readonly Option<string> WeeklyRetentionDaysOfWeek = new($"--{WeeklyRetentionDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week tagged for weekly retention (e.g., 'Sunday' or 'Saturday,Sunday'). Pair with --weekly-retention-weeks.",
```
alzimmermsft: When you say, "Pair with", does that mean when one option is used the other is required?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+
+    public static readonly Option<string> ScheduleDaysOfWeek = new($"--{ScheduleDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week the backup should run (e.g., 'Monday,Wednesday,Friday'). Required for Weekly schedules.",
+        Required = false
+    };
+
+    public static readonly Option<string> HourlyIntervalHours = new($"--{HourlyIntervalHoursName}")
+    {
+        Description = "Interval in hours between hourly backups (e.g., 4, 6, 8, 12). Used only when --schedule-frequency is 'Hourly' (RSV).",
```
alzimmermsft: Any validation range for the number of hours?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+
+    public static readonly Option<string> IsCompression = new($"--{IsCompressionName}")
+    {
+        Description = "Enable backup compression at the policy level: 'true' or 'false'. RSV VmWorkload only.",
+        Required = false
+    };
+
+    public static readonly Option<string> IsSqlCompression = new($"--{IsSqlCompressionName}")
+    {
+        Description = "Enable SQL-native backup compression: 'true' or 'false'. RSV SQL only.",
```
alzimmermsft: Does this support multiple forms of SQL or just PostgreSQL?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> HourlyWindowStartTime = new($"--{HourlyWindowStartTimeName}")
+    {
+        Description = "Start time of the hourly backup window in 24h HH:mm format (e.g., '08:00'). Used only when --schedule-frequency is 'Hourly' (RSV).",
+        Required = false
+    };
+
+    public static readonly Option<string> HourlyWindowDurationHours = new($"--{HourlyWindowDurationHoursName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> ScheduleDaysOfWeek = new($"--{ScheduleDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week the backup should run (e.g., 'Monday,Wednesday,Friday'). Required for Weekly schedules.",
+        Required = false
+    };
+
+    public static readonly Option<string> HourlyIntervalHours = new($"--{HourlyIntervalHoursName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> HourlyWindowDurationHours = new($"--{HourlyWindowDurationHoursName}")
+    {
+        Description = "Duration of the hourly backup window in hours (e.g., 12). Used only when --schedule-frequency is 'Hourly' (RSV).",
+        Required = false
+    };
+
+    public static readonly Option<string> WeeklyRetentionWeeks = new($"--{WeeklyRetentionWeeksName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> WeeklyRetentionDaysOfWeek = new($"--{WeeklyRetentionDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week tagged for weekly retention (e.g., 'Sunday' or 'Saturday,Sunday'). Pair with --weekly-retention-weeks.",
+        Required = false
+    };
+
+    public static readonly Option<string> MonthlyRetentionMonths = new($"--{MonthlyRetentionMonthsName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Number of months to keep monthly recovery points. Combine with either --monthly-retention-days-of-month (absolute) OR --monthly-retention-week-of-month + --monthly-retention-days-of-week (relative).",
+        Required = false
+    };
+
+    public static readonly Option<string> MonthlyRetentionWeekOfMonth = new($"--{MonthlyRetentionWeekOfMonthName}")
+    {
+        Description = "Which week of the month to tag for monthly retention: 'First', 'Second', 'Third', 'Fourth', or 'Last'. Use with --monthly-retention-days-of-week (relative scheme).",
```
alzimmermsft, line 305: Is this how the REST API takes this parameter too?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> MonthlyRetentionDaysOfMonth = new($"--{MonthlyRetentionDaysOfMonthName}")
+    {
+        Description = "Comma-separated days of the month for monthly retention (1-28 or 'Last'; e.g., '1,15,Last'). Absolute scheme; mutually exclusive with --monthly-retention-week-of-month.",
+        Required = false
+    };
+
+    public static readonly Option<string> YearlyRetentionYears = new($"--{YearlyRetentionYearsName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> DifferentialScheduleDaysOfWeek = new($"--{DifferentialScheduleDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week for the Differential backup (e.g., 'Monday,Thursday'). RSV VmWorkload only.",
+        Required = false
+    };
+
+    public static readonly Option<string> DifferentialRetentionDays = new($"--{DifferentialRetentionDaysName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> IncrementalScheduleDaysOfWeek = new($"--{IncrementalScheduleDaysOfWeekName}")
+    {
+        Description = "Comma-separated days of the week for the Incremental backup. RSV SAPHANA / SAPASE only.",
+        Required = false
+    };
+
+    public static readonly Option<string> IncrementalRetentionDays = new($"--{IncrementalRetentionDaysName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Retention period in days for Incremental backups. RSV SAPHANA / SAPASE only.",
+        Required = false
+    };
+
+    public static readonly Option<string> LogFrequencyMinutes = new($"--{LogFrequencyMinutesName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Transaction log backup frequency in minutes (e.g., 15, 30, 60). RSV VmWorkload only.",
+        Required = false
+    };
+
+    public static readonly Option<string> LogRetentionDays = new($"--{LogRetentionDaysName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Retention period in days for transaction log backups. RSV VmWorkload only.",
+        Required = false
+    };
+
+    public static readonly Option<string> IsCompression = new($"--{IsCompressionName}")
```
alzimmermsft: Why is this a `string` and not a `bool`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Enable backup compression at the policy level: 'true' or 'false'. RSV VmWorkload only.",
+        Required = false
+    };
+
+    public static readonly Option<string> IsSqlCompression = new($"--{IsSqlCompressionName}")
```
alzimmermsft: Why is this a `string` and not a `bool`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Enable SQL-native backup compression: 'true' or 'false'. RSV SQL only.",
+        Required = false
+    };
+
+    // ===== Stage 2 expansion =====
+
+
+    public static readonly Option<string> SmartTier = new($"--{SmartTierName}")
```
alzimmermsft: Why is this a `string` and not a `bool`? Also, why make this a separate parameter from `--archieve-tier-mode`? Shouldn't that be able to accept `TierRecommended` to infer this parameter's handling.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+    {
+        Description = "Enable smart-tiering (ML-based archive recommendation): 'true' or 'false'. RSV VM only  -  equivalent to TieringMode=TierRecommended.",
+        Required = false
+    };
+
+    public static readonly Option<string> EnableSnapshotBackup = new($"--{EnableSnapshotBackupName}")
```
alzimmermsft: Why is this a `string` and not a `bool`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> SnapshotInstantRpResourceGroup = new($"--{SnapshotInstantRpResourceGroupName}")
+    {
+        Description = "Resource group prefix for snapshot instant RPs. RSV SAPHANA snapshot only.",
+        Required = false
+    };
+
+    public static readonly Option<string> EnableVaultTierCopy = new($"--{EnableVaultTierCopyName}")
```
alzimmermsft: Why is this a `string` and not a `bool`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> BackupMode = new($"--{BackupModeName}")
+    {
+        Description = "Backup mode for storage workloads: 'Continuous' (default for AzureBlob, ADLS) or 'Vaulted' (discrete recovery points). DPP AzureBlob, AzureDataLakeStorage.",
+        Required = false
+    };
+
+    public static readonly Option<string> PitrRetentionDays = new($"--{PitrRetentionDaysName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> EnableVaultTierCopy = new($"--{EnableVaultTierCopyName}")
+    {
+        Description = "Enable vault-tier copy of operational store backups: 'true' or 'false'. DPP AzureDisk only.",
+        Required = false
+    };
+
+    public static readonly Option<string> VaultTierCopyAfterDays = new($"--{VaultTierCopyAfterDaysName}")
```
alzimmermsft: Why is this a `string` and not an `int`?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
```diff
@@ -175,9 +232,294 @@ public static class AzureBackupOptionDefinitions
...
+        Required = false
+    };
+
+    public static readonly Option<string> AksLabelSelectors = new($"--{AksLabelSelectorsName}")
+    {
+        Description = "Comma-separated label selectors (e.g. 'app=frontend,tier=web') applied to the AKS backup policy default scope. DPP AKS only.",
+        Required = false
+    };
+
+    public static readonly Option<string> AksIncludeClusterScopeResources = new($"--{AksIncludeClusterScopeResourcesName}")
```
alzimmermsft: Why is this a `string` and not a `bool`?

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
```diff
@@ -0,0 +1,300 @@
...
+        {
+            rules.Add(BuildTierRetentionRule("Weekly", TimeSpan.FromDays(weeks * 7), tierStore, request));
+        }
+        if (TryParsePositiveInt(request.MonthlyRetentionMonths, out var months))
+        {
+            rules.Add(BuildTierRetentionRule("Monthly", TimeSpan.FromDays(months * 30), tierStore, request));
+        }
+        if (TryParsePositiveInt(request.YearlyRetentionYears, out var years))
+        {
+            rules.Add(BuildTierRetentionRule("Yearly", TimeSpan.FromDays(years * 365), tierStore, request));
```
alzimmermsft: These concern me as a month isn't always 30 days and a year isn't always 365 days

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
```diff
@@ -0,0 +1,300 @@
...
+            IsDefault = true,
+        };
+    }
+    private static DataProtectionRetentionRule BuildVaultTierCopyRetentionRule(PolicyCreateRequest request)
+    {
+        // Default retention for the vault-tier rule is 30 days unless --vault-tier-copy-after-days specified.
+        var retentionDays = TryParsePositiveInt(request.VaultTierCopyAfterDays, out var d) ? d : 30;
+
+        var lifeCycle = new SourceLifeCycle(
+            new DataProtectionBackupAbsoluteDeleteSetting(TimeSpan.FromDays(retentionDays)),
```
alzimmermsft: Would it make more sense for `null` to be allowed here to represent a default rather than coding it in? That way the service can handle it with whatever the default is.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
```diff
@@ -0,0 +1,300 @@
...
+            Criteria = { criteria },
+        };
+    }
+
+    private static DateTimeOffset ParseScheduleStartTime(string? scheduleTimes)
+    {
+        var raw = scheduleTimes;
+        if (string.IsNullOrWhiteSpace(raw))
+        {
+            raw = "02:00";
```
alzimmermsft, line 262: Is it mentioned if this isn't set the default is 2am?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
shrja-ms: No - the service API takes a single datetime per schedule. Our CLI parses the comma-separated input client-side and uses only the first time as the schedule start time. Updated the description to clarify: 'Defaults to 02:00 UTC if not specified. Only the first time is used as the schedule start time.'

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
shrja-ms: Valid values are 4, 6, 8, 12 (RSV API constraint). Updated description. Changed type from Option string to Option int.

jongio (commented):
Eighth pass after rebase. Code hasn't changed from my last review - all prior findings still addressed. @alzimmermsft's review covers the key open items now. I've added data to a few of those threads below.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/Policy/DppPolicyBuilder.cs
jongio: These aren't doing calendar math - they're building the ISO 8601 duration values that the DPP API expects in its retention policy model. Azure Backup uses `TimeSpan`-based day counts where month=30 and year=365 are the standard API approximations. A brief code comment explaining this convention would prevent the confusion.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/AzureBackupOptionDefinitions.cs
jongio: Typed options are already used across the repo. `SqlOptionDefinitions` uses `Option<int>`, `Option<bool>`, `Option<long>`. `StorageOptionDefinitions` and `KeyVaultOptionDefinitions` use `Option<bool>`. Core `OptionDefinitions.cs` has `Option<double>`, `Option<int>`, `Option<AuthMethod>`. `GetValueOrDefault<T>` supports `bool`, `int`, `long`, `double`, and custom enums. Using `Option<string>` for numeric/boolean values here is inconsistent with the codebase pattern.

servers/Azure.Mcp.Server/docs/azmcp-commands.md
jongio: For context: `PolicyCreateCommand` registers 46 CLI options. The next largest command in the repo is `DiskCreateCommand` with 23. So this is roughly 2x the current repo maximum. Whether that means subcommands (`policy create-rsv` / `policy create-dpp`), a multi-step flow, or just accepting the complexity depends on how agents handle the context window cost in practice. But 46 options is an outlier worth discussing.

alzimmermsft (approved):

PR #2510, opened 2026-04-27
Add Performance Scenarios + Pipeline Infra

### LINKED ISSUE #3118: Measure startup and tool discovery performance
## Outcome

Azure MCP startup and tool discovery remain responsive as packages, namespaces, and the exposed tool surface grow.

## In scope

- Measure cold and warm process startup for stdio and remote HTTP modes
- Measure server readiness and first successful request
- Measure tool and namespace discovery at current and projected tool counts
- Separate process, dependency injection, registration, serialization, and client-handshake time
- Track scaling behavior as the tool catalog grows

## Out of scope

- Semantic tool-selection quality
- Downstream Azure service execution latency
- Host installation and package download time

## Acceptance criteria and budgets

- [ ] Controlled baselines exist for cold startup, warm startup, readiness, and discovery
- [ ] p50 and p95 may not regress by more than 10% from the approved baseline
- [ ] p99 may not regress by more than 20%
- [ ] Discovery scaling efficiency may not degrade by more than 15% at the defined projected tool count
- [ ] Changes above 5% are reported as warnings with comparison evidence
- [ ] Absolute p95 and p99 targets are ratcheted after the baseline qualification period

## Dependencies and risks

- Controlled workload and benchmark environment definitions
- Stable generated tool catalogs and server modes
- Performance Epic #2412

## Evidence for closure

- Benchmark implementation
- Baseline results and trend report
- Documented absolute targets after qualification

g2vinay:
Adds a performance measurement suite for the Azure MCP Server and wires it into the internal AzDO perf pipeline (part of the Performance epic #2412). Four areas, each with an end-to-end harness, a tiered baseline-gated regression check, and BenchmarkDotNet micro-benchmarks where useful:

1. Startup & discovery (#3118) — CLI cold-start and stdio/HTTP initialize → tools/list timing across default/namespace/all modes, GPT‑4o token budgets, and tool-surface scaling.
2. Resource usage (#3121) — CPU/memory/handle sampling across idle, startup, discovery, steady-state, and soak phases, with least-squares leak detection.
3. Dispatch & transport overhead (#3119) — per-layer dispatch micro-benchmarks over a deterministic no-op command, plus a stdio-vs-HTTP tools/list harness isolating transport cost.
4. Concurrency, throughput & scalability (#3122) — a concurrency sweep reporting throughput, latency percentiles, error rate, scaling efficiency, and a state-isolation signal.

5. Workload catalog & run metadata (#3120) — versioned perf-workloads.json and a self-describing environment block (commit, runtime, OS, CPU, memory, catalog version) on every result.
6. Gates use tiered budgets and fail closed on incompatible/empty baselines and metrics the baseline expects but the results drop; covered by 75 Pester tests.
7. Harnesses run through a shared bounded-process helper (timeout + stderr + exit-code checks); the pipeline calls standalone scripts (no inline YAML PowerShell).

Closes #3118 #3121 #3119 #3120 #3122


eng/scripts/Test-StartupPerformance.ps1
```diff
@@ -0,0 +1,321 @@
...
+#     full_schema:       { bytes, exact_tokens_gpt4o_o200k, approx_tokens_bytes_div_4 } }
+# ---------------------------------------------------------------------------
+function Invoke-BenchmarkMcpStartup {
+    param(
+        [string]   $BenchmarkExe,
+        [string]   $ExePath,
+        [string[]] $ServerArgTokens
+    )
+
+    $output  = & $BenchmarkExe --mcp-startup $ExePath @ServerArgTokens 2>$null
```
jongio: `2>$null` discards all diagnostic output from the benchmark exe. If the process crashes (missing dependency, runtime error), stderr is lost and the next line tries `ConvertFrom-Json` on empty output, producing a confusing error instead of the real failure.

Capture stderr and check the exit code instead - something like `2>&1` plus a `LASTEXITCODE` check before attempting JSON parse.

eng/scripts/Test-StartupPerformance.ps1
```diff
@@ -0,0 +1,321 @@
...
+        UseShellExecute        = $false
+        CreateNoWindow         = $true
+    }
+
+    $sw   = [System.Diagnostics.Stopwatch]::StartNew()
+    $proc = [System.Diagnostics.Process]::Start($psi)
+    # Drain both streams asynchronously to prevent buffer deadlocks
+    $stdoutTask = $proc.StandardOutput.ReadToEndAsync()
+    $stderrTask = $proc.StandardError.ReadToEndAsync()
+    $proc.WaitForExit()
```
jongio: `Measure-CliColdStart` records wall-clock time without checking `ExitCode`. If `azmcp tools list` fails (auth prompt, missing config), you get a misleadingly fast timing recorded as a valid sample, which corrupts median and regression comparisons.

Add an exit code check after `WaitForExit()` before returning elapsed time.

jongio (commented):
A few things beyond what the automated review already flagged.

The overall design is clean - BenchmarkDotNet for micro-benchmarks, MCP SDK for e2e timing, and token counting for LLM context budget tracking. The pipeline integration with optional regression gating is well-structured.

Main gaps: error handling in the PS helpers (stderr swallowed, exit codes unchecked), the `FullStartup` benchmark name is slightly misleading since it doesn't include `RegisterAreas`, and two cspell entries landed out of alphabetical order.

On the bot's first comment about missing the server build - the benchmark csproj has a `ProjectReference` to the server, so `dotnet build` of the benchmark project transitively builds the server too. The azmcp.exe will be at the expected path.

servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs
```diff
@@ -0,0 +1,95 @@
...
+        Program.InitializeServicesAsync(sp).GetAwaiter().GetResult();
+        return sp.GetRequiredService<ICommandFactory>().RootCommand;
+    }
+
+    // -------------------------------------------------------------------------
+    // Full pipeline – all phases combined (closest to real startup cost for
+    // the first DI container in Program.Main)
+    // -------------------------------------------------------------------------
+
+    [Benchmark(Description = "FullStartup – all phases combined (DI build → command tree)")]
```
jongio: `FullStartup` is described as 'all phases combined' but it doesn't include `RegisterAreas`. That phase runs once as a static initializer (`Program.cs:30`), so by the time any benchmark method executes, area registration is already done and cached in the static `Areas` field.

Either call `RegisterAreas()` explicitly inside the benchmark body (and isolate from the static cache), or rename to clarify the measurement boundary.

.vscode/cspell.json
```diff
@@ -468,6 +470,7 @@
         "livetests",
         "llm",
         "loadtest",
+        "launchcount",
```
jongio: Two words landed out of alphabetical order:
- `launchcount` is placed after `loadtest` but should come before `livetests`  
- `memorydiagnoser` (a few lines below) is placed after `midsole` but should come before `microbundle`

eng/pipelines/templates/common.yml
```diff
@@ -95,6 +98,25 @@ extends:
           MaxParallel: 4
           HostArchitecture: arm64
 
+    - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.RunPerfTests, 'true')) }}:
+      - stage: Perf
+        displayName: 'Performance test'
+        dependsOn:
+        - Initialize
```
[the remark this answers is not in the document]
g2vinay: Depends on build

eng/scripts/Test-StartupPerformance.ps1
```diff
@@ -0,0 +1,321 @@
...
+    $sw.Stop()
+    $null = $stdoutTask.Result
+    $null = $stderrTask.Result
+    $proc.Dispose()
+    return $sw.ElapsedMilliseconds
```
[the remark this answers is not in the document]
g2vinay: Added a 30 s timeout (WaitForExit($timeoutMs)). On timeout: kills the process, captures stderr, throws with a descriptive message. On normal exit: checks ExitCode -ne 0 and throws with stderr before returning elapsed time.

jongio (commented):
Incremental review (7b7a6bc vs dc9d706): addresses all four comments from my earlier review.

- stderr handling in `Invoke-BenchmarkMcpStartup` - rewritten with proper `ProcessStartInfo`, timeout, exit code check, and structured errors. Solid.
- Exit code check in `Measure-CliColdStart` - now throws on non-zero exit. Timeout guard added too.
- `FullStartup` description - accurately reflects that `RegisterAreas` runs separately as a static initializer.
- cspell.json ordering - fixed.

Bonus fixes: `ServiceProvider` leak plugged via `[IterationCleanup]` + `using`/`await using`, division-by-zero guard in `ToolCountMeasurement`, and `ServerArgs` properly typed as `string[]` instead of splitting on whitespace.

No new issues in the incremental diff.

anannya03 (approved):

eng/pipelines/templates/jobs/perf-test.yml
```diff
@@ -0,0 +1,114 @@
...
+
+  # Regression gate — only active when BaselinePath is provided.
+  # Reads the already-written results JSON; does NOT re-run the benchmarks.
+  - ${{ if ne(parameters.BaselinePath, '') }}:
+    - task: Powershell@2
+      displayName: "Check regression against baseline"
+      inputs:
+        pwsh: true
+        targetType: inline
+        script: |
```
alzimmermsft, line 65: Let's get out of having PowerShell scripts inline in this YAML file, it will make replicating CI behaviors locally much more difficult.

eng/pipelines/templates/jobs/perf-test.yml
```diff
@@ -0,0 +1,114 @@
...
+          -c Release /consoleloggerparameters:NoSummary /nowarn:NU1902
+        if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
+  - task: Powershell@2
+    displayName: "Run performance measurements"
+    inputs:
+      pwsh: true
+      filePath: $(Build.SourcesDirectory)/eng/scripts/Test-StartupPerformance.ps1
+      arguments: >
+        -Executable '$(Build.SourcesDirectory)/servers/Azure.Mcp.Server/src/bin/Release/net10.0/azmcp.exe'
```
alzimmermsft, line 53: Long term we should make this be based on the MCP server being built in the CI job. Right now, if Fabric MCP wanted to leverage this they couldn't.

eng/pipelines/templates/jobs/perf-test.yml
```diff
@@ -0,0 +1,114 @@
...
+      script: |
+        $results = Get-Content '$(Build.ArtifactStagingDirectory)/perf/startup-e2e.json' | ConvertFrom-Json
+        $s = $results.scenarios
+
+        $metrics = @(
+          @{ name = 'startup.cli.cold.median_ms';         value = $s.cli_cold_start_ms.median }
+          @{ name = 'startup.mcp.default.median_ms';      value = $s.mcp_stdio_to_tools_list_ms.median }
+          @{ name = 'startup.mcp.namespace.median_ms';    value = $s.mcp_namespace_mode_startup_ms.median }
+          @{ name = 'startup.mcp.all.median_ms';          value = $s.mcp_all_mode_startup_ms.median }
+          @{ name = 'tokens.default.exact_gpt4o';         value = $s.tools_list_payload.exact_tokens_gpt4o_o200k }
```
alzimmermsft, line 101: How did we decide that we're using GPT 4o for this?

wbreza (commented):
### What stood out
The end-to-end harness measures the MCP initialize/tools-list handshake separately from tokenization. Three MEDIUM issues remain in statistics, phase isolation, and output-path handling.

### Top issues
- [F-001] MEDIUM: Compute the median correctly for even sample counts — `eng/scripts/Test-StartupPerformance.ps1:226`.
- [F-002] MEDIUM: Exclude preceding phases from phase-specific benchmarks — `servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs:55`.
- [F-003] MEDIUM: Accept filename-only output paths — `eng/scripts/Test-StartupPerformance.ps1:76`.

The existing [token-metric field mismatch](https://github.com/microsoft/mcp/pull/2510#discussion_r3151955814) also remains unresolved.

eng/scripts/Test-StartupPerformance.ps1
```diff
@@ -0,0 +1,390 @@
...
+
+# ---------------------------------------------------------------------------
+# Helper: compute first/average/median timing stats from a sample array
+# ---------------------------------------------------------------------------
+function Get-TimingStats {
+    param([long[]] $Samples)
+    return [ordered]@{
+        first   = $Samples[0]
+        average = [math]::Round(($Samples | Measure-Object -Average).Average, 1)
+        median  = ($Samples | Sort-Object)[[math]::Floor($Samples.Count / 2)]
```
wbreza: [F-001] MEDIUM — Average the two middle samples for even run counts

`Sort-Object` followed by index `Floor(Count / 2)` selects the upper middle sample when `-Runs` is even, rather than the median. For example, samples `[100, 130]` report 130 ms instead of 115 ms. Against a 100 ms baseline at the default 1.20 threshold, this falsely fails the regression gate even though the true median is below 120 ms. The documented `-Runs 10` usage takes this path as well. Average the two middle values for even counts, retain the middle value for odd counts, and cover both cases with regression tests.

servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs
```diff
@@ -0,0 +1,107 @@
...
+    public IServiceCollection ConfigureServices() => CreateSilentServices();
+
+    // -------------------------------------------------------------------------
+    // Phase 3 – compile the DI container (BuildServiceProvider)
+    // -------------------------------------------------------------------------
+
+    [Benchmark(Description = "BuildServiceProvider – compile the DI container")]
+    public ServiceProvider BuildServiceProvider()
+    {
+        _lastBuiltProvider = CreateSilentServices().BuildServiceProvider();
```
wbreza: [F-002] MEDIUM — Move prerequisite phases outside the measured operation

The phase-3 benchmark calls `CreateSilentServices()` inside its timed body, so it measures all service registration as well as provider construction. The same contamination continues in `InitializeServicesAsync` (registration + provider construction + initialization + disposal) and `BuildCommandTree` (the full startup path + disposal). Consequently, a slowdown confined to `ConfigureServices` appears as a regression in every later phase, contrary to the class's stated isolation/attribution purpose; `BuildCommandTree` effectively repeats `FullStartup`. Prepare each phase's prerequisites in targeted setup hooks and dispose them in cleanup, leaving only the named operation in its benchmark body. Keep the cumulative measurement in `FullStartup`.

eng/scripts/Test-StartupPerformance.ps1
```diff
@@ -0,0 +1,390 @@
...
+        $Executable = Join-Path $RepoRoot 'servers/Azure.Mcp.Server/src/bin/Debug/net10.0/azmcp'
+    }
+}
+
+if (-not (Test-Path $Executable)) {
+    Write-Error "azmcp executable not found at: $Executable`nBuild first: dotnet build servers/Azure.Mcp.Server/src"
+}
+
+$OutputPath = if ($OutputPath) { $OutputPath } else { Join-Path $RepoRoot '.perf-results/startup-e2e.json' }
+$null = New-Item -ItemType Directory -Force -Path (Split-Path $OutputPath)
```
wbreza: [F-003] MEDIUM — Handle an output filename without a directory component

With valid built executables, invoking the script with `-OutputPath startup-e2e.json` still fails here: `Split-Path` returns an empty string, and `New-Item -Path ''` throws `Cannot bind argument to parameter 'Path' because it is an empty string`. This is a valid relative file path, but the script exits before collecting any samples. Normalize the output path to an absolute path before extracting its parent, or create the parent directory only when one is present. Cover filename-only, relative-directory, and absolute output paths.

eng/scripts/Measure-DispatchOverhead.ps1
```diff
@@ -0,0 +1,166 @@
...
+$Executable = (Resolve-Path $Executable).Path
+$BenchmarkExe = (Resolve-Path $BenchmarkExe).Path
+$OutputPath = Resolve-PerfOutputPath -Path $OutputPath
+
+$env:PERF_DISPATCH_ITERATIONS = $Iterations
+$env:PERF_DISPATCH_WARMUP = $Warmup
+
+function Get-DispatchStats {
+    param([double[]] $Samples)
+    $s = Get-TimingStats -Samples $Samples
```
wbreza, line 76: [F-004] MEDIUM — Preserve fractional milliseconds through dispatch statistics

`Get-DispatchStats` receives `double[]`, but `Get-TimingStats` and `Get-Percentile` take `long[]`. Binding therefore rounds every sample before computing percentiles: `[0.41, 0.49, 0.51, 0.59]` becomes `[0, 0, 1, 1]`, and p95 is reported as 1 ms instead of 0.578 ms. This discards the four-decimal precision emitted by `DispatchOverheadMeasurement` and can mask or manufacture changes far larger than the 5–10% budgets. Keep floating-point samples and sufficient precision throughout the calculations; apply display rounding only after the gate comparisons. Add sub-millisecond regression cases.

servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs
```diff
@@ -0,0 +1,153 @@
...
+            AddDiagnoser(MemoryDiagnoser.Default);
+        }
+    }
+
+    [IterationCleanup]
+    public void DisposeProviders()
+    {
+        _builtProvider?.Dispose();
+        _providerForInitialize?.Dispose();
+        _providerForCommandTree?.Dispose();
```
wbreza: [F-005] MEDIUM — Dispose the command-tree provider asynchronously

`BuildCommandTree` resolves `ICommandFactory`, whose registration walk resolves `CosmosListCommand` through `CosmosSetup.RegisterCommands`, and consequently the singleton `ICosmosService`. Its implementation, `CosmosService`, implements `IAsyncDisposable` but not `IDisposable` (nor does `BaseAzureService`). Calling `ServiceProvider.Dispose()` here throws for that resolved async-only service, so the command-tree benchmark fails during iteration cleanup rather than completing its measurements. Make the cleanup hook asynchronous and await `DisposeAsync()` on its providers, keeping disposal outside the timed body.

eng/scripts/Measure-DispatchOverhead.ps1
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+$scenarios = [ordered]@{}
+
+foreach ($transport in @('stdio', 'http')) {
+    Write-Host "=== Dispatch overhead: $transport ($Iterations iterations) ==="
+    $mode = "--mcp-dispatch-$transport"
+
+    $stdout = & $BenchmarkExe $mode $Executable server start 2>$null
```
wbreza: [F-006] MEDIUM — Reuse bounded, exit-code-checked harness execution

Unlike the startup runner, the dispatch, concurrency (`Measure-ConcurrencyScaling.ps1:79`), and resource (`Measure-ResourceUsage.ps1:106`) runners invoke the executable directly, discard stderr, and never inspect its exit code. With PowerShell's default native-error behavior, a process emitting valid JSON and then exiting 7 is accepted as a successful measurement. This is relevant because the C# dispatch harness emits its result before awaited client cleanup, which can still fail. A hung harness also has no outer timeout here, consuming the entire CI job timeout or waiting indefinitely locally. Reuse the startup runner's bounded `Process` execution pattern, preserving stderr and rejecting nonzero exits before parsing JSON.

eng/scripts/StartupPerformance.Common.ps1
```diff
@@ -0,0 +1,328 @@
...
+        if ($null -eq $curStats -or $null -eq $baseStats) {
+            Write-Host ("  [SKIP] {0}: metric absent in results or baseline" -f $name)
+            continue
+        }
+
+        foreach ($budget in $percentileBudgets) {
+            $metric   = $budget.Metric
+            $current  = Get-MemberValue $curStats $metric
+            $baseline = Get-MemberValue $baseStats $metric
+            if ($null -eq $current -or $null -eq $baseline) {
```
wbreza: [F-007] MEDIUM — Do not pass a gate that compared no measurements

Every missing percentile is skipped, and the gate returns empty failures even when all comparisons were skipped. A baseline produced by the previous version contains `median` and `all` but no p50/p95/p99; comparing its 100 ms samples against current 1000 ms samples prints only SKIP lines and then `All budget checks passed.` The dispatch, resource, and concurrency gates likewise accept inputs with no comparable metrics. Track evaluated checks and reject an incompatible/empty comparison instead of reporting success; for older startup baselines, derive percentiles from stored samples or require an explicit migration. Missing current metrics that the baseline expects also need an explicit failure/incomplete outcome.

eng/scripts/ConcurrencyPerformance.Common.ps1
```diff
@@ -0,0 +1,138 @@
...
+
+    $baselineByConcurrency = @{}
+    foreach ($bl in $BaselineLevels) {
+        $baselineByConcurrency[[string](Get-MemberValue $bl 'concurrency')] = $bl
+    }
+
+    foreach ($level in $ResultLevels) {
+        $concurrency = Get-MemberValue $level 'concurrency'
+        $baseline = $baselineByConcurrency[[string]$concurrency]
+        if ($null -eq $baseline) {
```
wbreza: [F-008] MEDIUM — Evaluate unconditional failures before baseline lookup

The early `continue` bypasses the `tool_count_mismatches` check below, despite that check being documented as a hard failure regardless of baseline. For example, extending a baseline containing concurrency 1 to a measured level 16 with five mismatches returns no failures for level 16. `Invoke-ResourceRegressionGate` has the same ordering at lines 59–61: an added `all` mode with `soak.unbounded_growth=true` is skipped if the baseline only contains `default`. Check these baseline-independent invariants for every current result first; only skip the relative budget comparisons when the matching baseline is absent.

eng/scripts/StartupPerformance.Common.ps1
```diff
@@ -0,0 +1,328 @@
...
+# Captures the controlled-environment metadata every perf result must record (issue #3120):
+# commit, runtime, OS, CPU, machine memory, plus the workload catalog version. Harnesses embed
+# this under an "environment" key so results are self-describing and comparable across runs.
+# Memory is best-effort and cross-platform; unresolved values are null rather than fatal.
+# ---------------------------------------------------------------------------
+function Get-PerfRunMetadata {
+    [CmdletBinding()]
+    param()
+
+    $commit = try { (& git rev-parse --short HEAD 2>$null).Trim() } catch { '' }
```
wbreza: [F-009] LOW — Resolve commit metadata independently of the caller's directory

`git rev-parse` runs in the current working directory, not the repository containing the measurement script. Invoking the script by absolute path from another checkout stamps that checkout's commit; from outside a repository it records an empty commit. The startup runner already computes its top-level commit with `git -C $RepoRoot`, so the same result can contain conflicting top-level and `environment.commit` values. Pass the resolved repository context into this helper and use `git -C`; apply the same correction to the new runners' top-level commit reads.

wbreza (commented):
## Re-review — full (force-push detected)



**Resolved (3):** F-001, F-002, F-003

**New (7):**
- MEDIUM F-004 — `eng/scripts/Measure-DispatchOverhead.ps1:71` — fractional dispatch samples are rounded to whole milliseconds.
- MEDIUM F-005 — `servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs:60` — synchronous cleanup rejects async-only services.
- MEDIUM F-006 — `eng/scripts/Measure-DispatchOverhead.ps1:88` — failed or hung harness processes are not handled correctly.
- MEDIUM F-007 — `eng/scripts/StartupPerformance.Common.ps1:184` — regression gates can pass without comparing any measurements.
- MEDIUM F-008 — `eng/scripts/ConcurrencyPerformance.Common.ps1:78` — missing baselines bypass unconditional failure checks.
- LOW F-009 — `eng/scripts/StartupPerformance.Common.ps1:301` — commit metadata depends on the caller's working directory.
- LOW F-010 — `eng/scripts/ResourcePerformance.Common.ps1:88` — positive CPU usage against a zero baseline reports no regression.

**To approve:** resolve F-004, F-005, F-006, F-007, F-008.

eng/scripts/ResourcePerformance.Common.ps1
```diff
@@ -0,0 +1,122 @@
...
+               Limit = $FailPeak }
+        )
+
+        foreach ($c in $checks) {
+            if ($null -eq $c.Cur -or $null -eq $c.Base) {
+                Write-Host ("  [SKIP] {0} {1}: value absent" -f $mode, $c.Label)
+                continue
+            }
+
+            $ratio = if ($c.Base -gt 0) { $c.Cur / $c.Base } else { 1 }
```
wbreza: [F-010] LOW — Handle a zero baseline without reporting a zero-percent change

The fallback ratio of 1 makes every positive current value pass when the baseline is zero. For example, steady-state CPU of 25% against a 0% baseline prints `[PASS] ... (+0%)` and returns no warning or failure. Treat zero-to-positive measurements explicitly (using a defined absolute budget, failure, or a non-comparable-baseline error); reserve unchanged status for zero-to-zero. The startup and dispatch ratio calculations use the same fallback.

wbreza (commented):
## Re-review — 1 commit since `6b8caec`



The build command now preserves project-specific warning suppressions. No new findings in this delta.

**Still open (7):**
- MEDIUM F-004 — `eng/scripts/Measure-DispatchOverhead.ps1:71`
- MEDIUM F-005 — `servers/Azure.Mcp.Server/perf/Azure.Mcp.Server.PerformanceBenchmarks/Benchmarks/ServerStartupBenchmarks.cs:60`
- MEDIUM F-006 — `eng/scripts/Measure-DispatchOverhead.ps1:88`
- MEDIUM F-007 — `eng/scripts/StartupPerformance.Common.ps1:184`
- MEDIUM F-008 — `eng/scripts/ConcurrencyPerformance.Common.ps1:78`
- LOW F-009 — `eng/scripts/StartupPerformance.Common.ps1:301`
- LOW F-010 — `eng/scripts/ResourcePerformance.Common.ps1:88`

**To approve:** resolve F-004, F-005, F-006, F-007, F-008.

g2vinay (commented):
Thanks for the thorough review! Pushed commit `43dc4efeb` addressing all the open feedback. Here's what changed for each item, in plain terms:

### wbreza findings

- **F-004 (dispatch precision):** The shared timing/percentile helpers were typed `[long[]]`, so sub-millisecond dispatch samples were rounded to whole milliseconds before percentiles were computed. They now take `[double[]]` and keep 4-decimal precision, so a p95 of `0.578 ms` stays `0.578 ms` instead of becoming `1 ms`. Added Pester cases for sub-ms samples.
- **F-005 (async disposal):** `ServerStartupBenchmarks` now disposes its service providers with `await DisposeAsync()` in an async `[IterationCleanup]`, so async-only-disposable services (like `CosmosService`, reached via `BuildCommandTree`) no longer throw during cleanup. Verified with a clean Release rebuild.
- **F-006 (harness execution):** Added a shared `Invoke-PerfHarnessProcess` / `Invoke-PerfHarnessJson` helper (hard timeout, captured stderr, non-zero exit rejected *before* parsing JSON) and moved the dispatch, concurrency, and resource runners onto it. The startup runner now uses the same helper too, so there's a single bounded-exec path instead of `2>$null`.
- **F-007 (empty comparisons):** All four gates now count how many metrics were actually compared and fail with `no_comparable_metrics` when none were — an old/empty/incompatible baseline can no longer print "All budget checks passed." A metric the baseline expects but that's missing from the results is now an explicit failure as well.
- **F-008 (baseline-independent invariants):** The concurrency state-leakage check and the resource soak-leak check now run *before* the "absent in baseline" skip, so a new level/mode that isn't in the baseline can no longer bypass those hard failures.
- **F-009 (commit metadata):** `Get-PerfRunMetadata` takes a `-RepoRoot` and uses `git -C`, and the runners pass their repo root, so the stamped commit always comes from the script's repository rather than the caller's working directory.
- **F-010 (zero baseline):** A positive current value against a zero baseline is now treated as non-comparable (a failure) instead of the old fallback ratio of `1` that reported a misleading `+0%` pass.

### Earlier threads

- **Build the server before running (copilot):** The benchmark csproj has a `ProjectReference` to the server, so building it in Release also builds `azmcp.exe` at the expected path — as jongio pointed out. No change needed.
- **`$ServerArgs` whitespace split (copilot):** Already handled earlier — server args are passed as a typed `string[]`, not split on whitespace.
- **Re-implemented repo-root resolution (copilot):** Left as-is for now; it resolves `$PSScriptRoot/../..` to a full path, and F-009 makes the commit/git resolution repo-relative. Happy to switch to the shared `common.ps1` pattern if you'd prefer.
- **Inline PowerShell in the YAML (alzimmermsft):** Moved the benchmark build and micro-benchmark steps into `eng/scripts/Build-PerfBenchmarks.ps1` and `eng/scripts/Invoke-PerfMicroBenchmarks.ps1` so CI behavior can be reproduced locally.
- **Generalize so other servers (e.g. Fabric) can use it (alzimmermsft):** Agree this is the right long-term direction. It's a larger change (parameterizing the built server + exe path per MCP server), so I'd prefer to do it as a dedicated follow-up rather than grow this PR — tracking it separately.

All 71 Pester tests pass, the perf project builds clean, and cspell is clean.

wbreza (commented):
## Re-review — 1 commit since `efbed919`



**Resolved (6):** F-004, F-005, F-006, F-008, F-009, F-010

**Still open (1):**
- MEDIUM F-007 — `eng/scripts/DispatchPerformance.Common.ps1:52`, `eng/scripts/ResourcePerformance.Common.ps1:103`, `eng/scripts/ConcurrencyPerformance.Common.ps1:107`

No new findings in this delta.

**To approve:** resolve F-007.

g2vinay (commented):
Good catch — you're right, F-007 was only half done. My previous commit handled the empty-baseline case and the innermost per-metric case (startup + dispatch), but the higher-level skip branches you pointed at still silently skipped a value the baseline has but the results dropped:

- `DispatchPerformance.Common.ps1:52` — the transport (and operation) level
- `ResourcePerformance.Common.ps1:103` — the per-check loop
- `ConcurrencyPerformance.Common.ps1:107` — the per-check loop

Fixed in `dcc869986`: each of those skip points now splits the condition —

- **baseline missing the entry** → `SKIP` (genuinely not comparable), and
- **baseline has it but results are missing it** → explicit `FAIL` (`… (missing in results)`).

So a results payload that drops a transport, mode, level, or metric that the baseline expects now fails the gate instead of quietly passing. Added a Pester case per gate (dispatch transport, resource steady metric, concurrency latency metric). All 74 tests pass and cspell is clean.

wbreza (commented):
## Re-review — 1 commit since `43dc4ef`



**Still open (1):**
- MEDIUM F-007 — `eng/scripts/ConcurrencyPerformance.Common.ps1:129–131`

No new findings in this delta.

**To approve:** resolve F-007.

wbreza (commented):
## Re-review — 1 commit since `dcc8699`



The cleanup simplification preserves asynchronous disposal. No new findings in this delta.

**Still open (1):**
- MEDIUM F-007 — `eng/scripts/ConcurrencyPerformance.Common.ps1:129–131`

**To approve:** resolve F-007.

g2vinay (commented):
Fixed in `eb29247a0`. That was the concurrency error-rate block (`error_rate_pct`), which still used the combined `$null -ne $cur -and $null -ne $base` guard — so a baseline that has `error_rate_pct` but results that dropped it were silently skipped. It now fails with `(missing in results)` in that case, matching the other per-metric checks. Added a Pester case.

I also audited every remaining null-guard skip across the four gates. The only other combined guards are the dispatch transport-overhead and startup scaling-efficiency blocks — both supplementary WARN-only checks derived from primitives (the transport p50s, and all-mode p50 + tool count). If a result drops one of those inputs, the underlying primitive already fails with `(missing in results)`, so the gate can't pass silently — but I'm happy to make those explicit too if you'd prefer.

Pester is 75/75 and cspell is clean.

wbreza (approved):
## Re-review — 1 commit since `e4eb53b`



**Resolved (1):** F-007

No new findings in this delta. All prior findings are resolved.

**Approved.**

PR #2535, opened 2026-04-29
Optimize: Redirect CLI log output from stdout to stderr (Optimizes Token usage for Agents).

g2vinay:
In CLI mode, azmcp was writing informational log messages (Azure SDK HTTP pipeline traces, etc.) to stdout alongside the JSON response. This inflated token consumption for AI agents and pipelines parsing stdout — benchmarks showed up to ~44.5% of stdout bytes were log noise, not JSON.

The issue was AddConsole() in Program.cs defaults LogToStandardErrorThreshold to LogLevel.None, routing all log levels to stdout.

This PR sets LogToStandardErrorThreshold = LogLevel.Trace so every log level routes exclusively to stderr. Stdout now contains only the JSON CommandResponse.

Impact:

No behavior change for interactive terminal users (stderr is visible by default)
AI agents and pipelines consuming stdout receive clean JSON only

### DIFF
--- a/servers/Azure.Mcp.Server/src/Program.cs
+++ b/servers/Azure.Mcp.Server/src/Program.cs
@@ -56,7 +56,7 @@ internal class Program
 
             services.AddLogging(builder =>
             {
-                builder.AddConsole();
+                builder.AddConsole(options => options.LogToStandardErrorThreshold = LogLevel.Trace);
                 builder.SetMinimumLevel(LogLevel.Information);
             });
 
--- a/servers/Azure.Mcp.Server/changelog-entries/1777452002350.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1777452002350.yaml
@@ -1,0 +1,3 @@
+changes:
+  - section: "Bugs Fixed"
+    description: "Fixed CLI mode log messages being written to stdout instead of stderr. Log output now goes exclusively to stderr, ensuring agents and pipelines consuming stdout receive clean JSON only. Use `2>&1` to see logs alongside output, or `2>file.log` to capture them separately."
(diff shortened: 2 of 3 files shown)


servers/Azure.Mcp.Server/src/Program.cs
```diff
@@ -56,7 +56,7 @@ private static async Task<int> Main(string[] args)
 
             services.AddLogging(builder =>
             {
-                builder.AddConsole();
+                builder.AddConsole(options => options.LogToStandardErrorThreshold = LogLevel.Trace);
```
jongio, line 59: Correct fix. `LogLevel.Trace` is the lowest level, so this routes everything to stderr.

Note that `Fabric.Mcp.Server/src/Program.cs:41` and `Template.Mcp.Server/src/Program.cs:39` have the same plain `AddConsole()` call. If those servers are also used in CLI mode, they'll have the same stdout pollution issue. Might be worth a follow-up PR to apply this fix there too.

jongio (commented):
Clean fix - `LogToStandardErrorThreshold = LogLevel.Trace` correctly routes all console log output to stderr, keeping stdout clean for JSON. Server mode is unaffected since `ServiceStartCommand` already calls `ClearProviders()` and only re-adds console logging under `--debug` (with stderr configured).

One thing to consider: `Fabric.Mcp.Server/src/Program.cs` and `Template.Mcp.Server/src/Program.cs` both have the identical plain `AddConsole()` without the stderr threshold. Worth fixing those in a follow-up (or this PR if scope allows) so the same issue doesn't persist in the other servers.

Docs and changelog look good.

servers/Azure.Mcp.Server/src/Program.cs
g2vinay: Template MCP is not consumed by users.
Fabric MCP, I'll check with the fabric team and open a follow up PR for that accordingly.

alzimmermsft (approved):

PR #2545, opened 2026-04-30
feat(kusto): add --chart-type option to render query results as images

kboom:
## Summary

Adds opt-in chart image rendering to the Azure MCP Kusto query and sample commands. When `--chart-type` is specified (`TimeSeries`, `Bar`, `Scatter`, or `Pie`), query results are validated against the requested chart type and rendered as a PNG using [ScottPlot 5.x](https://scottplot.net/) (MIT). The image is returned alongside the JSON results as an MCP `ImageContentBlock`, enabling vision-capable LLM clients (e.g. GitHub Copilot) to analyze telemetry visually without parsing raw JSON tables.

## Motivation

When an LLM consumes a Kusto query that returns hundreds of time-series rows, it is often forced to either truncate, summarize-by-eye, or render an ASCII table — all lossy. KQL has the `| render timechart …` operator that ADX clients use to draw charts, but the MCP tool today simply returns the underlying tabular data and ignores `render`.

This change lets the model (or the user, via the tool's input) explicitly ask for a chart rendering when it is the right shape for the question being asked, while leaving the existing JSON-only behavior untouched.

## How it works

- `--chart-type` is a single new `Option<string?>` on `QueryCommand` and `SampleCommand` (parsed via `Enum.TryParse<ChartType>`).
- When set, after the query executes, results are passed through `DataShapeValidator`.
- If valid, `ScottPlotChartRenderer` produces a PNG via `plot.GetImageBytes(width, height)`.
- The PNG is added to `CommandResponse.Images` (new `ResponseImage` record on the core response model).
- `CommandFactoryToolLoader` emits an MCP `ImageContentBlock` per image alongside the existing `TextContentBlock` JSON.

Fully opt-in: omitting `--chart-type` preserves today's JSON-only behavior.

## Changes

**Core (`core/Microsoft.Mcp.Core`):**
- New `ResponseImage` record + `Images` property on `CommandResponse`.
- `CommandFactoryToolLoader` emits `ImageContentBlock.FromBytes(...)` for each image.
- `ModelsJsonContext` registers the new types for AOT.

**Kusto toolset (`tools/Azure.Mcp.Tools.Kusto`):**
- New `Rendering/`: `ChartType`, `DataShapeValidator`, `IKustoChartRenderer`, `ScottPlotChartRenderer`.
- `--chart-type` option added; renderer registered in `KustoSetup`.
- `QueryCommand` and `SampleCommand` invoke the renderer when set.

**Build:**
- ScottPlot 5.1.58 added to `Directory.Packages.props`.
- Pre-existing `NU1010` fix: pinned `System.ClientModel 1.9.0` (referenced by `Azure.Mcp.Tools.Pricing` but missing from CPM).

**Docs / tests:**
- `azmcp-commands.md`, `e2eTestPrompts.md` updated.
- Changelog entry added.
- 185 Kusto unit tests pass.

## Backwards compatibility

Fully backwards compatible. No existing tool input shape changes; `--chart-type` is purely additive and optional.

## Invoking Livetests

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -91,7 +91,7 @@
     <PackageVersion Include="Npgsql" Version="10.0.1" />
     <PackageVersion Include="YamlDotNet" Version="16.3.0" />
     <PackageVersion Include="Yarp.ReverseProxy" Version="2.3.0" />
-    <PackageVersion Include="System.ClientModel" Version="1.9.0" />
+    <PackageVersion Include="ScottPlot" Version="5.1.58" />
     <PackageVersion Include="System.CommandLine" Version="2.0.6" />
     <PackageVersion Include="System.Formats.Asn1" Version="10.0.6" />
     <PackageVersion Include="System.IdentityModel.Tokens.Jwt" Version="8.17.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -92,6 +92,7 @@
     <PackageVersion Include="YamlDotNet" Version="16.3.0" />
     <PackageVersion Include="Yarp.ReverseProxy" Version="2.3.0" />
     <PackageVersion Include="ScottPlot" Version="5.1.58" />
+    <PackageVersion Include="System.ClientModel" Version="1.9.0" />
     <PackageVersion Include="System.CommandLine" Version="2.0.6" />
     <PackageVersion Include="System.Formats.Asn1" Version="10.0.6" />
     <PackageVersion Include="System.IdentityModel.Tokens.Jwt" Version="8.17.0" />
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
@@ -225,13 +225,22 @@ public sealed class CommandFactoryToolLoader(
             var jsonResponse = JsonSerializer.Serialize(commandResponse, ModelsJsonContext.Default.CommandResponse);
             var isError = commandResponse.Status < HttpStatusCode.OK || commandResponse.Status >= HttpStatusCode.Ambiguous;
 
+            var content = new List<ContentBlock>
+            {
+                new TextContentBlock { Text = jsonResponse }
+            };
+
+            if (commandResponse.Images != null)
+            {
+                foreach (var image in commandResponse.Images)
+                {
+                    content.Add(ImageContentBlock.FromBytes(image.Data, image.MimeType));
+                }
+            }
+
             return new CallToolResult
             {
-                Content = [
-                    new TextContentBlock {
-                        Text = jsonResponse
-                    }
-                ],
+                Content = content,
                 IsError = isError
             };
         }
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
@@ -225,10 +225,16 @@ public sealed class CommandFactoryToolLoader(
             var jsonResponse = JsonSerializer.Serialize(commandResponse, ModelsJsonContext.Default.CommandResponse);
             var isError = commandResponse.Status < HttpStatusCode.OK || commandResponse.Status >= HttpStatusCode.Ambiguous;
 
-            var content = new List<ContentBlock>
+            var content = new List<ContentBlock>();
+
+            // When images are present and the response is successful, suppress the JSON text
+            // block so vision-capable clients receive only the rendered image(s) and skip the
+            // raw data the image was meant to replace. Errors always include the JSON text.
+            var hasImages = commandResponse.Images != null && commandResponse.Images.Count > 0;
+            if (!hasImages || isError)
             {
-                new TextContentBlock { Text = jsonResponse }
-            };
+                content.Add(new TextContentBlock { Text = jsonResponse });
+            }
 
             if (commandResponse.Images != null)
             {
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
@@ -227,11 +227,10 @@ public sealed class CommandFactoryToolLoader(
 
             var content = new List<ContentBlock>();
 
-            // When images are present and the response is successful, suppress the JSON text
-            // block so vision-capable clients receive only the rendered image(s) and skip the
-            // raw data the image was meant to replace. Errors always include the JSON text.
-            var hasImages = commandResponse.Images != null && commandResponse.Images.Count > 0;
-            if (!hasImages || isError)
+            // The text JSON envelope is suppressed only when the command explicitly opted in via
+            // CommandResponse.OmitTextContent (e.g. an image is intended to fully replace raw data).
+            // Errors always include the JSON envelope so the caller can read status/message.
+            if (!commandResponse.OmitTextContent || isError)
             {
                 content.Add(new TextContentBlock { Text = jsonResponse });
             }
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
@@ -234,6 +234,12 @@ public sealed class CommandFactoryToolLoader(
             {
                 content.Add(new TextContentBlock { Text = jsonResponse });
             }
+            else if (commandResponse.Images is { Count: > 0 } && !string.IsNullOrEmpty(commandResponse.Message))
+            {
+                // Emit the message as a lightweight text hint so the model knows to look at
+                // the attached image(s) without receiving the full JSON data envelope.
+                content.Add(new TextContentBlock { Text = commandResponse.Message });
+            }
 
             if (commandResponse.Images != null)
             {
--- a/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
+++ b/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
@@ -29,8 +29,12 @@ public class CommandResponse
     /// <c>CallToolResult</c>, enabling vision-capable LLM clients to consume charts or
     /// visualizations directly without parsing raw data.
     /// </summary>
-    [JsonPropertyName("images")]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    /// <remarks>
+    /// This is a transport-level instruction to the MCP tool loader. Images are always delivered
+    /// via <c>ImageContentBlock</c> entries and must not appear in the serialized JSON envelope,
+    /// so this property is excluded from serialization.
+    /// </remarks>
+    [JsonIgnore]
     public IReadOnlyList<ResponseImage>? Images { get; set; }
 
     /// <summary>
--- a/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
+++ b/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
@@ -32,6 +32,21 @@ public class CommandResponse
     [JsonPropertyName("images")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public IReadOnlyList<ResponseImage>? Images { get; set; }
+
+    /// <summary>
+    /// When <see langword="true"/> and the response is successful, the MCP loader will
+    /// suppress the JSON <c>TextContentBlock</c> and emit only the <see cref="Images"/>
+    /// content blocks. Use this when the rendered image is intended to fully replace
+    /// the raw data (e.g. a chart that summarizes a large query result), so vision-capable
+    /// clients consume the visualization rather than re-parsing the underlying data.
+    /// Errors always include the JSON envelope regardless of this flag.
+    /// </summary>
+    /// <remarks>
+    /// This is a transport-level instruction to the MCP tool loader and is intentionally
+    /// excluded from the serialized response payload.
+    /// </remarks>
+    [JsonIgnore]
+    public bool OmitTextContent { get; set; }
 }
 
 /// <summary>
--- a/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
+++ b/core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
@@ -22,7 +22,28 @@ public class CommandResponse
 
     [JsonPropertyName("duration")]
     public long Duration { get; set; }
+
+    /// <summary>
+    /// Optional images to include as MCP image content blocks alongside the JSON response.
+    /// When populated, each image is emitted as an <c>ImageContentBlock</c> in the MCP
+    /// <c>CallToolResult</c>, enabling vision-capable LLM clients to consume charts or
+    /// visualizations directly without parsing raw data.
+    /// </summary>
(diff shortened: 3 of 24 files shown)


servers/Azure.Mcp.Server/changelog-entries/1777538926110.yaml
alzimmermsft, line 1: Mind adding a second CHANGELOG entry for adding MCP Image Content support

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
```diff
@@ -225,13 +225,28 @@ public override async ValueTask<CallToolResult> CallToolHandler(RequestContext<C
             var jsonResponse = JsonSerializer.Serialize(commandResponse, ModelsJsonContext.Default.CommandResponse);
             var isError = commandResponse.Status < HttpStatusCode.OK || commandResponse.Status >= HttpStatusCode.Ambiguous;
 
+            var content = new List<ContentBlock>();
+
+            // When images are present and the response is successful, suppress the JSON text
+            // block so vision-capable clients receive only the rendered image(s) and skip the
+            // raw data the image was meant to replace. Errors always include the JSON text.
```
alzimmermsft: Longer term it'd be nice to have this determined at compilation time rather than runtime. Can foresee this causing a hard to troubleshoot issue where a tool tries to return an image and text together.

tools/Azure.Mcp.Tools.Kusto/src/Commands/QueryCommand.cs
```diff
@@ -14,28 +15,32 @@ namespace Azure.Mcp.Tools.Kusto.Commands;
     Id = "d1e22074-53ce-4eef-8596-0ea134a9e317",
     Name = "query",
     Title = "Query Kusto Database",
-    Description = "Executes a query against an Azure Data Explorer/Kusto/KQL cluster to search for specific terms, retrieve records, or perform management operations. Required: --cluster-uri (or --cluster and --subscription), --database, and --query.",
+    Description = "Executes a query against an Azure Data Explorer/Kusto/KQL cluster to search for specific terms, retrieve records, or perform management operations. Required: --cluster-uri (or --cluster and --subscription), --database, and --query. Optionally specify --chart-type to receive a rendered chart image alongside the raw JSON results.",
```
alzimmermsft: > Optionally specify --chart-type to receive a rendered chart image alongside the raw JSON results.

This contradicts the changes you made in `CommandFactoryToolLoader` where if there are images in the response the raw JSON results are excluded from the response.

tools/Azure.Mcp.Tools.Kusto/src/Commands/QueryCommand.cs
```diff
@@ -14,28 +15,32 @@ namespace Azure.Mcp.Tools.Kusto.Commands;
...
         command.Options.Add(KustoOptionDefinitions.Query);
+        command.Options.Add(KustoOptionDefinitions.ChartType);
     }
 
     protected override QueryOptions BindOptions(ParseResult parseResult)
     {
         var options = base.BindOptions(parseResult);
         options.Query = parseResult.GetValueOrDefault<string>(KustoOptionDefinitions.Query.Name);
+        var chartTypeStr = parseResult.GetValueOrDefault<string>(KustoOptionDefinitions.ChartType.Name);
+        options.ChartType = Enum.TryParse<ChartType>(chartTypeStr, ignoreCase: true, out var ct) ? ct : (ChartType?)null;
```
alzimmermsft: Should this silently fail if the enum type is invalid or should the handling here ignore if the option isn't passed and fail if it is passed with an invalid chart type.

If the latter, perform the validation in `RegisterOptions` using `command.Validators`.

tools/Azure.Mcp.Tools.Kusto/src/Rendering/ScottPlotChartRenderer.cs
```diff
@@ -0,0 +1,280 @@
...
+    private const int ChartHeight = 480;
+    private const string ImageMimeType = "image/png";
+
+    private static readonly HashSet<string> s_numericKqlTypes = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "int", "long", "real", "double", "decimal"
+    };
+
+    /// <inheritdoc />
+    public ResponseImage? TryRender(IReadOnlyList<JsonElement> results, ChartType chartType, string? title = null)
```
alzimmermsft: Wouldn't it be better to fail the tool call if rendering fails when requested?

Directory.Packages.props
```diff
@@ -91,6 +91,7 @@
     <PackageVersion Include="Npgsql" Version="10.0.1" />
     <PackageVersion Include="YamlDotNet" Version="16.3.0" />
     <PackageVersion Include="Yarp.ReverseProxy" Version="2.3.0" />
+    <PackageVersion Include="ScottPlot" Version="5.1.58" />
```
alzimmermsft, line 94: Is this dependency AoT compatible?

kboom: Copilot says yes :-). But let me verify

servers/Azure.Mcp.Server/changelog-entries/1777538926110.yaml
kboom: Done in 078dddab. Added a separate changelog entry covering the core MCP image content support (the ResponseImage/Images/OmitTextContent plumbing in CommandResponse and CommandFactoryToolLoader), distinct from the Kusto --chart-type feature.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
kboom: Good call — the heuristic was footgun-y. Replaced it in 078dddab with an explicit, opt-in `CommandResponse.OmitTextContent` flag. The loader no longer infers behavior from the presence of images: a tool that wants both image and text simply leaves the flag unset; a tool that wants image-only sets it explicitly. Errors still always include the JSON envelope.

```csharp
if (!commandResponse.OmitTextContent || isError)
{
    content.Add(new TextContentBlock { Text = jsonResponse });
}
```

Not fully compile-time, but it eliminates the ambiguity — there's now exactly one way to ask for image-only output and it's explicit at the call site.

tools/Azure.Mcp.Tools.Kusto/src/Commands/QueryCommand.cs
kboom: Fixed in 078dddab. Updated both `QueryCommand` and `SampleCommand` descriptions to:

> Optionally specify --chart-type to receive a rendered chart image instead of the raw JSON results (the JSON is omitted when an image is returned).

Also tightened the wording on `KustoOptionDefinitions.ChartType.Description` for consistency.

tools/Azure.Mcp.Tools.Kusto/src/Commands/QueryCommand.cs
kboom: Went with the latter — fail fast if the option is passed with an invalid value, ignore if it's not passed. Added in 078dddab via a command-level validator registered from `RegisterOptions`:

```csharp
protected override void RegisterOptions(Command command)
{
    base.RegisterOptions(command);
    command.Options.Add(KustoOptionDefinitions.Query);
    command.Options.Add(KustoOptionDefinitions.ChartType);
    KustoOptionDefinitions.AddChartTypeValidator(command);
}
```

Invalid values now produce: `Invalid value 'Bogus' for --chart-type. Allowed values: TimeSeries, Bar, Scatter, Pie.` before the command body ever runs. Covered by a new unit test (`ExecuteAsync_FailsWithValidationError_WhenChartTypeValueIsInvalid`).

tools/Azure.Mcp.Tools.Kusto/src/Rendering/ScottPlotChartRenderer.cs
kboom: Agreed — silently falling back to JSON when the user explicitly asked for a chart is misleading. Fixed in 078dddab:

- Renamed `IKustoChartRenderer.TryRender` → `Render`.
- `Render` now throws a new `ChartRenderingException` with a user-facing message whenever rendering fails (data shape doesn't match, no rows, no parseable datetime values, ScottPlot exception, etc.).
- The exception flows through `HandleException(context, ex)` so the tool call returns an error with the exact reason, e.g.:
  > `ChartType.TimeSeries requires at least one 'datetime' column. Found columns: Category:string, Count:long.`

Added unit test `ExecuteAsync_FailsWithBadRequest_WhenRendererThrowsChartRenderingException` to lock the new behavior in.

Directory.Packages.props
kboom: Verified, with caveats:

- The Kusto toolset has `<IsAotCompatible>true</IsAotCompatible>` and the AOT analyzer reports **0 warnings** after adding ScottPlot 5.1.58 — so the static analysis is clean.
- ScottPlot 5.x renders via SkiaSharp 3.x, which has Native AOT support (it's used under AOT in `Microsoft.Maui.Graphics`-style scenarios in production).
- I haven't run a full `Build-Local.ps1 -IncludeNative` pass to actually publish a NativeAOT binary in this worktree — it timed out locally. Happy to re-run if a CI lane validates this.

If the team prefers to be conservative until a clean NativeAOT publish is verified, an alternative is to gate the Kusto rendering bits behind `#if !BUILD_NATIVE` (matching the pattern used by other not-yet-AOT-validated toolsets) and remove `Azure.Mcp.Tools.Kusto` from the AOT build via the `ProjectReference-Remove` mechanism in `Azure.Mcp.Server.csproj`. Let me know your preference and I'll wire it up.

tools/Azure.Mcp.Tools.Kusto/src/Rendering/ScottPlotChartRenderer.cs
```diff
@@ -0,0 +1,308 @@
...
+        var plot = new ScottPlot.Plot();
+        plot.Title(title ?? columnNames[valueIdx]);
+        var pie = plot.Add.Pie(slices);
+        pie.ExplodeFraction = 0.05;
+        pie.SliceLabelDistance = 1.35;
+        plot.ShowLegend();
+        plot.Axes.Frameless();
+        plot.HideGrid();
+
+        return ToPngImage(plot, $"Pie chart with {slices.Count} slices");
```
jongio, line 279: This double.TryParse uses the current thread's culture. On a server with a non-US locale (e.g. `de-DE`), `"1.5"` won't parse correctly. Use `double.TryParse(element.GetString(), System.Globalization.NumberStyles.Any, CultureInfo.InvariantCulture, out var d)` to match the invariant-culture parsing already used in `TryParseDateTime` above.

tools/Azure.Mcp.Tools.Kusto/src/Rendering/ScottPlotChartRenderer.cs
```diff
@@ -0,0 +1,308 @@
...
+            throw new ChartRenderingException(
+                "Pie chart could not be rendered: no rows contained a positive numeric value to chart.");
+        }
+
+        var plot = new ScottPlot.Plot();
+        plot.Title(title ?? columnNames[valueIdx]);
+        var pie = plot.Add.Pie(slices);
+        pie.ExplodeFraction = 0.05;
+        pie.SliceLabelDistance = 1.35;
+        plot.ShowLegend();
```
jongio, line 275: `ToDouble` silently returns 0.0 for unparseable string values. If a cell in a numeric column contains `"N/A"` or another non-numeric string, it'll be plotted as zero - producing a misleading chart without any indication to the caller. Consider either logging a warning, skipping the row, or failing when a threshold of unparseable cells is exceeded.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs
```diff
@@ -225,13 +225,27 @@ public override async ValueTask<CallToolResult> CallToolHandler(RequestContext<C
             var jsonResponse = JsonSerializer.Serialize(commandResponse, ModelsJsonContext.Default.CommandResponse);
             var isError = commandResponse.Status < HttpStatusCode.OK || commandResponse.Status >= HttpStatusCode.Ambiguous;
 
+            var content = new List<ContentBlock>();
+
+            // The text JSON envelope is suppressed only when the command explicitly opted in via
+            // CommandResponse.OmitTextContent (e.g. an image is intended to fully replace raw data).
+            // Errors always include the JSON envelope so the caller can read status/message.
```
jongio, line 232: `ImageContentBlock.FromBytes(image.Data, image.MimeType)` drops `image.AltText`. If the `ImageContentBlock` type supports alt text (or a description field), pass it through. Otherwise, consider encoding it in the MCP response metadata so vision-impaired clients or non-vision LLMs get the fallback text.

core/Microsoft.Mcp.Core/src/Models/Command/CommandResponse.cs
```diff
@@ -22,8 +22,48 @@ public class CommandResponse
...
+    /// </summary>
+    /// <remarks>
+    /// This is a transport-level instruction to the MCP tool loader and is intentionally
+    /// excluded from the serialized response payload.
+    /// </remarks>
+    [JsonIgnore]
+    public bool OmitTextContent { get; set; }
 }
 
+/// <summary>
```
jongio, line 56: `ResponseImage` has `[JsonPropertyName]` attributes but `CommandResponse.Images` is `[JsonIgnore]`, so this type is never serialized through the normal response path. The JSON annotations are dead decoration. Either remove them (this type flows through `ImageContentBlock.FromBytes`, not JSON serialization) or add a comment explaining if they're intentionally kept for a future serialization path.

jongio (commented):
Solid feature add - opt-in chart rendering for Kusto queries makes sense for vision-capable LLM clients. The architecture (validator + renderer + core image response) is clean and the opt-in design preserves backwards compat. A few issues to address beyond what alzimmermsft and Copilot already raised:

**AltText is dropped during emission.** Both `CommandFactoryToolLoader` and `NamespaceToolLoader` call `ImageContentBlock.FromBytes(image.Data, image.MimeType)` without passing `image.AltText`. This means the accessibility text ("Bar chart with 3 categories") never reaches the MCP client. Copilot flagged this too - it needs fixing.

**`ToDouble` silently returns 0.0 for unparseable values.** In `ScottPlotChartRenderer`, if a cell that's supposed to be numeric contains garbage (e.g. `"N/A"` or `null`), it silently becomes 0.0 and gets plotted. This produces misleading charts. Consider either skipping the row or throwing a `ChartRenderingException` when a numeric column contains unparseable data.

**`double.TryParse` in `ToDouble` is culture-sensitive.** When the element is a `JsonValueKind.String`, `double.TryParse(element.GetString(), out var d)` uses the current thread culture. A server running in `de-DE` locale would parse `"1,5"` as 1.5 but fail on `"1.5"`. Use `CultureInfo.InvariantCulture` explicitly.

**`ResponseImage` has `[JsonPropertyName]` attributes but is never serialized through the JSON response path.** Since `CommandResponse.Images` is `[JsonIgnore]`, the `JsonPropertyName` decorations on the `ResponseImage` record are misleading dead code. Either remove them (cleaner - the type isn't serialized) or document why they exist (e.g. future use, testing). The `ModelsJsonContext` registration of `ResponseImage` and `IReadOnlyList<ResponseImage>` is similarly unused at runtime.

**CI is failing on Build Analyze.** This needs to be green before merge.

jongio (commented):
Incremental review of 2 new commits (90b3190, cb906dc) since my last review.

The new `else if` branch in both tool loaders correctly emits `commandResponse.Message` as a lightweight text hint alongside chart images when the JSON envelope is suppressed. The guidance text in QueryCommand gives vision-capable LLMs a clear directive to describe visual patterns rather than extract exact values. Both changes are solid.

My 4 previous comments are still open:
- `ToDouble` silently maps unparseable strings to 0.0 (misleading charts)
- `double.TryParse` lacks `CultureInfo.InvariantCulture` (locale-dependent parsing)
- `ImageContentBlock.FromBytes` drops `AltText`
- `ResponseImage` has dead `[JsonPropertyName]` attributes

No new issues in this delta.

Issue #2496, opened 2026-04-24
[EPIC] Command Contract and Documentation Drift Elimination


vcolin7:
## Problem statement

Azure MCP command contracts change continuously as tools are added, migrated, renamed, or corrected. The command reference, end-to-end prompts, README examples, option guidance, and release documentation are maintained separately, allowing user-facing documentation to drift from the executable command surface. Current gaps span command names and structure, option presence and requiredness, valid values and constraints, behavioral semantics, and scenario or prompt coverage across many toolsets.

## Vision

The documented Azure MCP command surface is validated against the product contract so command names, parameters, constraints, behavior, examples, and prompts remain accurate as implementation changes land.

## Current work

Active work is tracked through GitHub's native sub-issue relationships. The child portfolio is the source of truth for remediation work and evolves as documentation drift is detected and resolved.

The merged [documentation gap detector](https://github.com/microsoft/mcp/pull/2550) provides an existing foundation by identifying documentation gaps caused by tool changes and filing focused follow-up issues.

## Goals (in scope)

- Resolve documented drift between executable command contracts and user-facing command reference material
- Keep command names, command structure, parameters, requiredness, valid values, constraints, and behavior accurate
- Keep README examples and end-to-end prompts aligned with supported commands and scenarios
- Validate command references against executable metadata where reliable automation is available
- Produce actionable drift reports that identify the affected command contract and documentation surface

## Non-goals (out of scope)

- Replacing conceptual, tutorial, or scenario documentation with generated reference material
- Treating editorial preferences as command-contract failures
- Owning general pull request or release guardrails outside command-contract documentation
- Blocking unrelated changes on pre-existing documentation gaps

## Success criteria

- [ ] Direct child issues are resolved or explicitly dispositioned through the native sub-issue hierarchy
- [ ] Supported command names and structures are represented accurately in command reference documentation
- [ ] Documented options match executable presence, requiredness, valid values, and constraints
- [ ] User-visible command behavior changes are reflected in command behavior documentation
- [ ] README examples and end-to-end prompts cover the supported commands and scenarios identified by the child portfolio
- [ ] Automated drift detection or validation produces focused, actionable follow-up for supported contract changes

## Dependencies and boundaries

- Stable command metadata and tool schema generation are required for reliable validation
- The documentation gap detector identifies gaps caused by merged changes; resolving and closing the resulting child issues remains owned by the relevant toolset and documentation owners
[...]

Issue #2483, opened 2026-04-23
Implement Agentic workflow to ensure documentation is up to date.


anannya03:
Ensuring documentation gets updated when new commands are added or existing ones are modified in Azure MCP server.
[Documentation Update Agentic Workflow](https://github.github.com/gh-aw/blog/2026-01-13-meet-the-workflows-documentation/)

anannya03:
Part of this meta tracking issue: #2494

PR #2550, opened 2026-04-30
Agent workflow to update Documentation

anannya03:
## What does this PR do?

Adds a GitHub Agentic Workflow (doc-gap-detector.md) that automatically detects documentation gaps when code is
pushed to main. It analyzes diffs for tool changes and files a focused GitHub issue describing what documentation
needs updating — assignable to @copilot for automated implementation.

### What it does

 - Triggers on every push to main (or manual dispatch)
 - Analyzes the push diff to identify MCP tool additions, renames, removals, or parameter changes
 - Cross-references against three documentation files:
    -  servers/Azure.Mcp.Server/README.md
    - servers/Azure.Mcp.Server/docs/azmcp-commands.md
    - servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
 - Files a single [Doc Gap] issue with structured context, gaps found, and an embedded Implementation Guide for the 
coding agent
 - No-ops when no documentation updates are needed

### Key design decisions

 - Max 1 issue per run to prevent spam
 - Scoped strictly to gaps caused by the triggering push (not pre-existing debt)
 - Checks for duplicate issues before filing
 - Implementation Guide in the issue body includes exact file structure expectations and validation steps (dotnet 
build, dotnet test, spell check)

## GitHub issue number?
https://github.com/microsoft/mcp/issues/2483
Should partly fix this as well- https://github.com/microsoft/mcp/issues/2496


.github/workflows/doc-gap-detector.md
```diff
@@ -0,0 +1,115 @@
...
+  Documentation gap detector for the Azure MCP repository
+  Triggered on every push to main, analyzes diff to identify documentation gaps
+  and files GitHub issues for copilot coding agent to implement.
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions: read-all
```
jongio: `permissions: read-all` grants read access to every permission scope. The issue-triage workflow scopes this correctly - consider doing the same here:

```yaml
permissions:
  contents: read
  issues: write
```

The workflow only needs to read repo contents (for diffs) and write issues (for filing doc gap issues). `read-all` includes scopes like `security-events` and `organization-administration` that aren't needed here.

jongio (commented):
Early feedback on this draft. A few things to tighten up before it's ready.

The doc-gap-detector uses `permissions: read-all` which is broader than needed - the issue-triage workflow scopes permissions correctly and is a good pattern to follow. There's also a safe-outputs/prompt mismatch on issue creation limits (details inline).

Minor note: issue-triage doesn't set `timeout-minutes` while doc-gap-detector does (15 min). Worth adding for consistency and to cap runaway agent sessions.

.github/workflows/doc-gap-detector.md
```diff
@@ -0,0 +1,115 @@
...
+    toolsets: [default]
+
+timeout-minutes: 15
+
+safe-outputs:
+  report-failure-as-issue: false
+  create-issue:
+    title-prefix: "[Docs]"
+    labels: ['documentation']
+    max: 1
```
jongio, line 35: `max: 1` here conflicts with rule #4 in the prompt below ("Max 3 issues per run"). The safe-output limit will silently cap the agent at 1 issue regardless of what the prompt says. Either bump this to `max: 3` to match the prompt, or update the prompt rule to say max 1.

.github/workflows/doc-gap-detector.md
```diff
@@ -0,0 +1,115 @@
...
+    branches: [main]
+  workflow_dispatch:
+
+permissions: read-all
+
+network:
+  allowed:
+    - defaults
+    - github
+    - dotnet
```
jongio: Is `dotnet` network access needed here? This workflow analyzes diffs and creates issues - it doesn't build or run .NET code. If the agent doesn't need to reach NuGet or dotnet endpoints, dropping this tightens the network surface.

.github/workflows/doc-gap-detector.md
```diff
@@ -0,0 +1,208 @@
...
+  and files GitHub issues for copilot coding agent to implement.
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: read
```
jongio, line 15: `issues: read` lets the agent check for duplicates (rule #8), but `create-issue` needs write access to file issues. This should be `issues: write` (which implicitly includes read).

jongio (commented):
Addresses my previous feedback: permissions are scoped, `dotnet` network access is removed, and `max: 1` is consistent with the prompt. The PR scope is also tightened to just the doc-gap-detector (issue-triage removed), and the restructured prompt is much clearer.

One new item on the permissions scope below.

Also noting the Copilot bot raised valid points about the `sdk/` path in the issue template (should match the actual `tools/` structure), the `copilot` label mismatch, and the "capbilities" typo.

.github/workflows/doc-gap-detector.md
anannya03: `issues:write` is not allowed for security reasons. 
To create issues safely, the recommended method is to use `safe-outputs.create-issue`. 

`gh aw compile` fails on trying to set write permission for issues.

jsquire (approved):
This looks good to me.   The flow is well specified; the resulting issue is detailed and has the recommended implementation steps.

KarishmaGhiya (approved):

PR #2552, opened 2026-04-30
Fix/Add  Resource Group Scoping

g2vinay:
# Fix/Add: `--resource-group` parameter ignored in 6 list/get commands

Six commands were not supporting a `--resource-group`  parameter.  As a result, passing `--resource-group my-rg` had no effect and always returned subscription-wide results.

**Affected commands:**
- `azmcp grafana list`
- `azmcp kusto cluster list`
- `azmcp appconfig account list`
- `azmcp storage account get`
- `azmcp monitor workspace list`
- `azmcp search service list`

## What was fixed

The `resourceGroup` value is now correctly passed from each command's `BindOptions` through to the service layer, so results are filtered to the specified resource group when provided.

## Changes

- **Bug fix** — 6 commands now forward `--resource-group` to their service methods
- **Tests** — added `ExecuteAsync_WithResourceGroup_ForwardsResourceGroupToService` unit test to each affected command's test file
- **Docs** — updated `azmcp-commands.md` to show `--resource-group` as an optional parameter for all 6 commands
- **Changelog** — added entry under "Bugs Fixed"
- **Code quality** — cleaned up an awkward variable name in `SearchService.cs` (`subscriptionResourceSub` → `subscriptionResource`, inner scope renamed to `subForRg`)

### DIFF
--- a/tools/Azure.Mcp.Tools.AppConfig/src/Commands/Account/AccountListCommand.cs
+++ b/tools/Azure.Mcp.Tools.AppConfig/src/Commands/Account/AccountListCommand.cs
@@ -7,7 +7,9 @@ using Azure.Mcp.Tools.AppConfig.Options.Account;
 using Azure.Mcp.Tools.AppConfig.Services;
 using Microsoft.Extensions.Logging;
 using Microsoft.Mcp.Core.Commands;
+using Microsoft.Mcp.Core.Extensions;
 using Microsoft.Mcp.Core.Models.Command;
+using Microsoft.Mcp.Core.Models.Option;
 
 namespace Azure.Mcp.Tools.AppConfig.Commands.Account;
 
@@ -30,6 +32,19 @@ public sealed class AccountListCommand(ILogger<AccountListCommand> logger, IAppC
     private readonly ILogger<AccountListCommand> _logger = logger;
     private readonly IAppConfigService _appConfigService = appConfigService;
 
+    protected override void RegisterOptions(Command command)
+    {
+        base.RegisterOptions(command);
+        command.Options.Add(OptionDefinitions.Common.ResourceGroup.AsOptional());
+    }
+
+    protected override AccountListOptions BindOptions(ParseResult parseResult)
+    {
+        var options = base.BindOptions(parseResult);
+        options.ResourceGroup = parseResult.GetValueOrDefault<string>(OptionDefinitions.Common.ResourceGroup.Name);
+        return options;
+    }
+
     public override async Task<CommandResponse> ExecuteAsync(CommandContext context, ParseResult parseResult, CancellationToken cancellationToken)
     {
         if (!Validate(parseResult.CommandResult, context.Response).IsValid)
@@ -43,6 +58,7 @@ public sealed class AccountListCommand(ILogger<AccountListCommand> logger, IAppC
         {
             var accounts = await _appConfigService.GetAppConfigAccounts(
                 options.Subscription!,
+                options.ResourceGroup,
                 options.Tenant,
                 options.RetryPolicy,
                 cancellationToken);
--- a/tools/Azure.Mcp.Tools.Monitor/src/Services/IMonitorService.cs
+++ b/tools/Azure.Mcp.Tools.Monitor/src/Services/IMonitorService.cs
@@ -40,6 +40,7 @@ public interface IMonitorService
 
     Task<List<WorkspaceInfo>> ListWorkspaces(
         string subscription,
+        string? resourceGroup = null,
         string? tenant = null,
         RetryPolicyOptions? retryPolicy = null,
         CancellationToken cancellationToken = default);
--- a/tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorService.cs
+++ b/tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorService.cs
@@ -198,11 +198,29 @@ public class MonitorService(
 
     public async Task<List<WorkspaceInfo>> ListWorkspaces(
         string subscription,
-        string? tenant,
-        RetryPolicyOptions? retryPolicy,
-        CancellationToken cancellationToken)
+        string? resourceGroup = null,
+        string? tenant = null,
+        RetryPolicyOptions? retryPolicy = null,
+        CancellationToken cancellationToken = default)
     {
         ValidateRequiredParameters((nameof(subscription), subscription));
+
+        if (!string.IsNullOrEmpty(resourceGroup))
+        {
+            var rgResource = await resourceGroupService.GetResourceGroupResource(subscription, resourceGroup, tenant, retryPolicy, cancellationToken)
+                ?? throw new Exception($"Resource group '{resourceGroup}' not found in subscription '{subscription}'.");
+
+            return await rgResource
+                .GetOperationalInsightsWorkspaces()
+                .GetAllAsync(cancellationToken)
+                .Select(workspace => new WorkspaceInfo
+                {
+                    Name = workspace.Data.Name,
+                    CustomerId = workspace.Data.CustomerId?.ToString() ?? string.Empty,
+                })
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
+        }
 
         var subscriptionResource = await subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken);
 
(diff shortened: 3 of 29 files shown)


tools/Azure.Mcp.Tools.AppConfig/src/Commands/Account/AccountListCommand.cs
```diff
@@ -30,6 +32,19 @@ public sealed class AccountListCommand(ILogger<AccountListCommand> logger, IAppC
     private readonly ILogger<AccountListCommand> _logger = logger;
     private readonly IAppConfigService _appConfigService = appConfigService;
 
+    protected override void RegisterOptions(Command command)
+    {
+        base.RegisterOptions(command);
+        command.Options.Add(OptionDefinitions.Common.ResourceGroup.AsOptional());
```
alzimmermsft, line 38: nit: `ResourceGroup` is defined as an optional parameter


```suggestion
        command.Options.Add(OptionDefinitions.Common.ResourceGroup);
```

alzimmermsft (commented):
Should add at least unit tests with a specific resource group being passed in the listing operation. Also, should add live tests listing with the deployment resource group being passed but given this spans many tool areas that'll be a much more intensive add.

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
```diff
@@ -41,12 +41,28 @@ public sealed partial class SearchService(
...
+            var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenantId, retryPolicy, cancellationToken);
+            var rgResource = (await subscriptionResource.GetResourceGroupAsync(resourceGroup, cancellationToken)).Value;
+            var rgServices = new List<string>();
+            await foreach (var service in rgResource.GetSearchServices().GetAllAsync(cancellationToken: cancellationToken))
+            {
+                if (service?.Data?.Name != null)
+                {
+                    rgServices.Add(service.Data.Name);
+                }
+            }
```
jongio, line 72: The resource-group path returns directly without touching the cache. The subscription-wide path below caches results for 1 hour, so RG-scoped queries always hit the API and their results don't warm the subscription-wide cache.

If that's intentional (RG queries are cheap enough), a short comment explaining why would help. If not, consider adding cache support with a key that includes the resource group.

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
```diff
@@ -57,9 +73,9 @@ public async Task<List<string>> ListServices(
             return cachedServices;
         }
 
-        var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenantId, retryPolicy, cancellationToken);
+        var subscriptionResourceSub = await _subscriptionService.GetSubscription(subscription, tenantId, retryPolicy, cancellationToken);
```
jongio:
```suggestion
        var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenantId, retryPolicy, cancellationToken);
```

This rename to `subscriptionResourceSub` isn't needed - the resource-group block above returns early, so there's no name conflict. Keeping the original name avoids unnecessary diff noise.

jongio (commented):
Adds optional `--resource-group` to 6 list/get commands. The pattern for AppConfig, Grafana, Kusto, and Storage is clean - they pass through to `ExecuteResourceQueryAsync` which already handles RG scoping.

A couple of things to look at in the Monitor and Search services where the implementation is more involved. Details in inline comments.

+1 on @alzimmermsft's comment about tests - the existing test updates just accommodate the new parameter with `Arg.Any<string?>()`, but there aren't any tests that actually pass a specific resource group and verify the behavior. At minimum, each command should have a test that verifies the resource group value is forwarded to the service.

On the `.AsOptional()` nit: the codebase convention is to use `.AsOptional()` when registering ResourceGroup (25+ commands do this). Even though the definition already sets `Required = false`, keeping `.AsOptional()` is consistent with how every other command does it. Either approach works, but I'd lean toward consistency.

Minor: no PR description. For a 26-file change across 7 tool areas, context on what was broken and how to test would help reviewers.

tools/Azure.Mcp.Tools.AppConfig/src/Commands/Account/AccountListCommand.cs
g2vinay: per Jon's comment below, consistent pattern is to use .AsOptional(), so sticking with that for now.

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
g2vinay: It is needed due to C# scoping rule CS0136. The compiler rejects it with "A local named 'subscriptionResource' cannot be declared in this scope because that name is used in an enclosing local scope".

jongio (commented):
Addresses my previous feedback. The `subForRg` rename is a clean workaround for CS0136, keeping `subscriptionResource` as the canonical name in the main path. All 6 services now have RG forwarding tests with `Arg.Is()` assertions. Changelog added.

Cache bypass in the RG-scoped path (`SearchService.cs:51-64`) is still open from my earlier comment. If that's intentional for targeted queries, it's fine.

Still no PR description for a 27-file change.

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
jongio: Makes sense. `subForRg` works - keeps `subscriptionResource` as the readable name for the common path.

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
g2vinay: added the cache logic for RG.

alzimmermsft (approved):

tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorService.cs
```diff
@@ -484,7 +502,7 @@ private async Task<ActivityLogListResponse> MakeActivityLogRequestAsync(string u
     {
         // If we're given an ID and need an ID, or given a name and need a name, return as is
         bool isId = IsWorkspaceId(workspace);
-        var workspaces = await ListWorkspaces(subscription, tenant, retryPolicy, cancellationToken);
+        var workspaces = await ListWorkspaces(subscription, resourceGroup: null, tenant, retryPolicy, cancellationToken);
```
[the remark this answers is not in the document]
g2vinay: Since C# 7.2+, non-trailing named arguments are allowed, a named argument can appear anywhere in the argument list as long as it's used in the correct position or the remaining positional arguments still match up in order. The compiler accepts this because the order is unambiguous.

tools/Azure.Mcp.Tools.Monitor/src/Services/IMonitorService.cs
```diff
@@ -40,6 +40,7 @@ Task<List<string>> ListTables(
 
     Task<List<WorkspaceInfo>> ListWorkspaces(
         string subscription,
+        string? resourceGroup = null,
         string? tenant = null,
```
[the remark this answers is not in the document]
g2vinay: good catch, fixed.

jongio (commented):
All prior findings addressed:

- Cache bypass in the RG-scoped path now has proper caching with a distinct key that includes `resourceGroup`, matching the existing subscription-wide pattern. Cache tests verify the keys don't collide.
- `LogAnalyticsHelper` positional arg fix is necessary since `resourceGroup` was inserted before `tenantId` in the method signature. Without the `tenant:` named arg, `_tenantId` would silently land in the wrong parameter.
- PR description added and thorough.

No new issues. This is ready for approval once it's out of draft.

jongio (approved):
All prior findings addressed across the last 4 commits:

- Cache bypass for RG-scoped search results: fixed with distinct cache key including resourceGroup
- Variable rename in SearchService: resolved with `subForRg` (CS0136 workaround)
- RG forwarding tests: added with `Arg.Is()` assertions for all 6 commands
- LogAnalyticsHelper positional arg fix: `tenant:` named arg prevents misrouting after parameter reorder
- PR description and changelog entry: both present

Clean incremental review - no new issues found. Approving.

PR #2618, opened 2026-05-10
Add user_config to Azure MCP Bundle

pl4nty:
## What does this PR do?

Add user_config to Azure MCP Bundle, so users can configure settings in clients like Claude Desktop. Some settings were excluded because bundles only support stdio.

image

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/servers/Azure.Mcp.Server/mcpb/manifest.json
+++ b/servers/Azure.Mcp.Server/mcpb/manifest.json
@@ -1,6 +1,6 @@
 {
-  "$schema": "https://raw.githubusercontent.com/modelcontextprotocol/mcpb/refs/heads/main/schemas/mcpb-manifest-v0.4.schema.json",
-  "manifest_version": "0.4",
+  "$schema": "https://raw.githubusercontent.com/modelcontextprotocol/mcpb/refs/heads/main/schemas/mcpb-manifest-v0.3.schema.json",
+  "manifest_version": "0.3",
   "name": "Azure.Mcp.Server",
   "display_name": "Azure MCP Server",
   "version": "2.0.0-beta.17",
--- a/servers/Azure.Mcp.Server/mcpb/manifest.json
+++ b/servers/Azure.Mcp.Server/mcpb/manifest.json
@@ -1,5 +1,6 @@
 {
-  "manifest_version": "0.3",
+  "$schema": "https://raw.githubusercontent.com/modelcontextprotocol/mcpb/refs/heads/main/schemas/mcpb-manifest-v0.4.schema.json",
+  "manifest_version": "0.4",
   "name": "Azure.Mcp.Server",
   "display_name": "Azure MCP Server",
   "version": "2.0.0-beta.17",
@@ -28,8 +29,31 @@
         }
       },
       "command": "${__dirname}/server/azmcp",
-      "args": [ "server", "start" ],
-      "env": {}
+      "args": [
+        "server",
+        "start",
+        "--mode",
+        "${user_config.mode}",
+        "--debug",
+        "${user_config.debug}",
+        "--read-only",
+        "${user_config.read_only}",
+        "--dangerously-disable-elicitation",
+        "${user_config.dangerously_disable_elicitation}"
+      ],
+      "env": {
+        "AZURE_TENANT_ID": "${user_config.azure_tenant_id}",
+        "AZURE_CLIENT_ID": "${user_config.azure_client_id}",
+        "AZURE_CLIENT_SECRET": "${user_config.azure_client_secret}",
+        "AZURE_SUBSCRIPTION_ID": "${user_config.azure_subscription_id}",
+        "AZURE_CLOUD": "${user_config.azure_cloud}",
+        "AZURE_TOKEN_CREDENTIALS": "${user_config.azure_token_credentials}",
+        "AZURE_MCP_COLLECT_TELEMETRY": "${user_config.azure_mcp_collect_telemetry}",
+        "AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT": "${user_config.azure_mcp_collect_telemetry_microsoft}",
+        "AZURE_MCP_ONLY_USE_BROKER_CREDENTIAL": "${user_config.azure_mcp_only_use_broker_credential}",
+        "AZURE_MCP_ENABLE_OTLP_EXPORTER": "${user_config.azure_mcp_enable_otlp_exporter}",
+        "APPLICATIONINSIGHTS_CONNECTION_STRING": "${user_config.applicationinsights_connection_string}"
+      }
     }
   },
   "tools": [],
@@ -94,6 +118,98 @@
     "quick reference",
     "cli"
   ],
+  "user_config": {
+    "azure_tenant_id": {
+      "type": "string",
+      "title": "Entra Tenant ID",
+      "description": "Entra tenant ID. Use this to authenticate against a specific tenant. Example: 00000000-0000-0000-0000-000000000000",
+      "default": ""
+    },
+    "azure_client_id": {
+      "type": "string",
+      "title": "Entra Client ID",
+      "description": "Client ID of a service principal or user-assigned managed identity. Example: 00000000-0000-0000-0000-000000000000",
+      "default": ""
+    },
+    "azure_client_secret": {
+      "type": "string",
+      "title": "Entra Client Secret",
+      "description": "Client secret for service principal authentication. Used together with AZURE_TENANT_ID and AZURE_CLIENT_ID for non-interactive environments.",
+      "default": ""
+    },
+    "azure_subscription_id": {
+      "type": "string",
+      "title": "Azure Subscription ID",
+      "description": "Default Azure subscription ID or name to use when no subscription is specified in a command. Example: 00000000-0000-0000-0000-000000000000",
+      "default": ""
+    },
+    "azure_cloud": {
+      "type": "string",
+      "title": "Azure Cloud Environment",
+      "description": "Azure cloud environment to connect to. Supported values: AzureCloud (default public cloud), AzureChinaCloud, AzureUSGovernment.",
+      "default": ""
+    },
+    "azure_token_credentials": {
+      "type": "string",
+      "title": "Azure Token Credentials",
+      "description": "Credential chain to use for authentication. Values: 'dev' (VS/VSCode/CLI/PowerShell/AzureDeveloperCLI + browser fallback), 'prod' (Workload Identity + Managed Identity, no interactive fallback), or a specific provider name such as 'AzureCliCredential' or 'ManagedIdentityCredential'.",
+      "default": ""
+    },
+    "azure_mcp_collect_telemetry": {
+      "type": "boolean",
+      "title": "Collect Telemetry",
+      "description": "Whether to allow Azure MCP Server to collect anonymous usage telemetry. Set to false to opt out.",
+      "default": true
+    },
+    "azure_mcp_collect_telemetry_microsoft": {
+      "type": "boolean",
+      "title": "Collect Microsoft Telemetry",
+      "description": "Whether to allow Azure MCP Server to send anonymous usage telemetry to Microsoft. Enabled by default in release builds. Set to false to opt out of Microsoft-owned telemetry while keeping other telemetry settings unchanged.",
+      "default": true
+    },
+    "azure_mcp_only_use_broker_credential": {
+      "type": "boolean",
+      "title": "Only Use Broker Credential",
+      "description": "When set to true, restricts authentication to the broker credential only (e.g. Windows Web Account Manager). Useful in environments where only broker-based authentication is permitted.",
+      "default": false
+    },
+    "azure_mcp_enable_otlp_exporter": {
+      "type": "boolean",
+      "title": "Enable OTLP Exporter",
+      "description": "When set to true, enables the OpenTelemetry Protocol (OTLP) exporter for traces, metrics, and logs. Configure the exporter endpoint via standard OTEL_EXPORTER_OTLP_ENDPOINT environment variables.",
+      "default": false
+    },
+    "applicationinsights_connection_string": {
+      "type": "string",
+      "title": "Application Insights Connection String",
+      "description": "Azure Application Insights connection string for custom telemetry reporting.",
+      "default": ""
+    },
+    "debug": {
+      "type": "boolean",
+      "title": "Debug",
+      "description": "Enable debug mode with verbose logging to stderr. Default: false.",
+      "default": false
+    },
+    "dangerously_disable_elicitation": {
+      "type": "boolean",
+      "title": "Disable User Confirmation (Not Recommended)",
+      "description": "Disable user confirmation (elicitation) before allowing high risk commands to run, such as returning secrets (passwords) from KeyVault. When enabled, tools that handle secrets, credentials, or sensitive data will execute without user confirmation. This removes an important security layer designed to prevent unauthorized access to sensitive information. Only use this option in trusted, automated environments where user interaction isn't possible. Never use this option in production environments or when handling untrusted input. Default: false.",
+      "default": false
+    },
+    "mode": {
+      "type": "string",
+      "title": "Mode",
+      "description": "Server mode: namespace, consolidated, all, or single. Default: namespace.",
+      "default": "namespace"
+    },
+    "read_only": {
+      "type": "boolean",
+      "title": "Read Only",
+      "description": "Whether the MCP server should be read-only. If true, no write operations are allowed. Default: false.",
+      "default": false
+    }
+  },
   "license": "MIT",
   "privacy_policies": [
     "https://www.microsoft.com/privacy/privacystatement"
--- a/servers/Azure.Mcp.Server/mcpb/manifest.json
+++ b/servers/Azure.Mcp.Server/mcpb/manifest.json
@@ -135,7 +135,8 @@
       "type": "string",
       "title": "Entra Client Secret",
       "description": "Client secret for service principal authentication. Used together with AZURE_TENANT_ID and AZURE_CLIENT_ID for non-interactive environments.",
-      "default": ""
+      "default": "",
+      "sensitive": true
     },
     "azure_subscription_id": {
       "type": "string",
@@ -146,7 +147,7 @@
     "azure_cloud": {
       "type": "string",
       "title": "Azure Cloud Environment",
-      "description": "Azure cloud environment to connect to. Supported values: AzureCloud (default public cloud), AzureChinaCloud, AzureUSGovernment.",
+      "description": "Azure cloud environment to connect to. Examples: AzureCloud (default public cloud), AzureChinaCloud, AzureUSGovernment.",
       "default": ""
     },
     "azure_token_credentials": {
@@ -183,7 +184,8 @@
       "type": "string",
       "title": "Application Insights Connection String",
       "description": "Azure Application Insights connection string for custom telemetry reporting.",
-      "default": ""
+      "default": "",
+      "sensitive": true
     },
     "debug": {
       "type": "boolean",


servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,98 @@
     "quick reference",
     "cli"
   ],
+  "user_config": {
+    "azure_tenant_id": {
+      "type": "string",
+      "title": "Entra Tenant ID",
+      "description": "Entra tenant ID. Use this to authenticate against a specific tenant. Example: 00000000-0000-0000-0000-000000000000",
+      "default": ""
```
pl4nty, line 126: While implementing https://github.com/microsoft/fabric-rti-mcp/pull/172, I noticed Claude Desktop only substitutes string variables if a value is provided. This can cause server issues like trying to parse a guid from raw `${user_config.azure_tenant_id}`, so I've set default values

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,98 @@
...
+      "type": "string",
+      "title": "Azure Token Credentials",
+      "description": "Credential chain to use for authentication. Values: 'dev' (VS/VSCode/CLI/PowerShell/AzureDeveloperCLI + browser fallback), 'prod' (Workload Identity + Managed Identity, no interactive fallback), or a specific provider name such as 'AzureCliCredential' or 'ManagedIdentityCredential'.",
+      "default": ""
+    },
+    "azure_mcp_collect_telemetry": {
+      "type": "boolean",
+      "title": "Collect Telemetry",
+      "description": "Whether to allow Azure MCP Server to collect anonymous usage telemetry. Set to false to opt out.",
+      "default": true
```
pl4nty, line 163: Booleans in Claude Desktop are non-nullable and currently default to false

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,98 @@
...
+    "azure_subscription_id": {
+      "type": "string",
+      "title": "Azure Subscription ID",
+      "description": "Default Azure subscription ID or name to use when no subscription is specified in a command. Example: 00000000-0000-0000-0000-000000000000",
+      "default": ""
+    },
+    "azure_cloud": {
+      "type": "string",
+      "title": "Azure Cloud Environment",
+      "description": "Azure cloud environment to connect to. Supported values: AzureCloud (default public cloud), AzureChinaCloud, AzureUSGovernment.",
```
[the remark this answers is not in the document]
pl4nty: I think aliases here would be more confusing than helpful. Using "Examples" makes sense though, in case other clouds are supported (Delos, GovSG, etc)

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -28,8 +29,31 @@
...
+        "AZURE_CLIENT_ID": "${user_config.azure_client_id}",
+        "AZURE_CLIENT_SECRET": "${user_config.azure_client_secret}",
+        "AZURE_SUBSCRIPTION_ID": "${user_config.azure_subscription_id}",
+        "AZURE_CLOUD": "${user_config.azure_cloud}",
+        "AZURE_TOKEN_CREDENTIALS": "${user_config.azure_token_credentials}",
+        "AZURE_MCP_COLLECT_TELEMETRY": "${user_config.azure_mcp_collect_telemetry}",
+        "AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT": "${user_config.azure_mcp_collect_telemetry_microsoft}",
+        "AZURE_MCP_ONLY_USE_BROKER_CREDENTIAL": "${user_config.azure_mcp_only_use_broker_credential}",
+        "AZURE_MCP_ENABLE_OTLP_EXPORTER": "${user_config.azure_mcp_enable_otlp_exporter}",
+        "APPLICATIONINSIGHTS_CONNECTION_STRING": "${user_config.applicationinsights_connection_string}"
```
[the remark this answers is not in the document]
pl4nty: I can't find any examples of the suggested syntax, so I'm not sure it's supported

jongio (commented):
This adds a useful user_config surface for MCP clients. A few things to address before merge:

1. The `platform_overrides.win32` section still has `"args": []` and no `env` block. Depending on how the mcpb runtime handles overrides (merge vs replace), Windows users may not receive the new args or env vars from user_config. Should the win32 override be updated to mirror the main mcp_config, or does the runtime inherit unset fields from the parent?

2. The boolean env vars (`AZURE_MCP_COLLECT_TELEMETRY`, `AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT`) have `default: true` in user_config. You mentioned Claude Desktop defaults booleans to false. If the client ignores the manifest default and uses false, telemetry would be disabled for all bundle users by default. This contradicts the server's built-in behavior (telemetry on unless explicitly opted out). Can you confirm which default wins at runtime?

3. The `sensitive: true` additions look correct and address the credential storage concern.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -28,8 +29,31 @@
...
-      "env": {}
+      "args": [
+        "server",
+        "start",
+        "--mode",
+        "${user_config.mode}",
+        "--debug",
+        "${user_config.debug}",
+        "--read-only",
+        "${user_config.read_only}",
```
jongio, line 40: When `user_config.debug` is false (the default), this expands to `--debug false`. System.CommandLine parses it fine, but it means every launch always includes `--debug false --read-only false --dangerously-disable-elicitation false` which is noisy. Not blocking, but if mcpb supports conditional arg inclusion (only add when true), that'd be cleaner.

servers/Azure.Mcp.Server/mcpb/manifest.json
jongio, line 55: The server reads `AZURE_MCP_COLLECT_TELEMETRY` at ServiceStartCommand.cs:967-969. Empty/whitespace is treated as true (telemetry enabled). So empty string defaults for string env vars are safe and won't override existing user env in a harmful way.

But boolean-typed user_config values always resolve to a concrete true/false string. If a user previously set `AZURE_MCP_COLLECT_TELEMETRY=false` in their shell profile and the bundle sets it to true (from the manifest default), their opt-out is silently overridden. Worth documenting this behavior in the description fields, something like 'Overrides any existing environment variable when configured.'

pl4nty:
My testing was on Windows, user_config works without modifying the win32 overrides. Sorry I wasn't clear about the defaults though - Claude uses false, if a default isn't provided. That's why I used `default: true`.

servers/Azure.Mcp.Server/mcpb/manifest.json
pl4nty: I think the server isn't run from a shell, so envvar overrides shouldn't be an issue. But I can mention just in case if you'd like

servers/Azure.Mcp.Server/mcpb/manifest.json
pl4nty: Conditional inclusion would be ideal, but I couldn't find any docs nor get it working with common syntax. I may end up writing a feature request on the spec repo for it

pl4nty:
Bump, this is a significant headache for my users and I. eg `AZURE_TOKEN_CREDENTIALS` would allow us to use multiple accounts

pl4nty:
we gave up on the MCP bundle and hosted the server remotely instead

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
...
+      "default": "",
+      "sensitive": true
+    },
+    "debug": {
+      "type": "boolean",
+      "title": "Debug",
+      "description": "Enable debug mode with verbose logging to stderr. Default: false.",
+      "default": false
+    },
+    "dangerously_disable_elicitation": {
```
wbreza, line 196: [F-001] **High - Security**

The bundle wires `--dangerously-disable-elicitation` to a `user_config` boolean, which turns the elicitation consent gate into a checkbox in the client settings form.

Elicitation is the only in-band consent layer for tools whose metadata marks them `Secret` or `Destructive` (for example Key Vault secret retrieval). `HandleElicitationAsync` short-circuits and proceeds without prompting whenever the option is true (`core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs:206-226`), and every tool invocation routes through that check (`CommandFactoryToolLoader.cs:171`, `NamespaceToolLoader.cs:409`), fed from `ServerStartOptions.DangerouslyDisableElicitation` (`ServerStartOptions.cs:68`) via `ServiceCollectionExtensions.cs:56`.

The option's own documentation states it should only be used "in trusted, automated environments where user interaction is not possible" (`servers/Azure.Mcp.Server/docs/azmcp-commands.md:271-283`). A GUI bundle install in a desktop client is the opposite of that environment, and a single misclick silently disables secret/destructive confirmation for every subsequent tool call.

Comparable stdio-safe dangerous flags are deliberately not exposed here: `--dangerously-write-support-logs-to-dir` (`ServerStartOptions.cs:83`) and `--dangerously-disable-retry-limits` (`ServerStartOptions.cs:91`).

Suggested fix: drop the entry from `user_config` and remove the matching pair from `args`, leaving the flag reachable only through an explicit manual config override. If it must stay, prefix the title with "DANGEROUS" to match the treatment in `azmcp-commands.md`.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
     "quick reference",
     "cli"
   ],
+  "user_config": {
```
wbreza, line 121: [F-002] **High - Repo standards**

This PR adds a user-facing configuration surface but no changelog entry; the diff touches only `manifest.json`.

`docs/changelog-entries.md:67-72` lists "New user-facing features or tools" as a required case, `docs/changelog-entries.md:156` defines the location and naming convention (`servers/<server>/changelog-entries/alias-brief-description.yaml`), and entries are schema-validated at compile time against `eng/schemas/changelog-entry.schema.json:13-29`. The repo checklist carries the same requirement.

Suggested fix - add `servers/Azure.Mcp.Server/changelog-entries/pl4nty-mcpb-user-config.yaml`:

```yaml
changes:
  - section: "Features Added"
    description: "Added user_config to the Azure MCP Server MCPB bundle so clients such as Claude Desktop can configure authentication, telemetry, and server mode settings."
```

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
...
+      "description": "Enable debug mode with verbose logging to stderr. Default: false.",
+      "default": false
+    },
+    "dangerously_disable_elicitation": {
+      "type": "boolean",
+      "title": "Disable User Confirmation (Not Recommended)",
+      "description": "Disable user confirmation (elicitation) before allowing high risk commands to run, such as returning secrets (passwords) from KeyVault. When enabled, tools that handle secrets, credentials, or sensitive data will execute without user confirmation. This removes an important security layer designed to prevent unauthorized access to sensitive information. Only use this option in trusted, automated environments where user interaction isn't possible. Never use this option in production environments or when handling untrusted input. Default: false.",
+      "default": false
+    },
+    "mode": {
```
wbreza, line 202: [F-003] **Medium - API contract**

`user_config.mode` is typed as a bare `string` and interpolated straight into `--mode` at line 36, but the server accepts only four exact values and rejects anything else at startup.

`ServerStartCommand.ValidateMode` (`core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServerStartCommand.cs:204-215`) validates against `ModeTypes` (`core/Microsoft.Mcp.Core/src/Areas/Server/Options/ModeTypes.cs:14,20,25,30`) - `single`, `namespace`, `all`, `consolidated`. Rendered as a free-text box, a typo such as `Namespace` or `default` produces a server that fails to start, and the failure surfaces at launch rather than at configuration time.

The listed values and the `namespace` default are both accurate; only the missing constraint is at issue.

Suggested fix - constrain the field so the client renders a picker:

```json
"mode": {
  "type": "string",
  "title": "Mode",
  "description": "Server mode: namespace, consolidated, all, or single. Default: namespace.",
  "enum": ["namespace", "consolidated", "all", "single"],
  "default": "namespace"
}
```

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
...
+      "description": "Azure cloud environment to connect to. Examples: AzureCloud (default public cloud), AzureChinaCloud, AzureUSGovernment.",
+      "default": ""
+    },
+    "azure_token_credentials": {
+      "type": "string",
+      "title": "Azure Token Credentials",
+      "description": "Credential chain to use for authentication. Values: 'dev' (VS/VSCode/CLI/PowerShell/AzureDeveloperCLI + browser fallback), 'prod' (Workload Identity + Managed Identity, no interactive fallback), or a specific provider name such as 'AzureCliCredential' or 'ManagedIdentityCredential'.",
+      "default": ""
+    },
+    "azure_mcp_collect_telemetry": {
```
wbreza, line 159: [F-004] **Medium - Documentation**

The MCPB section of the server README is install-only (`servers/Azure.Mcp.Server/README.md:829-857`) and does not describe the settings this manifest now exposes, so users get a configuration form with no reference documentation behind it.

There is a concrete interaction worth calling out: the README documents opting out of telemetry by exporting `AZURE_MCP_COLLECT_TELEMETRY=false` (`servers/Azure.Mcp.Server/README.md:1375-1386`), while this manifest always writes that variable from `user_config` at line 51 with a `true` default. For bundle installs the manifest value is what the process sees, so the documented opt-out no longer describes bundle behavior.

Suggested fix: add a Configuration subsection under `### MCPB` with a table mapping each `user_config` key to its target arg or env var, its default, and whether it is sensitive; cross-link the telemetry opt-out guidance to the new toggles and note that bundle-provided values are what the server process receives.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -28,8 +29,31 @@
         }
       },
       "command": "${__dirname}/server/azmcp",
-      "args": [ "server", "start" ],
-      "env": {}
+      "args": [
```
wbreza, line 32: [F-005] **Medium - Description alignment**

The PR description explains the omissions with "Some settings were excluded because bundles only support stdio". That rationale covers just the HTTP-bound options - `--dangerously-disable-http-incoming-auth` and the `UseOnBehalfOf` outgoing-auth strategy are the ones actually gated on transport (`core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServerStartCommand.cs:226-231,274-276`).

Several stdio-safe options are also omitted with no stated reason: `--namespace` (`ServerStartOptions.cs:25`), `--tool` (`:39`), `--dangerously-write-support-logs-to-dir` (`:83`), `--dangerously-disable-retry-limits` (`:90`), `--disable-cache` (`:109`), and `--cloud` (`:97`, exposed only as the `AZURE_CLOUD` env var here). Meanwhile a stdio-safe dangerous option was included rather than excluded (see F-001).

The description also does not mention that a safety-bypass toggle is being elevated into the client settings form, which is the most review-relevant aspect of the change.

Suggested fix: restate the exclusion criterion in the description (HTTP-only options excluded; `--namespace`/`--tool` deferred pending multi-value UI; other dangerous flags intentionally left CLI-only) and call out the elicitation toggle explicitly.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
...
+    "dangerously_disable_elicitation": {
+      "type": "boolean",
+      "title": "Disable User Confirmation (Not Recommended)",
+      "description": "Disable user confirmation (elicitation) before allowing high risk commands to run, such as returning secrets (passwords) from KeyVault. When enabled, tools that handle secrets, credentials, or sensitive data will execute without user confirmation. This removes an important security layer designed to prevent unauthorized access to sensitive information. Only use this option in trusted, automated environments where user interaction isn't possible. Never use this option in production environments or when handling untrusted input. Default: false.",
+      "default": false
+    },
+    "mode": {
+      "type": "string",
+      "title": "Mode",
+      "description": "Server mode: namespace, consolidated, all, or single. Default: namespace.",
```
wbreza, line 205: [F-006] **Medium - Maintainability**

Each `title`, `description` and `default` added here duplicates values that already exist in `ServerStartOptions` (`core/Microsoft.Mcp.Core/src/Areas/Server/Options/ServerStartOptions.cs:18-110`) and in the options table of `servers/Azure.Mcp.Server/docs/azmcp-commands.md:254-270`. Nothing keeps the three in sync: `eng/scripts/Pack-Mcpb.ps1:197` runs `dotnet mcpb validate`, which is schema validation of the manifest only and performs no cross-file comparison.

The drift is already observable at this line: the manifest and `azmcp-commands.md:259` both enumerate four modes, while the option description at `ServerStartOptions.cs:32` still lists only `single`, `namespace` and `all`, omitting `consolidated`.

Suggested fix: add a check under `eng/scripts/` that parses `ServerStartOptions` and asserts every `--flag` referenced in the manifest `args` has a matching `user_config` key with the same default and description, failing CI on mismatch. As a smaller immediate step, update `ServerStartOptions.cs:32` so the enumerations agree.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
...
+      "default": "namespace"
+    },
+    "read_only": {
+      "type": "boolean",
+      "title": "Read Only",
+      "description": "Whether the MCP server should be read-only. If true, no write operations are allowed. Default: false.",
+      "default": false
+    }
+  },
```
wbreza, line 214: [F-007] **Medium - Testing**

The manifest is schema-validated only inside the packaging flow (`eng/scripts/Pack-Mcpb.ps1:195-199`), and that flow does not run for pull requests. `eng/pipelines/pullrequest.yml:11-15` extends the common template without enabling packaging, and the defaults are `PublishTarget: none` / `PackageMCPB: false` (`eng/pipelines/templates/common.yml:5-8,24-26`); the sign-and-pack stage is gated on the publish target (`common.yml:141-143`) and the MCPB job is conditional (`eng/pipelines/templates/jobs/sign-and-pack.yml:131-134`).

The practical effect is that a malformed manifest - now a 227-line file carrying 15 typed settings rather than two static keys - reaches the release pipeline before anything validates it.

Suggested fix: add a Build-stage PR job that runs `dotnet mcpb validate` whenever a `servers/*/mcpb/manifest.json` file is touched.

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://raw.githubusercontent.com/modelcontextprotocol/mcpb/refs/heads/main/schemas/mcpb-manifest-v0.3.schema.json",
```
wbreza, line 2: [F-008] **Low - Supply chain**

The `$schema` reference targets `refs/heads/main`, so the schema this manifest is authored against can change upstream without any change here, making editor and local validation non-reproducible over time. The repo otherwise pins external refs by tag (for example `eng/pipelines/templates/common.yml:47`, `eng/pipelines/templates/1es-redirect.yml:6`), and the MCPB CLI itself is version-pinned at `0.3.5` in `.config/dotnet-tools.json:5-7`.

Scope note: this URL is consumed by schema-aware editors and by `dotnet mcpb validate`; it is not fetched at server runtime and does not affect the contents of the shipped bundle, which is why this is rated Low.

Suggested fix: pin to the tag matching the declared `manifest_version` of `0.3`:

```json
"$schema": "https://raw.githubusercontent.com/modelcontextprotocol/mcpb/refs/tags/v0.3.0/schemas/mcpb-manifest-v0.3.schema.json",
```

servers/Azure.Mcp.Server/mcpb/manifest.json
```diff
@@ -94,6 +118,100 @@
     "quick reference",
     "cli"
   ],
+  "user_config": {
+    "azure_tenant_id": {
```
wbreza, line 122: [O-001] **Optional**

Clients render `user_config` in declaration order, and the current ordering interleaves the four `args`-driven behavior toggles with the eleven env-driven identity and telemetry values - `azure_mcp_only_use_broker_credential` sits between the telemetry and OTLP entries, while `mode`, `read_only`, `debug` and `dangerously_disable_elicitation` trail after the Application Insights field.

Grouping them by concern mirrors the split the manifest already makes between `env` and `args` and gives a settings pane that scans top to bottom. This is key ordering within one JSON object - no wiring changes.

Suggested order: Azure identity (`azure_tenant_id`, `azure_client_id`, `azure_client_secret`, `azure_subscription_id`, `azure_token_credentials`), then cloud and observability (`azure_cloud`, both telemetry toggles, `azure_mcp_only_use_broker_credential`, `azure_mcp_enable_otlp_exporter`, `applicationinsights_connection_string`), then server behavior (`mode`, `read_only`, `debug`, `dangerously_disable_elicitation`).

There is no prior `user_config` convention in the repo to compare against - this manifest is the only one - so this is a greenfield ordering choice rather than a departure.

wbreza (commented):
Flag and env var contracts all check out; the elicitation toggle, missing changelog, and stale docs need attention before merge.

### What stood out
- **Contracts verified end to end** — every flag name, boolean arity, env var spelling, mode value and default in the manifest matches the server implementation; `--flag false` binds correctly rather than flipping true.
- **Release hygiene is the weak spot** — a user-facing feature lands with no changelog entry and no update to the bundle documentation.

### Top issues
- [F-001] Elicitation consent bypass surfaced as a settings-form checkbox — `servers/Azure.Mcp.Server/mcpb/manifest.json:196`
- [F-002] User-facing feature added without the required changelog entry — `servers/Azure.Mcp.Server/mcpb/manifest.json:121`
- [F-003] `mode` passed to `--mode` as unconstrained free text — `servers/Azure.Mcp.Server/mcpb/manifest.json:202`

### Cross-cutting
- [F-004] **Bundle documentation does not cover the new settings** _(docs)_ — README MCPB section is install-only; telemetry opt-out guidance now interacts with a bundle toggle.
- [F-006] **Option metadata duplicated across three files with no sync guard** _(process)_ — the mode wording already disagrees between manifest and options source.
- [F-007] **Manifest schema validation does not run on pull requests** _(testing)_ — `mcpb validate` only executes in the packaging stage, which PR builds disable.

### Improvement opportunities
- [O-001] Group `user_config` entries by concern so the settings form reads coherently _(Optional)_ — `servers/Azure.Mcp.Server/mcpb/manifest.json:122`

PR #2636, opened 2026-05-12
feat(datafactory): bridge host auth to DataFactory.MCP.Core via TokenCredential

Ebram-Tawfik:
## Summary

Bridges the Fabric MCP Server's authentication to DataFactory.MCP.Core so that Data Factory commands automatically use the host's IAzureTokenCredentialProvider instead of requiring a separate login flow.

## Changes

- **DataFactoryAreaSetup.cs** — Register TokenCredential from IAzureTokenCredentialProvider before calling AddDataFactoryMcpServices(). DataFactory.MCP.Core v0.21.0 auto-detects a TokenCredential in DI and uses it via TokenCredentialAuthenticationService.
- **Directory.Packages.props** — Bump Microsoft.DataFactory.MCP.Core from 0.20.0-beta to 0.21.0-preview (includes the TokenCredentialAuthenticationService adapter).

## Auth Flow

Fabric MCP Server (DefaultAzureCredential)
-> IAzureTokenCredentialProvider
-> TokenCredential (registered by DataFactoryAreaSetup)
-> TokenCredentialAuthenticationService (auto-detected)
-> FabricAuthenticationHandler -> Bearer token on API calls

## Why

Without this bridge, Data Factory commands fail at auth time because FabricAuthenticationHandler calls IAuthenticationService.GetAccessTokenAsync() which requires an explicit login that never happens in the Fabric MCP Server context. Other Fabric tools (OneLake, Core) accept TokenCredential via constructor — this aligns DataFactory with that pattern.

## Related

- DataFactory.MCP.Core PR: https://github.com/microsoft/DataFactory.MCP/pull/102 (adds TokenCredentialAuthenticationService + version 0.21.0)

### DIFF
--- a/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
+++ b/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
@@ -1,12 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Azure.Core;
 using Fabric.Mcp.Tools.DataFactory.Commands.Dataflow;
 using Fabric.Mcp.Tools.DataFactory.Commands.Pipeline;
 using global::DataFactory.MCP.Extensions;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Mcp.Core.Areas;
 using Microsoft.Mcp.Core.Commands;
+using Microsoft.Mcp.Core.Services.Azure.Authentication;
 
 namespace Fabric.Mcp.Tools.DataFactory;
 
@@ -17,7 +20,16 @@ public class DataFactoryAreaSetup : IAreaSetup
 
     public void ConfigureServices(IServiceCollection services)
     {
-        // Register global::DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
+        // Bridge host auth to DataFactory.MCP.Core: resolve TokenCredential from the host's
+        // IAzureTokenCredentialProvider so DataFactory services auto-use host authentication
+        services.TryAddSingleton<TokenCredential>(sp =>
+        {
+            var provider = sp.GetRequiredService<IAzureTokenCredentialProvider>();
+            return provider.GetTokenCredentialAsync(tenantId: null, CancellationToken.None)
+                .GetAwaiter().GetResult();
+        });
+
+        // Register DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
         services.AddDataFactoryMcpServices();
 
         // Register command instances
--- a/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
+++ b/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
@@ -20,10 +20,15 @@ public class DataFactoryAreaSetup : IAreaSetup
 
     public void ConfigureServices(IServiceCollection services)
     {
-        // Bridge host auth to DataFactory.MCP.Core via a delegating TokenCredential that
-        // calls through to IAzureTokenCredentialProvider on each token request. This avoids
-        // caching a single credential (which breaks OBO/multi-user) and sync-over-async.
-        services.TryAddSingleton<TokenCredential, ProviderDelegatingCredential>();
+        // Bridge host auth to DataFactory.MCP.Core: only register TokenCredential if the host
+        // provides IAzureTokenCredentialProvider. DataFactory.MCP.Core auto-detects TokenCredential
+        // in DI — if present, uses it; otherwise falls back to standalone auth.
+        // The delegating credential calls through to the provider on each token request,
+        // avoiding credential caching (which breaks OBO/multi-user).
+        if (services.Any(d => d.ServiceType == typeof(IAzureTokenCredentialProvider)))
+        {
+            services.TryAddSingleton<TokenCredential, ProviderDelegatingCredential>();
+        }
 
         // Register DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
         services.AddDataFactoryMcpServices();
--- a/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
+++ b/tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
@@ -20,14 +20,10 @@ public class DataFactoryAreaSetup : IAreaSetup
 
     public void ConfigureServices(IServiceCollection services)
     {
-        // Bridge host auth to DataFactory.MCP.Core: resolve TokenCredential from the host's
-        // IAzureTokenCredentialProvider so DataFactory services auto-use host authentication
-        services.TryAddSingleton<TokenCredential>(sp =>
-        {
-            var provider = sp.GetRequiredService<IAzureTokenCredentialProvider>();
-            return provider.GetTokenCredentialAsync(tenantId: null, CancellationToken.None)
-                .GetAwaiter().GetResult();
-        });
+        // Bridge host auth to DataFactory.MCP.Core via a delegating TokenCredential that
+        // calls through to IAzureTokenCredentialProvider on each token request. This avoids
+        // caching a single credential (which breaks OBO/multi-user) and sync-over-async.
+        services.TryAddSingleton<TokenCredential, ProviderDelegatingCredential>();
 
         // Register DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
         services.AddDataFactoryMcpServices();
--- a/tools/Fabric.Mcp.Tools.DataFactory/tests/Fabric.Mcp.Tools.DataFactory.UnitTests/ProviderDelegatingCredentialTests.cs
+++ b/tools/Fabric.Mcp.Tools.DataFactory/tests/Fabric.Mcp.Tools.DataFactory.UnitTests/ProviderDelegatingCredentialTests.cs
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Mcp.Core.Services.Azure.Authentication;
+using NSubstitute;
+
+namespace Fabric.Mcp.Tools.DataFactory.UnitTests;
+
+public class ProviderDelegatingCredentialTests
+{
+    [Fact]
+    public async Task GetTokenAsync_DelegatesToProviderOnEachCall()
+    {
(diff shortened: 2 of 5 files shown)


tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
```diff
@@ -17,7 +20,16 @@ public class DataFactoryAreaSetup : IAreaSetup
...
     {
-        // Register global::DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
+        // Bridge host auth to DataFactory.MCP.Core: resolve TokenCredential from the host's
+        // IAzureTokenCredentialProvider so DataFactory services auto-use host authentication
+        services.TryAddSingleton<TokenCredential>(sp =>
+        {
+            var provider = sp.GetRequiredService<IAzureTokenCredentialProvider>();
+            return provider.GetTokenCredentialAsync(tenantId: null, CancellationToken.None)
+                .GetAwaiter().GetResult();
+        });
```
jongio: This singleton factory resolves the credential once and caches it forever. In OBO mode (`HttpOnBehalfOfTokenCredentialProvider`), each call to `GetTokenCredentialAsync` returns a different credential scoped to the current HTTP request's user. Caching the first one means all subsequent requests use the first user's credential - a cross-user token leak.

Register a delegating `TokenCredential` wrapper instead (see review body for pattern). The wrapper is itself a singleton, but each `GetTokenAsync` call flows through the provider's per-request resolution.

jongio (commented):
Two issues with the auth bridging approach that need resolution before this can merge:

1. **Singleton caching breaks OBO/remote mode**: `IAzureTokenCredentialProvider.GetTokenCredentialAsync` is designed to return per-execution-context credentials (see the interface xmldoc). Caching the first resolved `TokenCredential` in a singleton means all subsequent users share one credential. If the Fabric server ever runs in HTTP mode with OBO, this is a correctness bug.

2. **Sync-over-async in DI factory**: `.GetAwaiter().GetResult()` can deadlock on platforms with a `SynchronizationContext` and is generally an anti-pattern in DI registration.

The fix is to register a delegating `TokenCredential` subclass that wraps `IAzureTokenCredentialProvider` and calls through to the provider on each `GetToken`/`GetTokenAsync` invocation - no caching, no blocking. Something like:

```csharp
internal sealed class ProviderDelegatingCredential(IAzureTokenCredentialProvider provider) : TokenCredential
{
    public override AccessToken GetToken(TokenRequestContext ctx, CancellationToken ct)
        => GetTokenAsync(ctx, ct).GetAwaiter().GetResult();

    public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext ctx, CancellationToken ct)
    {
        var credential = await provider.GetTokenCredentialAsync(tenantId: null, ct);
        return await credential.GetTokenAsync(ctx, ct);
    }
}
```

Then register: `services.TryAddSingleton<TokenCredential, ProviderDelegatingCredential>();`

This preserves OBO correctness (provider resolves per-context) and avoids sync-over-async in the factory.

Also: all CI builds are currently failing - worth checking if the `0.21.0-preview` package is available on the feed.

tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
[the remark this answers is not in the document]
Ebram-Tawfik: Fixed in 014be5b — replaced the singleton factory with a `ProviderDelegatingCredential` that extends `TokenCredential` and delegates to `IAzureTokenCredentialProvider.GetTokenCredentialAsync()` on every `GetTokenAsync` call. No caching, no sync-over-async in DI registration. The wrapper itself is a singleton, but each token request flows through the provider per-context resolution.

tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
```diff
@@ -17,7 +20,16 @@ public class DataFactoryAreaSetup : IAreaSetup
...
+
+        // Register DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
         services.AddDataFactoryMcpServices();
```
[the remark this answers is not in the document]
Ebram-Tawfik: Good call — will add a unit test that registers a stub `IAzureTokenCredentialProvider`, builds the service provider, resolves `TokenCredential`, and verifies that each `GetTokenAsync` call flows through the provider (not cached).

tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
Ebram-Tawfik: Fixed in 014be5b — this factory is replaced with `ProviderDelegatingCredential`, a `TokenCredential` subclass that calls `IAzureTokenCredentialProvider.GetTokenCredentialAsync()` on each `GetTokenAsync` invocation. No credential caching, OBO correctness preserved.

jongio (approved):
Addresses my previous feedback cleanly. The delegating `TokenCredential` wrapper is the right pattern - per-call resolution preserves OBO correctness without sync-over-async in the DI factory. Tests are solid (proving no caching + context propagation).

AmosHersch (approved):

tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
```diff
@@ -17,7 +20,17 @@ public class DataFactoryAreaSetup : IAreaSetup
...
     {
-        // Register global::DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
+        // Bridge host auth to DataFactory.MCP.Core: only register TokenCredential if the host
+        // provides IAzureTokenCredentialProvider. DataFactory.MCP.Core auto-detects TokenCredential
+        // in DI — if present, uses it; otherwise falls back to standalone auth.
+        // The delegating credential calls through to the provider on each token request,
+        // avoiding credential caching (which breaks OBO/multi-user).
+        if (services.Any(d => d.ServiceType == typeof(IAzureTokenCredentialProvider)))
+        {
+            services.TryAddSingleton<TokenCredential, ProviderDelegatingCredential>();
```
vukelich, line 30: @Ebram-Tawfik I'd like to provide some suggestions on design improvements to be mindful that `TokenCredential` can mean different things to different code paths. For example, it's totally reasonable for a remote MCP server to use `TokenCredential` instances to talk to its own resources (e.g., using its own MI and RBAC to its own Redis cache) versus using `TokenCredential` instances that reflect accessing the user's intended resource for MCP tool invocations.

That being said, this PR is fine for the state of this repo today because the code only expects downstream authentication to be getting data for the MCP client/user. If/when we add functionality for the MCP server to interact with separate downstream auth credentials, then I'd be a little more worried about this possible pollution of the DI container. Such future functionality could be using a Redis cache to manage long-running or distributed MCP tool work.

(Caveat: technically, the remote MCP server will use the Azure Identity SDK + `Microsoft.Identity.Web` to authenticate as the Entra app registration for remote MCP server OBO flows. I do not believe those libraries would consume a `TokenCredential` from DI for that, so I'm not worried on affecting remote scenarios today.) 

Does the `Microsoft.DataFactory.MCP.Core` package have DI configurability to very directly provide `TokenCredential`, such as through [keyed services](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-10.0#keyed-services) or [named options](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-10.0#specify-a-custom-key-name-for-a-configuration-property-using-configurationkeyname)?

I asked Opus 4.7, and [it provided a well-laid out explanation and supports this suggestion](https://claude.ai/share/33438876-918a-406d-9131-c74b38f61e66) for the `Microsoft.DataFactory.MCP.Core` library. Claude also called out the Azure SDK's similar `WithCredential` DI extension method to tightly couple an outbound client instance with the credential that client instance should use.

tools/Fabric.Mcp.Tools.DataFactory/src/DataFactoryAreaSetup.cs
```diff
@@ -17,7 +20,17 @@ public class DataFactoryAreaSetup : IAreaSetup
 
     public void ConfigureServices(IServiceCollection services)
     {
-        // Register global::DataFactory.MCP.Core services (auth, HttpClients, all service implementations)
+        // Bridge host auth to DataFactory.MCP.Core: only register TokenCredential if the host
+        // provides IAzureTokenCredentialProvider. DataFactory.MCP.Core auto-detects TokenCredential
+        // in DI — if present, uses it; otherwise falls back to standalone auth.
+        // The delegating credential calls through to the provider on each token request,
+        // avoiding credential caching (which breaks OBO/multi-user).
+        if (services.Any(d => d.ServiceType == typeof(IAzureTokenCredentialProvider)))
```
vukelich, line 28: When did you see cases where `IAzureTokenCredentialProvider` was _not_ in the DI container? I'd expect that to always be present, but maybe there's a use case I'm not expecting.

vukelich: If we expect it to always be there, then I'd recommend a pattern where we throw an exception to fast-fail. Such failures _should_ be caught in any manual or automated testing and never end up in the customer's hands. Catching unexpected state before we complete pull requests is the best time to fix them.

tools/Fabric.Mcp.Tools.DataFactory/tests/Fabric.Mcp.Tools.DataFactory.UnitTests/ProviderDelegatingCredentialTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        services.AddSingleton(provider);
+        services.AddSingleton<TokenCredential, ProviderDelegatingCredential>();
+
+        await using var sp = services.BuildServiceProvider();
+        var resolved = sp.GetRequiredService<TokenCredential>();
+
+        var context = new TokenRequestContext(["https://api.fabric.microsoft.com/.default"]);
+
+        // Act
+        var token1 = await resolved.GetTokenAsync(context, CancellationToken.None);
```
vukelich, line 39: Please do not use `CancellationToken.None`. Using a meaningful `CancellationToken` allows for responsive ending of unit tests. [As detailed in the `CONTRIBUTING.md`,](https://github.com/microsoft/mcp/blob/c2c34e9d47eec600c5caeab311a604701e36e806/CONTRIBUTING.md?plain=1#L216) use `Xunit.TestContext.Current.CancellationToken`

vukelich (commented):
Late suggestions. Main feedback is around the `Microsoft.DataFactory.MCP.Core` DI extensibility and suggesting a more direct DI contract with the very broad `TokenCredential` type.

PR #2637, opened 2026-05-12
Introduce the new options pattern

hallipr:
## What does this PR do?
This pull request introduces significant refactoring to the Azure MCP command and options infrastructure, primarily to simplify option binding and validation, and to improve consistency across commands. The changes also include updates to retry policy configuration, test adjustments, and new project references.

- Removed SearchService's ConfigureRetryPolicy in favor of the common base class method

## GitHub issue number?

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
+++ b/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
@@ -33,7 +33,7 @@ public abstract class SubscriptionCommand<
     {
         var options = base.BindOptions(parseResult);
         // Always post-process subscription via resolver (env var / CLI profile fallback)
-        options.Subscription = _subscriptionResolver.GetSubscription(parseResult);
+        options.Subscription = _subscriptionResolver.ResolveSubscription(options.Subscription);
         return options;
     }
 }
--- a/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
+++ b/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
@@ -1,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+using Azure.Mcp.Core.Options;
+using Azure.Mcp.Core.Services.Azure.Subscription;
+using Microsoft.Mcp.Core.Commands;
+
+namespace Azure.Mcp.Core.Commands.Subscription;
+
+public abstract class SubscriptionCommand<
+    [DynamicallyAccessedMembers(TrimAnnotations.CommandAnnotations)] TOptions, TResult>
+     : AuthenticatedCommand<TOptions, TResult> where TOptions : class, ISubscriptionOption
+{
+    private readonly ISubscriptionResolver _subscriptionResolver;
+
+    protected SubscriptionCommand(ISubscriptionResolver subscriptionResolver)
+    {
+        _subscriptionResolver = subscriptionResolver;
+    }
+
+    public override void ValidateOptions(TOptions options, ValidationResult validationResult)
+    {
+        base.ValidateOptions(options, validationResult);
+
+        if (string.IsNullOrEmpty(options.Subscription))
+        {
+            validationResult.Errors.Add("Missing Required options: --subscription");
+        }
+    }
+
+    public override TOptions BindOptions(ParseResult parseResult)
+    {
+        var options = base.BindOptions(parseResult);
+        // Always post-process subscription via resolver (env var / CLI profile fallback)
+        options.Subscription = _subscriptionResolver.GetSubscription(parseResult);
+        return options;
+    }
+}
--- a/core/Azure.Mcp.Core/src/Options/OptionDescriptions.cs
+++ b/core/Azure.Mcp.Core/src/Options/OptionDescriptions.cs
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Mcp.Core.Options;
+
+/// <summary>
+/// Description constants for global options shared across all Azure MCP commands.
+/// Used with <see cref="Microsoft.Mcp.Core.Options.OptionAttribute"/> on options classes.
+/// </summary>
+public static class OptionDescriptions
+{
+    public const string Subscription =
+        "Specifies the Azure subscription to use. Accepts either a subscription ID (GUID) or display name. " +
+        "If not specified, the AZURE_SUBSCRIPTION_ID environment variable will be used instead.";
+
+    public const string Tenant =
+        "The Microsoft Entra ID tenant ID or name. " +
+        "This can be either the GUID identifier or the display name of your Entra ID tenant.";
+
+    public const string AuthMethod =
+        "Authentication method to use. " +
+        "Options: 'credential' (Azure CLI/managed identity), 'key' (access key), or 'connectionString'.";
+
+    public const string ResourceGroup =
+        "The name of the Azure resource group. This is a logical container for Azure resources.";
+
+    public const string Scope =
+        "Scope at which the role assignment or definition applies to, " +
+        "e.g., /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333, " +
+        "/subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup, " +
+        "or /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup/providers/Microsoft.Compute/virtualMachines/myVM.";
+}
--- a/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
+++ b/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Azure.Core;
+using Microsoft.Mcp.Core.Options;
+
+namespace Azure.Mcp.Core.Options
+{
+    public static class RetryOptionsExtensions
+    {
+        public static void ConfigureRetryOptions(this ClientOptions clientOptions, RetryPolicyOptions? retryPolicy)
+        {
+            if (retryPolicy == null)
+            {
+                return;
+            }
+
+            if (retryPolicy.MaxRetries.HasValue)
+            {
+                clientOptions.Retry.MaxRetries = retryPolicy.MaxRetries.Value;
+            }
+
+            if (retryPolicy.Mode.HasValue)
+            {
+                clientOptions.Retry.Mode = retryPolicy.Mode.Value;
+            }
+
+            if (retryPolicy.DelaySeconds.HasValue)
+            {
+                clientOptions.Retry.Delay = TimeSpan.FromSeconds(retryPolicy.DelaySeconds.Value);
+            }
+
+            if (retryPolicy.MaxDelaySeconds.HasValue)
+            {
+                clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(retryPolicy.MaxDelaySeconds.Value);
+            }
+
+            if (retryPolicy.NetworkTimeoutSeconds.HasValue)
+            {
+                clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(retryPolicy.NetworkTimeoutSeconds.Value);
+            }
+        }
+    }
+}
--- a/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
+++ b/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
@@ -8,6 +8,7 @@ using Azure.Core;
 using Microsoft.Mcp.Core.Options;
 
 namespace Azure.Mcp.Core.Options;
+
 public static class RetryOptionsExtensions
 {
     public static void ConfigureRetryOptions(this ClientOptions clientOptions, RetryPolicyOptions? retryPolicy)
--- a/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
+++ b/core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
@@ -1,44 +1,45 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
 using System.Collections.Generic;
 using System.Text;
 using Azure.Core;
 using Microsoft.Mcp.Core.Options;
 
-namespace Azure.Mcp.Core.Options
+namespace Azure.Mcp.Core.Options;
+public static class RetryOptionsExtensions
 {
-    public static class RetryOptionsExtensions
+    public static void ConfigureRetryOptions(this ClientOptions clientOptions, RetryPolicyOptions? retryPolicy)
     {
-        public static void ConfigureRetryOptions(this ClientOptions clientOptions, RetryPolicyOptions? retryPolicy)
+        if (retryPolicy == null)
         {
-            if (retryPolicy == null)
-            {
-                return;
-            }
+            return;
+        }
 
-            if (retryPolicy.MaxRetries.HasValue)
-            {
-                clientOptions.Retry.MaxRetries = retryPolicy.MaxRetries.Value;
-            }
+        if (retryPolicy.MaxRetries.HasValue)
+        {
+            clientOptions.Retry.MaxRetries = retryPolicy.MaxRetries.Value;
+        }
 
-            if (retryPolicy.Mode.HasValue)
-            {
-                clientOptions.Retry.Mode = retryPolicy.Mode.Value;
-            }
+        if (retryPolicy.Mode.HasValue)
+        {
+            clientOptions.Retry.Mode = retryPolicy.Mode.Value;
+        }
 
-            if (retryPolicy.DelaySeconds.HasValue)
-            {
-                clientOptions.Retry.Delay = TimeSpan.FromSeconds(retryPolicy.DelaySeconds.Value);
-            }
+        if (retryPolicy.DelaySeconds.HasValue)
+        {
+            clientOptions.Retry.Delay = TimeSpan.FromSeconds(retryPolicy.DelaySeconds.Value);
+        }
 
-            if (retryPolicy.MaxDelaySeconds.HasValue)
-            {
-                clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(retryPolicy.MaxDelaySeconds.Value);
-            }
+        if (retryPolicy.MaxDelaySeconds.HasValue)
+        {
+            clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(retryPolicy.MaxDelaySeconds.Value);
+        }
 
-            if (retryPolicy.NetworkTimeoutSeconds.HasValue)
-            {
-                clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(retryPolicy.NetworkTimeoutSeconds.Value);
-            }
+        if (retryPolicy.NetworkTimeoutSeconds.HasValue)
+        {
+            clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(retryPolicy.NetworkTimeoutSeconds.Value);
         }
     }
 }
--- a/core/Azure.Mcp.Core/src/Services/Azure/Subscription/ISubscriptionResolver.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/Subscription/ISubscriptionResolver.cs
@@ -12,10 +12,10 @@ namespace Azure.Mcp.Core.Services.Azure.Subscription;
 public interface ISubscriptionResolver
 {
     /// <summary>
-    /// Resolves the subscription from the parse result, falling back to Azure CLI profile
+    /// Resolves the subscription from the provided value, falling back to Azure CLI profile
     /// or AZURE_SUBSCRIPTION_ID environment variable.
     /// </summary>
-    string? GetSubscription(ParseResult parseResult);
+    string? ResolveSubscription(string? subscription);
 
     /// <summary>
     /// Checks if a subscription is available from the command option, Azure CLI profile,
--- a/core/Azure.Mcp.Core/src/Services/Azure/Subscription/ISubscriptionResolver.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/Subscription/ISubscriptionResolver.cs
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.CommandLine.Parsing;
+
+namespace Azure.Mcp.Core.Services.Azure.Subscription;
+
+/// <summary>
+/// Resolves subscription values from command-line arguments, environment variables,
+/// and Azure CLI profile, providing a testable seam for subscription resolution.
+/// </summary>
+public interface ISubscriptionResolver
+{
+    /// <summary>
+    /// Resolves the subscription from the parse result, falling back to Azure CLI profile
+    /// or AZURE_SUBSCRIPTION_ID environment variable.
+    /// </summary>
+    string? GetSubscription(ParseResult parseResult);
+
+    /// <summary>
+    /// Checks if a subscription is available from the command option, Azure CLI profile,
+    /// or AZURE_SUBSCRIPTION_ID environment variable.
+    /// </summary>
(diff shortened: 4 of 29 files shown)


core/Azure.Mcp.Core/src/Areas/Subscription/Options/SubscriptionListOptions.cs
```diff
@@ -2,7 +2,19 @@
...
 using Microsoft.Mcp.Core.Options;
+using Microsoft.Mcp.Core.Models.Option;
+using Microsoft.Mcp.Core.Models;
 
 namespace Azure.Mcp.Core.Areas.Subscription.Options;
 
-public class SubscriptionListOptions : GlobalOptions;
+public class SubscriptionListOptions : SubscriptionOptions
+{
+    public string? ResourceGroup { get; set; }
```
alzimmermsft: Doesn't make sense for this to continue having this option

core/Azure.Mcp.Core/src/Areas/Group/Options/BaseGroupOptions.cs
```diff
@@ -2,7 +2,17 @@
...
 namespace Azure.Mcp.Core.Areas.Group.Options;
 
-public class BaseGroupOptions : SubscriptionOptions;
+public class BaseGroupOptions : SubscriptionOptions
+{
+    public string? Tenant { get; set; }
+
+    public AuthMethod? AuthMethod { get; set; }
+
+    [Option(Name = "retry")]
```
alzimmermsft: What happens to the `Name` here given that this is annotated on a model which has another set of values?

core/Azure.Mcp.Core/src/Areas/Subscription/Options/SubscriptionListOptions.cs
hallipr: true, but I didn't want to affect the user-facing api in this PR.  All options are preserved with their existing "requiredness", name and description

core/Azure.Mcp.Core/src/Areas/Group/Options/BaseGroupOptions.cs
hallipr: The name becomes a prefix to the properties of the RetrtyPolicyOptions class.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.UnitTests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
```diff
@@ -15,6 +15,7 @@
 using NSubstitute;
 using Xunit;
 
+using Microsoft.Mcp.Core.Areas.Server.Models;
```
alzimmermsft: Seems to be in an odd spot

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -128,62 +92,42 @@ private static void ValidateSupportLoggingFolder(CommandResult commandResult)
...
+        // Apply defaults for unspecified options
+        options.Mode ??= ModeTypes.NamespaceProxy;
+        options.Transport ??= TransportTypes.StdIo;
+        options.OutgoingAuthStrategy ??= OutgoingAuthStrategy.NotSet;
+        var disableIncomingAuth = options.DangerouslyDisableHttpIncomingAuth ?? false;
+
+        if (options.Tool != null && options.Tool.Length > 0)
+        {
+            options.Mode = ModeTypes.All;
+        }
```
alzimmermsft: Well, this is really interesting behavior... doesn't seem right for another option to implicitly override another like this. Not an issue in this PR but something that probably should be discussed as a follow up.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
```diff
@@ -299,42 +243,38 @@ private static void ValidateTransportConfiguration(string transport, bool httpIn
...
+    /// <param name="validationResult">Validation result to update with validation errors.</param>
+    private static void ValidateNamespaceAndToolMutualExclusion(string[]? namespaces, string[]? tools, ValidationResult validationResult)
     {
         bool hasNamespace = namespaces != null && namespaces.Length > 0;
         bool hasTool = tools != null && tools.Length > 0;
 
         if (hasNamespace && hasTool)
         {
-            commandResult.AddError("The --namespace and --tool options cannot be used together. Please specify either --namespace to filter by service namespace or --tool to filter by specific tool names, but not both.");
+            validationResult.AddError("The --namespace and --tool options cannot be used together. Please specify either --namespace to filter by service namespace or --tool to filter by specific tool names, but not both.");
```
alzimmermsft: Comment for tracking, not an issue in this PR.

Using `--namespace`, etc. ties the error message to a CLI-like mode. MCP parameters don't use `--` prefixes and would just be `namespace` here.

Can we write an issue to clean this up so that our error messages return based on the mode used (CLI or MCP right now). In CLI we use `--namespace ... options` and in MCP we use `namespace ... parameters`.

core/Microsoft.Mcp.Core/src/Areas/Server/Options/ServiceStartOptions.cs
```diff
@@ -14,98 +14,84 @@ public class ServiceStartOptions
...
-    public string[]? Namespace { get; set; } = null;
+    public string[]? Namespace { get; set; }
 
     /// <summary>
     /// Gets or sets the mode mode for the server.
     /// Defaults to 'namespace' mode which exposes one tool per service namespace.
     /// </summary>
-    [JsonPropertyName("mode")]
-    public string? Mode { get; set; } = ModeTypes.NamespaceProxy;
+    public string? Mode { get; set; }
```
alzimmermsft: Another thing noticed due to changes here but isn't due to the PR.

We should make an enum for `Mode` instead of is being a `string`

core/Microsoft.Mcp.Core/src/Commands/CommandFactory.cs
```diff
@@ -358,6 +351,19 @@ private void ConfigureCommandHandler(Command command, IBaseCommand implementatio
 
                 return (int)response.Status;
             }
+            catch (CommandValidationException ex)
```
alzimmermsft: When does this get thrown here? It seems that `BaseCommand` should handle catching `CommandValidationException` and using `HandleException` to process it by having `BindOptions` and `ValidateOptions` in the same block.

core/Microsoft.Mcp.Core/src/Commands/ValidationResult.cs
```diff
@@ -1,11 +1,25 @@
...
+
 namespace Microsoft.Mcp.Core.Commands;
 
 public class ValidationResult
 {
     public bool IsValid => Errors.Count == 0;
 
     public List<string> Errors { get; } = [];
+
+    public void AddError(string error) => Errors.Add(error);
```
alzimmermsft: If we're adding this, should we make `Errors` a private field instead of a property?

core/Microsoft.Mcp.Core/analyzers/Microsoft.Mcp.Core.Analyzers/OptionalGroupAnalyzer.cs
```diff
@@ -0,0 +1,235 @@
...
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class OptionalGroupAnalyzer : DiagnosticAnalyzer
+{
+    public const string DiagnosticId = "MCP0001";
+
+    private static readonly DiagnosticDescriptor Rule = new(
+        DiagnosticId,
+        title: "Optional group contains required member",
+        messageFormat: "Nullable complex property '{0}' contains non-nullable member '{1}'. All properties within a nullable complex type must be nullable.",
```
alzimmermsft: Does this need to handle chains more deeply nested than one level? For example, Foo.Bar.Baz where `Foo` is nullable, `Bar` is nullable, but `Baz` is non-null.

core/Microsoft.Mcp.Core/analyzers/Microsoft.Mcp.Core.Analyzers/OptionalGroupAnalyzer.cs
```diff
@@ -0,0 +1,235 @@
...
+        ValidateOptionsType(context, optionsType, parentPropertyName: null, parentLocation: null);
+    }
+
+    private static INamedTypeSymbol? GetOptionsTypeFromBaseCommand(INamedTypeSymbol type)
+    {
+        INamedTypeSymbol? current = type.BaseType;
+        while (current is not null)
+        {
+            if (current.IsGenericType &&
+                current.Name == "BaseCommand" &&
```
alzimmermsft: Can be a follow-up but it also seems that we should have a rule where a command never implements `IBaseCommand` directly, as the current state of this PR that wouldn't function properly.

Or the design in this PR needs to be changed so that `BaseCommand` doesn't contain common implementation needed by all command.

core/Microsoft.Mcp.Core/src/Models/Option/OptionInfo.cs
```diff
@@ -1,12 +1,40 @@
...
+    [JsonPropertyName("default")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? DefaultValue { get; set; } = defaultValue;
+
+    [JsonPropertyName("type")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string Type { get; set; } = "string";
+
+    [JsonPropertyName("required")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
```
alzimmermsft: Should we be ignoring this in JSON serialization when the value is `false`?

core/Azure.Mcp.Core/src/Services/Azure/Subscription/ISubscriptionResolver.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Resolves subscription values from command-line arguments, environment variables,
+/// and Azure CLI profile, providing a testable seam for subscription resolution.
+/// </summary>
+public interface ISubscriptionResolver
+{
+    /// <summary>
+    /// Resolves the subscription from the parse result, falling back to Azure CLI profile
+    /// or AZURE_SUBSCRIPTION_ID environment variable.
```
alzimmermsft: This really should be the docstring for the default implementation, not the interface. The interface should allow the implementation to do whatever it wants.

core/Microsoft.Mcp.Core/src/Options/OptionAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// Override the CLI option name (without the "--" prefix).
+    /// When null, the name is derived from the property name in kebab-case.
+    /// </summary>
+    public string? Name { get; init; }
+
+    /// <summary>
+    /// A description of what the option controls. Used in help text and by AI agents.
+    /// </summary>
+    public string? Description { get; init; }
```
alzimmermsft, line 46: I thought the plan was for `Description` to be non-null

core/Microsoft.Mcp.Core/src/Options/OptionDescriptor.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    [UnconditionalSuppressMessage("Trimming", "IL2070:UnrecognizedReflectionPattern",
+        Justification = "Nested option types are rooted by the application.")]
+    private static void CollectDescriptors(Type type, string? prefix, List<OptionDescriptor> descriptors, NullabilityInfoContext nullabilityContext, bool parentOptional = false, PropertyInfo? parentProperty = null)
+    {
+        PropertyInfo[] allProperties = type.GetProperties(BindingFlags.Public | BindingFlags.Instance);
+
+        // When a derived class hides a base property with 'new', GetProperties returns both.
+        // Keep only the most-derived version (the one whose DeclaringType is closest to 'type').
```
alzimmermsft, line 35: Is this a behavior handled by how `type.GetProperties` works where it checks super types first for their properties? If not, the check below seems wrong as it just checks `IsAssignableFrom` without checking depth.

core/Azure.Mcp.Core/src/Areas/Group/Options/BaseGroupOptions.cs
vcolin7: And the parent object in the `inputSchema` would be `retry: {}`? I don't think that would be clear to the LLM. Maybe we can have `Name` override the property name and a different annotation for the prefix on child properties. Also, would the current prefix apply only to direct children or nested ones too?

core/Microsoft.Mcp.Core/analyzers/Microsoft.Mcp.Core.Analyzers/OptionalGroupAnalyzer.cs
vcolin7: Maybe this checks recursively?

core/Azure.Mcp.Core/src/Areas/Group/Options/BaseGroupOptions.cs
hallipr: The current inputSchema just displays a flat list.  I would expect that to continue.
```json
"inputSchema": {
    "type": "object",
    "properties": {
        ...
        "retry-delay": {
            "type": "number",
            "description": "Initial delay in ..."
        },
        "retry-max-delay": {
            "type": "number",
            "description": "Maximum delay ..."
        },
        "retry-max-retries": {
            "type": "integer",
            "description": "Maximum number ..."
        },
```

core/Azure.Mcp.Core/src/Options/OptionDescriptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Description constants for global options shared across all Azure MCP commands.
+/// Used with <see cref="Microsoft.Mcp.Core.Options.OptionAttribute"/> on options classes.
+/// </summary>
+public static class OptionDescriptions
+{
+    public const string Subscription =
+        "Specifies the Azure subscription to use. Accepts either a subscription ID (GUID) or display name. " +
+        "If not specified, the AZURE_SUBSCRIPTION_ID environment variable will be used instead.";
```
alzimmermsft, line 14: Can be done in a follow-up PR, should add in that the logged in Azure account's default subscription will be used before the `AZURE_SUBSCRIPTION_ID` fallback.

core/Azure.Mcp.Core/src/Options/OptionDescriptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public const string Tenant =
```
alzimmermsft, line 16: Not a PR problem, but something we should file an issue for, this should also fallback to the logged in Azure account's default tenant.

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
alzimmermsft, line 1: Missing copyright

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Azure.Core;
+using Microsoft.Mcp.Core.Options;
+
+namespace Azure.Mcp.Core.Options
```
alzimmermsft: This uses a scoped namespace block rather, which differs from the other files added.

tools/Azure.Mcp.Tools.AppConfig/tests/Azure.Mcp.Tools.AppConfig.UnitTests/KeyValue/Lock/KeyValueLockSetCommandTests.cs
```diff
@@ -15,6 +17,11 @@ namespace Azure.Mcp.Tools.AppConfig.UnitTests.KeyValue.Lock;
 
 public class KeyValueLockSetCommandTests : CommandUnitTestsBase<KeyValueLockSetCommand, IAppConfigService>
 {
+    public KeyValueLockSetCommandTests()
+    {
+        Services.AddSingleton<ISubscriptionResolver>(SubscriptionResolver.Instance);
```
alzimmermsft: Should just make this part of `CommandUnitTestBase`'s default setup but have this return a no-op `Substitute` which is a protected property on `CommandUnitTestBase` so that special handling can be injected when needed.

hallipr: maybe.  I'm a proponent of tests being really obvious.  ideally, there would be no need for CommandUnitTestBase because each test class fully encapsulates what it means to test the command

hallipr: And subscription resolution is an Azure-only concern

hallipr: I switched to substitute to prevent env variables from affecting tests

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceStartCommand.cs
hallipr: We also use the names in command and option descriptions.  These are constants now and would need to be changed by the command description -> command line -> MCP converters.   Token replacement could work but would need a linter/analyzer to ensure all the tokens are registered/convertable

core/Microsoft.Mcp.Core/src/Commands/ValidationResult.cs
hallipr: It may still be useful for inspection along the stack

core/Microsoft.Mcp.Core/analyzers/Microsoft.Mcp.Core.Analyzers/OptionalGroupAnalyzer.cs
hallipr: Removed in this PR.  I'll check when I add it back in the next pr

core/Microsoft.Mcp.Core/analyzers/Microsoft.Mcp.Core.Analyzers/OptionalGroupAnalyzer.cs
hallipr: will address in analyzer pr

core/Microsoft.Mcp.Core/src/Options/OptionDescriptor.cs
hallipr: It's a mix of both.   GetProperties gets all the properties, deeply.  Working by name, it replaces a deeper property with one from a more derived class.

core/Microsoft.Mcp.Core/src/Options/OptionAttribute.cs
hallipr: There's no reason to add descriptions to complex objects as they only contribute a prefix to their deeper options

core/Microsoft.Mcp.Core/src/Commands/CommandFactory.cs
hallipr: no longer in this pr

tools/Azure.Mcp.Tools.Search/src/Services/SearchService.cs
```diff
@@ -392,18 +392,6 @@ private static async Task<List<JsonElement>> ProcessSearchResults(Response<Searc
         return results;
```
jongio, line 392: Removing the private `ConfigureRetryPolicy` method means the two call sites (lines 219 and 327) now silently resolve to `BaseAzureService.ConfigureRetryPolicy`, which applies retry clamping (MaxAllowedRetries=10, MaxAllowedDelaySeconds=60, etc.). The old private method applied retry values directly without bounds.

This is probably an improvement — the private method was bypassing safety limits that exist for a reason — but it's a behavior change worth calling out explicitly in the PR description since it affects retry behavior for all Search operations.

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Azure.Core;
+using Microsoft.Mcp.Core.Options;
+
+namespace Azure.Mcp.Core.Options;
+
+public static class RetryOptionsExtensions
+{
+    public static void ConfigureRetryOptions(this ClientOptions clientOptions, RetryPolicyOptions? retryPolicy)
```
jongio, line 14: There are now two methods that configure retry on `ClientOptions`:

- `RetryOptionsExtensions.ConfigureRetryOptions` (this one) — no clamping, applies values directly
- `BaseAzureService.ConfigureRetryPolicy` — has safety bounds (MaxAllowedRetries=10, etc.)

MonitorService uses the unclamped version here, while SearchService (after the private method removal) falls through to the clamped version. Future tool authors picking between these could easily choose the wrong one. Consider either: (a) adding a doc comment noting the difference, or (b) consolidating into one method that optionally clamps.

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
```
jongio, line 6: `System`, `System.Collections.Generic`, and `System.Text` are unused here.

jongio (commented):
Clean refactoring of the options binding infrastructure. The attribute-driven pattern via `OptionBinder`/`OptionDescriptor` is a solid replacement for the manual `RegisterOptions`/`BindOptions` overrides, and the nullable `RetryPolicyOptions` simplification removes a lot of boilerplate.

A few things I noticed that existing reviewers haven't covered:

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
hallipr: Extra implementation removed

core/Azure.Mcp.Core/src/Options/RetryOptionsExtensions.cs
hallipr: File removed

tools/Azure.Mcp.Tools.AppConfig/src/Options/KeyValue/Lock/KeyValueLockSetOptions.cs
```diff
@@ -1,12 +1,31 @@
...
+    [Option("The name of the key to access within the App Configuration store.")]
+    public required string Key { get; set; }
+
+    [Option("The label to apply to the configuration key. Labels are used to group and organize settings.")]
+    public string? Label { get; set; }
+
+    [Option(OptionDescriptions.Tenant)]
+    public string? Tenant { get; set; }
+
+    [Option(Name = "retry")]
```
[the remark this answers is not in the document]
hallipr: The flattened / prefixed names match what's already in use by all consuming commands

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
```diff
@@ -0,0 +1,39 @@
...
+        _subscriptionResolver = subscriptionResolver;
+    }
+
+    public override void ValidateOptions(TOptions options, ValidationResult validationResult)
+    {
+        base.ValidateOptions(options, validationResult);
+
+        if (string.IsNullOrEmpty(options.Subscription))
+        {
+            validationResult.Errors.Add("Missing Required options: --subscription");
```
alzimmermsft, line 28: Future change, from seeing this it makes me think we should have two buckets for errors. Missing options and validation issues. This would allow for the error message to be constructed in a better manner.

For example, mentioned either somewhere else in this PR or a previous prototype of this PR, how we convey the name of the option/parameter differs by running mode (CLI vs MCP). Having a specialized path for missing options could allow for us to pass the internal name used in code and when generating the error message to be returned it will format the name for the mode (`--option` for CLI and `option` for MCP).

alzimmermsft (approved):

PR #2654, opened 2026-05-18
feat(compute): unified VM/VMSS create with discovery tools and VMSS Flex default

deankroker:
## TL;DR

Adds four read-only Azure discovery tools and makes VMSS Flex the default shape for a new `compute create` command. **Purely additive — every existing `compute vm` and `compute vmss` invocation works unchanged.**

|  |  |
|---|---|
| New commands | 5 (`compute create` + 4 discovery aliases at the `compute` root) |
| Modified commands | `compute vmss create` gains network parity flags |
| Breaking changes | None |
| Tests | New unit tests across all four discovery commands and the unified create; 2 live-gated smoke tests |

> **VMSS Flex** = Virtual Machine Scale Set in Flexible orchestration mode — manages individual VMs as a group, supports availability-zone spread, and scales from 1 to N without rebuild.

## Why

Agents asked to "create a VM" today guess SKUs, images, quotas, and regions before calling create — and they pick a single VM when the user almost always wants something that can scale. This PR adds the discovery tools that let the agent verify before committing, and shifts the default so VMSS Flex is the recommended path while single VM remains an explicit fallback.

### Commit 1 — `feat(compute): add guided VM/VMSS create discovery tools`

Four new read-only discovery tools under `compute vm`. All use the Azure SDK directly (no `az` CLI subprocess), so they inherit auth, RBAC, and retry behavior from the MCP host:

- `list-skus` — VM sizes available in a region, with vCPU / memory / AZ / AcceleratedNetworking filters
- `list-images` — marketplace images by alias or publisher/offer/sku
- `check-quota` — current vs limit on the SKU family the caller is about to deploy into
- `recommend-region` — region scoring with workload-hint, geography, and AZ requirements

These are intended as validation gates immediately before a create call.

### Commit 2 — `feat(compute): reframe VMSS Flex as the default + unified compute create`

- **New `compute create` top-level command.** Defaults to VMSS Flex, 2 instances, AZ-spread. `--single-instance` falls back to a single VM.
- **Discovery aliases at the root.** The four tools from Commit 1 are also registered directly under `compute` so agents don't have to descend into `vm/`.
- **Network parity for `compute vmss create`.** Adds `--public-ip-address`, `--network-security-group`, `--no-public-ip`, and `--source-address-prefix` so picking the recommended path no longer silently downgrades networking.
- **The same four discovery tools gain four enrichments for the guided flow:**
  - `list-skus` surfaces `vmScaleSetsSupported` per SKU
  - `list-images` adds `--include-shared-gallery` for Compute Gallery image refs
  - `check-quota` surfaces VMSS-relevant dimensions (`virtualMachineScaleSets`, `lowPriorityCores`, `availabilitySets`)
  - `recommend-region` weighs AZ-rich regions more heavily on scale / HA / production / vmss hints

## Customer experience

Ask the agent for a VM → discovery tools confirm a SKU exists in the region with quota headroom and scale-set support, in an AZ-rich region → `compute create` returns a Flex scale set, 2 instances, AZ-spread, with public IP / NSG support (not just defaults). Only need one box? `--single-instance`. Need fifty next quarter? Same resource, no rebuild.

## Sample agent flow

```text
user:  give me a VM in westus2 for a web app
agent → compute recommend-region    --workload web
agent → compute list-skus westus2   --min-vcpu 2 --vmss-supported
agent → compute check-quota westus2 --sku Standard_D2s_v5
agent → compute create              --location westus2 --size Standard_D2s_v5
        → VMSS Flex, 2 instances, AZ-spread, public IP + NSG
```

## Test plan

- [x] Unit tests for all four discovery commands (`VmSkuListCommandTests`, `VmImageListCommandTests`, `VmQuotaCheckCommandTests`, `VmRegionRecommendCommandTests`), including the new `--include-shared-gallery` flag, VMSS quota dimensions, and AZ-weighting heuristic
- [x] Unit tests for `UnifiedCreateCommand` covering both the VMSS Flex default and the `--single-instance` fallback
- [x] Unit tests for the four new VMSS network options
- [x] Live test smoke (gated behind opt-in env var; **not run in this PR — invocations covered by** `Should_create_with_unified_create_defaults_to_vmss_flex` and `Should_create_with_unified_create_single_instance_falls_back_to_vm` in `Azure.Mcp.Tools.Compute.Tests/ComputeCommandTests.cs` lines 1509–1584)
- [x] MCP client smoke — `ComputeSetupTests` in `Azure.Mcp.Tools.Compute.Tests/ComputeSetupTests.cs` asserts that:
  - `compute create` registers at the root
  - the four discovery aliases (`list-skus`, `list-images`, `check-quota`, `recommend-region`) resolve directly under `compute` **and** share class instances with their `compute vm` counterparts (stable `CommandMetadata.Id` GUIDs)
  - the `vm` / `vmss` / `disk` subgroups still resolve
  - the group description leads with the VMSS Flex recommendation


deankroker:
@microsoft-github-policy-service agree company="Microsoft"

PR #2674, opened 2026-05-19
Fix Storage Tools Issues.

g2vinay:
Fixed three bugs in `storage account create`:
      - Subscription display names (e.g. "My Subscription") now resolve correctly instead of failing with a FormatException — consistent with all other storage commands.
      - The `--access-tier` option now documents all valid values (Hot, Cool, Cold, Premium); Cold and Premium were silently accepted but never listed, making them undiscoverable.
      - The `--sku` option description now clarifies that `storage account create` creates StorageV2 accounts and defaults to Standard_LRS if no SKU is specified.

### DIFF
--- a/tools/Azure.Mcp.Tools.Storage/src/Options/StorageOptionDefinitions.cs
+++ b/tools/Azure.Mcp.Tools.Storage/src/Options/StorageOptionDefinitions.cs
@@ -36,13 +36,14 @@ public static class StorageOptionDefinitions
 
     public static readonly Option<string> Sku = new($"--{SkuName}")
     {
-        Description = "The storage account SKU. Valid values: Standard_LRS, Standard_GRS, Standard_RAGRS, Standard_ZRS, Premium_LRS, Premium_ZRS, Standard_GZRS, Standard_RAGZRS.",
+        Description = "The storage account SKU for StorageV2 accounts. Valid values: Standard_LRS, Standard_GRS, Standard_RAGRS, Standard_ZRS, " +
+            "Premium_LRS, Premium_ZRS, Standard_GZRS, Standard_RAGZRS. Defaults to Standard_LRS.",
         Required = false
     };
 
     public static readonly Option<string> AccessTier = new($"--{AccessTierName}")
     {
-        Description = "The default access tier for blob storage. Valid values: Hot, Cool.",
+        Description = "The default access tier for blob storage. Valid values: Hot, Cool, Cold, Premium.",
         Required = false
     };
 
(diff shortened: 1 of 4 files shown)


tools/Azure.Mcp.Tools.Storage/src/Options/StorageOptionDefinitions.cs
```diff
@@ -36,13 +36,15 @@ public static class StorageOptionDefinitions
 
     public static readonly Option<string> Sku = new($"--{SkuName}")
     {
-        Description = "The storage account SKU. Valid values: Standard_LRS, Standard_GRS, Standard_RAGRS, Standard_ZRS, Premium_LRS, Premium_ZRS, Standard_GZRS, Standard_RAGZRS.",
+        Description = "The storage account SKU. Valid values: Standard_LRS, Standard_GRS, Standard_RAGRS, Standard_ZRS, " +
```
alzimmermsft: This change is incorrect, instead we should be updating the tool description to indicate we only support StorageV2 accounts. The additional SKUs added here are for different account types.

https://learn.microsoft.com/rest/api/storagerp/srp_sku_types

tools/Azure.Mcp.Tools.Storage/src/Commands/Blob/Container/ContainerCreateCommand.cs
```diff
@@ -65,5 +65,6 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
         return context.Response;
     }
 
+    // Strongly-typed result record
```
alzimmermsft: We don't need this

tools/Azure.Mcp.Tools.Storage/src/Services/StorageService.cs
```diff
@@ -99,8 +100,14 @@ public async Task<StorageAccountResult> CreateStorageAccount(
         // Create ArmClient for deployments
         ArmClient armClient = await CreateArmClientWithApiVersionAsync("Microsoft.Storage/storageAccounts", "2024-01-01", tenant, retryPolicy, cancellationToken);
 
+        // Resolve subscription display name to GUID (consistent with all other storage operations).
+        // Skip resolution when subscription is already a GUID to avoid an unnecessary round-trip.
+        var subscriptionId = _subscriptionService.IsSubscriptionId(subscription)
+            ? subscription
+            : (await _subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken)).Data.SubscriptionId;
```
alzimmermsft, line 106: Do we actually need to resolve to a GUID here?

g2vinay: Yeah, with sub name as input, we currently get back. 
The GUID for subscription is invalid <sub-name>.. To mitigate this

alzimmermsft (approved):

PR #2677, opened 2026-05-20
Fix My SQL Tools Issues.

g2vinay:


servers/Azure.Mcp.Server/changelog-entries/1779224686901.yaml
```diff
@@ -0,0 +1,8 @@
+changes:
+  - section: "Bugs Fixed"
+    description: |
+      Fixed several issues in the MySQL toolset:
+      - `azmcp mysql list` no longer crashes when `--resource-group` is omitted. It now lists all MySQL servers across the subscription using a single ARM API call, matching the behavior of the PostgreSQL toolset.
```
alzimmermsft:
```suggestion
      - `azmcp mysql list` now works when `--resource-group` is omitted. It now lists all MySQL servers across the subscription using a single ARM API call, matching the behavior of the PostgreSQL toolset.
```

Let's use friendlier terminology 😄

tools/Azure.Mcp.Tools.MySql/src/Commands/BaseMySqlCommand.cs
alzimmermsft, line 1: This class doesn't do anything now, just delete it.

tools/Azure.Mcp.Tools.MySql/src/Services/MySqlService.cs
```diff
@@ -347,24 +360,24 @@ public async Task<List<string>> GetTablesAsync(string subscriptionId, string res
...
         {
             tables.Add(reader.GetString(0));
             tableCount++;
         }
 
-        if (tableCount >= MaxResultLimit)
+        if (tableCount >= MaxRowCount)
         {
-            tables.Add($"... (output limited to {MaxResultLimit:N0} tables for security and performance reasons)");
+            tables.Add($"... (output limited to {MaxRowCount:N0} tables for security and performance reasons)");
```
alzimmermsft: Seems this should be returning a list of string for the tables and truncation information. This could errantly be interpreted as a table name

alzimmermsft (approved):

PR #2678, opened 2026-05-20
Fix EventHubs Tool Issues.

g2vinay:
### What changed

**Correctness**
- `DeleteNamespaceAsync` now returns `false` for 404 instead of `true`. `NamespaceDeleteCommand` reports "not found" rather than "deleted successfully" when the namespace never existed.
- `EventHubUpdateCommand` — `--status` option is now forwarded to the service and applied. Previously it was accepted but silently discarded, leaving the hub unchanged.
- Invalid SKU tier now throws a clear `ArgumentException` instead of silently defaulting to `Standard`.

**Reliability**
- `CreateOrUpdateConsumerGroupAsync` now validates parent resources (resource group → namespace → event hub) before proceeding, throwing `KeyNotFoundException` with an actionable message at each level instead of an unhandled `NullReferenceException`.
- `GetNamespaceAsync` now validates `namespaceName` and `resourceGroup` before making any ARM call.
- All not-found conditions now throw `KeyNotFoundException` → HTTP 404, not `InvalidOperationException` → HTTP 500.

**Performance**
- `GetNamespacesAsync` (no resource group filter) replaced an N+1 per-resource-group loop with a single `GetEventHubsNamespacesAsync` subscription-scoped call.

**Polish**
- SKU tier and Event Hub status inputs are now matched case-insensitively.
- `EventHubGetCommand` error log messages were swapped between the single-get and list-all paths — corrected.
- `EventHubsService` is now `sealed`.


tools/Azure.Mcp.Tools.EventHubs/src/Services/EventHubsService.cs
```diff
@@ -152,12 +153,15 @@ public async Task<Namespace> CreateOrUpdateNamespaceAsync(
...
                 {
-                    "Basic" => EventHubsSkuTier.Basic,
-                    "Standard" => EventHubsSkuTier.Standard,
-                    "Premium" => EventHubsSkuTier.Premium,
-                    _ => EventHubsSkuTier.Standard
+                    "BASIC" => EventHubsSkuTier.Basic,
+                    "STANDARD" => EventHubsSkuTier.Standard,
+                    "PREMIUM" => EventHubsSkuTier.Premium,
+                    null or "" => EventHubsSkuTier.Standard,
+                    _ => throw new ArgumentException(
```
alzimmermsft, line 162: This will need to be listed as a breaking change

tools/Azure.Mcp.Tools.EventHubs/src/Services/EventHubsService.cs
```diff
@@ -243,10 +247,10 @@ public async Task<bool> DeleteNamespaceAsync(
         }
         catch (RequestFailedException ex) when (ex.Status == 404)
         {
-            _logger.LogInformation(
-                "Event Hubs namespace '{NamespaceName}' not found in resource group '{ResourceGroup}'",
+            _logger.LogWarning(
```
alzimmermsft: I don't think this warrants being a warning level log.

alzimmermsft (approved):

PR #2682, opened 2026-05-20
Fix Monitor Tools Issues

g2vinay:
Fix several correctness, reliability, and performance issues in the Monitor toolset:
      - Activity log level filtering now works correctly. The OData filter used the wrong field name ('levels' instead of 'level'), causing the API to silently ignore the filter and return all severity levels regardless of what was requested.
      - Workspace log queries no longer require --resource-group. The option was incorrectly enforced on all Monitor commands, including workspace queries that never use it. It is now only required on the commands that actually need it.
      - Table-type filtering is now case-insensitive. Passing 'customlog' instead of 'CustomLog' previously returned an empty list with no error. Filtering now matches regardless of casing.
      - Query limit detection is now locale-safe. The 'limit' keyword check used a culture-sensitive comparison that could fail in Turkish locale, causing a duplicate limit clause and invalid KQL. This now uses ordinal comparison.
      - QueryWorkspace now logs and classifies errors correctly. Exceptions previously propagated as bare stack traces; they are now caught, logged with context, and 404s are surfaced as a clear 'not found' error.
      - Activity log pagination now stops as soon as the requested number of results is collected. Previously, all pages were fetched into memory before truncation, causing unnecessary ARM round-trips and potential timeouts on busy subscriptions.

### DIFF
--- a/servers/Azure.Mcp.Server/changelog-entries/1779260700545.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1779260700545.yaml
@@ -1,0 +1,10 @@
+changes:
+  - section: "Bug Fixes"
+    description: |
+      Fix several correctness, reliability, and performance issues in the Monitor toolset:
+      - Activity log level filtering now works correctly. The OData filter used the wrong field name ('levels' instead of 'level'), causing the API to silently ignore the filter and return all severity levels regardless of what was requested.
+      - Workspace log queries no longer require --resource-group. The option was incorrectly enforced on all Monitor commands, including workspace queries that never use it. It is now only required on the commands that actually need it.
+      - Table-type filtering is now case-insensitive. Passing 'customlog' instead of 'CustomLog' previously returned an empty list with no error. Filtering now matches regardless of casing.
+      - Query limit detection is now locale-safe. The 'limit' keyword check used a culture-sensitive comparison that could fail in Turkish locale, causing a duplicate limit clause and invalid KQL. This now uses ordinal comparison.
+      - QueryWorkspace now logs and classifies errors correctly. Exceptions previously propagated as bare stack traces; they are now caught, logged with context, and 404s are surfaced as a clear 'not found' error.
+      - Activity log pagination now stops as soon as the requested number of results is collected. Previously, all pages were fetched into memory before truncation, causing unnecessary ARM round-trips and potential timeouts on busy subscriptions.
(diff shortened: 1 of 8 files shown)


servers/Azure.Mcp.Server/changelog-entries/1779260700545.yaml
```diff
@@ -0,0 +1,10 @@
+changes:
+  - section: "Bug Fixes"
+    description: |
+      Fix several correctness, reliability, and performance issues in the Monitor toolset:
+      - Activity log level filtering now works correctly. The OData filter used the wrong field name ('levels' instead of 'level'), causing the API to silently ignore the filter and return all severity levels regardless of what was requested.
+      - Workspace log queries no longer require --resource-group. The option was incorrectly enforced on all Monitor commands, including workspace queries that never use it. It is now only required on the commands that actually need it.
```
alzimmermsft, line 6: Are we sure the change here should be the workspace query command now takes resource group? Looking at its implementation it calls to list workspaces (which the tool there takes resource group) without using resource group.

Unless workspace name or ID is unique across subscriptions, then this is the correct change.

tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorService.cs
```diff
@@ -118,34 +118,46 @@ public async Task<List<JsonNode>> QueryWorkspace(
...
                     }
-                    results.Add(rowDict);
                 }
             }
+            return results;
+        }
+        catch (RequestFailedException ex) when (ex.Status == 404)
+        {
+            throw new KeyNotFoundException($"Workspace '{workspaceId}' not found.", ex);
+        }
```
alzimmermsft, line 155: Are we doing this remapping just to make the error message better? When we get `RequestFailedException` we use the `Status` to determine the `statusCode` JSON property in the response.

tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorService.cs
```diff
@@ -118,34 +118,46 @@ public async Task<List<JsonNode>> QueryWorkspace(
...
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error querying workspace '{WorkspaceId}'. Query: {Query}", workspaceId, query);
+            throw;
```
alzimmermsft, line 159: This isn't needed, the Command class calling this is already doing a try-catch-log, this will just create duplicate logs. Instead improve the log message in `WorkspaceLogQueryCommand`.

Also, we shouldn't log the query as it may contain PII.

PR #2683, opened 2026-05-20
Fix appconfig issues

g2vinay:
Fixed multiple bugs in the App Configuration toolset:
      - Store lookup no longer passes subscription as both subscription-based arguments when resolving the store by name
      - The user-supplied retry policy is now applied to all data-plane key-value operations (get, set, delete, lock); previously it was only applied to the ARM store-discovery call, causing custom retry settings to have no effect on actual configuration reads and writes
      - The `kv delete` command response now includes an `existed` field (`true` if the key was present and deleted, `false` if it was already absent) and a human-readable `message`, allowing callers to distinguish a successful deletion from a no-op on a non-existent key

### DIFF
--- a/servers/Azure.Mcp.Server/changelog-entries/1779262158643.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1779262158643.yaml
@@ -3,6 +3,5 @@ changes:
     description: |
       Fixed multiple bugs in the App Configuration toolset:
       - Store lookup no longer passes subscription as both the store identifier and the error context argument; "store not found" messages now correctly name the missing store instead of echoing the subscription value
-      - Setting a key-value with a malformed tag (missing the required `key=value` separator) now returns a 400 validation error with a clear message instead of silently creating a tag with an empty value
       - The user-supplied retry policy is now applied to all data-plane key-value operations (get, set, delete, lock); previously it was only applied to the ARM store-discovery call, causing custom retry settings to have no effect on actual configuration reads and writes
-      - The `kv delete` command response now includes an `existed` field (`true` if the key was present and deleted, `false` if it was already absent), allowing callers to distinguish a successful deletion from a no-op on a non-existent key
+      - The `kv delete` command response now includes an `existed` field (`true` if the key was present and deleted, `false` if it was already absent) and a human-readable `message`, allowing callers to distinguish a successful deletion from a no-op on a non-existent key
--- a/servers/Azure.Mcp.Server/changelog-entries/1779262158643.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1779262158643.yaml
@@ -2,6 +2,6 @@ changes:
   - section: "Bugs Fixed"
     description: |
       Fixed multiple bugs in the App Configuration toolset:
-      - Store lookup no longer passes subscription as both the store identifier and the error context argument; "store not found" messages now correctly name the missing store instead of echoing the subscription value
+      - Store lookup no longer passes subscription as both subscription-based arguments when resolving the store by name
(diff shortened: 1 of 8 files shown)


tools/Azure.Mcp.Tools.AppConfig/src/Services/AppConfigService.cs
```diff
@@ -209,7 +208,7 @@ private async Task<AppConfigurationAccount> FindAppConfigStore(
             additionalFilter: $"name =~ '{EscapeKqlString(accountName)}'",
             tenant: tenant,
             cancellationToken: cancellationToken)
-            ?? throw new KeyNotFoundException($"App Configuration store '{accountName}' not found for subscription '{subscriptionIdentifier}'.");
+            ?? throw new KeyNotFoundException($"App Configuration store '{accountName}' not found for subscription '{subscription}'.");
```
alzimmermsft:
```suggestion
            ?? throw new KeyNotFoundException($"App Configuration store '{accountName}' not found in subscription '{subscription}'.");
```

tools/Azure.Mcp.Tools.AppConfig/src/Commands/KeyValue/KeyValueDeleteCommand.cs
```diff
@@ -50,7 +50,10 @@ await _appConfigService.DeleteKeyValue(
                 options.Label,
                 cancellationToken);
 
-            context.Response.Results = ResponseResult.Create(new(options.Key, options.Label), AppConfigJsonContext.Default.KeyValueDeleteCommandResult);
+            var message = existed
+                ? $"Key '{options.Key}' deleted successfully."
+                : $"Key '{options.Key}' did not exist in store '{options.Account}'.";
```
alzimmermsft: Bit nuanced here as we should also include the Label as well in the message if Label was passed.

servers/Azure.Mcp.Server/changelog-entries/1779262158643.yaml
```diff
@@ -0,0 +1,7 @@
+changes:
+  - section: "Bugs Fixed"
+    description: |
+      Fixed multiple bugs in the App Configuration toolset:
+      - Store lookup no longer passes subscription as both the store identifier and the error context argument; "store not found" messages now correctly name the missing store instead of echoing the subscription value
```
alzimmermsft: It was never doing that, it was just weirdly passing `subscription` as the parameter for two subscription-based parameters.

tools/Azure.Mcp.Tools.AppConfig/tests/Azure.Mcp.Tools.AppConfig.Tests/Services/AppConfigServiceTests.cs
```diff
@@ -34,6 +34,22 @@ public void FindAppConfigStore_InvokesExecuteSingleResourceQueryAsyncWithNamedAd
...
+    public void FindAppConfigStore_DoesNotHaveRedundantSubscriptionIdentifierParameter()
+    {
+        var method = typeof(AppConfigService).GetMethod(
+            "FindAppConfigStore",
+            BindingFlags.NonPublic | BindingFlags.Instance);
+        Assert.NotNull(method);
+
+        var paramNames = method!.GetParameters().Select(p => p.Name).ToArray();
+        Assert.DoesNotContain("subscriptionIdentifier", paramNames);
+    }
```
alzimmermsft, line 38: Let's not add this test, this doesn't really prevent what happened there

alzimmermsft (approved):

PR #2687, opened 2026-05-20
Fix Compute Tools Issues

g2vinay:
Fixed multiple bugs in Compute VM/VMSS commands:
      - `vm delete` now correctly reports 'not found' when the target VM does not exist instead of always reporting success.
      - `vm update` and `vmss update` now correctly clear all tags when `--tags ''` is passed; fixed comma-separated tag delimiter mismatch between documentation and implementation.
      - `DetermineOsType` now uses word-boundary token matching: no longer misidentifies images with 'win' mid-word (e.g., 'twin-ubuntu') as Windows.

### DIFF
--- a/servers/Azure.Mcp.Server/changelog-entries/1779264486206.yaml
+++ b/servers/Azure.Mcp.Server/changelog-entries/1779264486206.yaml
@@ -1,7 +1,10 @@
 changes:
+  - section: "Breaking Changes"
+    description: |
+      `vm delete` now returns a not-found outcome (`Success: false`) when the target VM does not exist, instead of reporting a successful deletion.
+      This may affect automation that previously treated non-existent VM deletions as success.
   - section: "Bugs Fixed"
     description: |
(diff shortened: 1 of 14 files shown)


servers/Azure.Mcp.Server/changelog-entries/1779264486206.yaml
```diff
@@ -0,0 +1,7 @@
+changes:
+  - section: "Bugs Fixed"
+    description: |
+      Fixed multiple bugs in Compute VM/VMSS commands:
+      - `vm delete` now correctly reports 'not found' when the target VM does not exist instead of always reporting success.
```
alzimmermsft: This one is technically a breaking change

alzimmermsft (approved):

haagha (changes requested):
VMSS delete has the same behaviour - can we change that too?

alzimmermsft (approved):

PR #2725, opened 2026-05-25
Emit AzSubscriptionGuid telemetry tag in namespace tool loader

shrja-ms:
## Summary

In **namespace** server mode (the default for the Azure MCP Server), `McpRuntime.CallToolHandler` only inspects the outer `request.Params.Arguments` for a top-level `subscription` key. But in namespace mode the MCP request envelope is `{intent, command, parameters: {…}}` — the real `--subscription` value lives one level deeper inside `parameters`. As a result, the `AzSubscriptionGuid` Activity tag is **never emitted for namespace-mode tool calls**, which made per-subscription telemetry empty for the entire Azure MCP Server tool surface.

## Fix

`NamespaceToolLoader.InvokeChildToolAsync` now extracts the inner `--subscription` value from the parsed `parameters` dictionary and emits `AzureTagName.SubscriptionGuid`, matching the existing `McpRuntime` behavior for direct tool invocations.

The key match uses the same `NameNormalization` logic that the schema generator uses, so the lookup is case-insensitive and aligned with however the MCP client serialises the option name.

## Tests

Adds 6 unit tests in `NamespaceToolLoaderTests.cs`:
- String subscription value is emitted as the tag
- Key match is case-insensitive
- No tag when `subscription` parameter is absent
- No tag when value is empty
- No-op when `Activity` is null
- No-op when `parameters` is null

All passing locally (`dotnet test --filter "FullyQualifiedName~NamespaceToolLoaderTests.TryEmitSubscriptionTag"` → 6/6).

## Impact

After this lands, all Azure MCP Server namespace-mode tool calls will carry the `AzSubscriptionGuid` Activity tag → restores per-subscription telemetry for downstream reporting (e.g., Internal vs External customer classification via `mabprod1.AzureBackup.GetInternalSubscriptions()`).

## Invoking Livetests

### DIFF
--- a/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
+++ b/core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
@@ -1022,9 +1022,11 @@ public sealed class NamespaceToolLoaderTests : IAsyncDisposable
     }
 
     [Fact]
-    public void TryEmitSubscriptionTag_DoesNothing_WhenSubscriptionIsEmpty()
+    public void TryEmitSubscriptionTag_EmitsEmptyString_MatchingMcpRuntimeBehavior()
     {
         // Arrange
+        // McpRuntime.CallToolHandler emits the AzSubscriptionGuid tag for any non-null string,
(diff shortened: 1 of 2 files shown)


shrja-ms:
Addressed in e4e91d614 — changed the guard from !string.IsNullOrEmpty(subscription) to subscription != null so the tag is emitted for any non-null string (including empty), matching McpRuntime.CallToolHandler. Updated the corresponding unit test to assert the empty-string tag is now emitted. Build + the 6 telemetry unit tests pass locally.

PR #2726, opened 2026-05-25
azurebackup: fix telemetry-triaged bugs NEW-1, NEW-3, NEW-4

shrja-ms:
## Summary

Fixes three telemetry-triaged Azure Backup MCP bugs from the most recent report (NEW-1, NEW-3, NEW-4) in a single PR.

| Bug | Symptom (before) | Behavior (after) |
| --- | --- | --- |
| **NEW-1** | When both RSV and DPP vault listings failed, `azmcp_azurebackup_vault_get` (and `azmcp_azurebackup_governance_find-unprotected`) leaked a raw `AggregateException` to the client, so the actual HTTP status / error code / message from the backends were buried. | A single meaningful exception is thrown. If both backends fail with the same HTTP status (e.g. 401/403), the inner `RequestFailedException` is surfaced directly. Otherwise both inner messages are folded into one `InvalidOperationException`. |
| **NEW-3** | Passing a subscription **name** (display name) in `--subscription` caused `FormatException` from `Azure.Core.ResourceIdentifier.SubscriptionId` for every `azmcp_azurebackup_*` command. This violated the documented MCP contract that `--subscription` accepts both subscription IDs and names. | Every public `AzureBackupService` method (24 sites) resolves the input via `ISubscriptionService.GetSubscription` before constructing any ARM `ResourceIdentifier`. GUID inputs short-circuit through `Guid.TryParse` so the existing happy path makes zero extra calls. |
| **NEW-4** | `azmcp_azurebackup_protectableitem_list` was leaking a 500 `ArgumentException` from the service layer when callers passed an unrecognised `--workload-type`. | A `System.CommandLine` validator on the command rejects the unknown value as a 400 ValidationError with a friendly message listing the supported aliases. The validator's alias set mirrors exactly the existing service-layer guard (`RsvBackupOperations.NormalizeWorkloadTypeForFilter`); the service-layer guard is intentionally left in place as defense-in-depth. |

## Tests

- **`Azure.Mcp.Tools.AzureBackup.UnitTests`: 584 / 584 passing** (was 581 before; +3 net new tests beyond the existing test that was updated to assert the new NEW-1 behavior).
  - NEW-1: `ListVaultsAsync_BothFailWithSameStatus_ThrowsInnerRequestFailedException`, `ListVaultsAsync_BothFailWithDifferentExceptions_ThrowsInvalidOperationWithBothMessages`.
  - NEW-3: `ListVaultsAsync_WhenSubscriptionIsName_ResolvesToGuidBeforeCallingOps`, `ListVaultsAsync_WhenSubscriptionIsGuid_DoesNotCallSubscriptionService` (asserts `ISubscriptionService.DidNotReceive().GetSubscription(...)` on the GUID happy path).
  - NEW-4: 19 positive cases (one per canonical token + alias, including case-insensitivity) and 4 negative cases (unknown, DPP-only, garbage, and an OData-injection-style input); the negative cases also assert `Service.DidNotReceive()`.
  - Existing `AzureBackupServiceTests` migrated to GUID-literal subscriptions so they exercise the `Guid.TryParse` short-circuit without any mock setup.
  - `AzureBackupSetupTests` DI container updated to register `ISubscriptionService` for the new constructor parameter.
- **Live tests: no re-recording required.** The NEW-4 change is a pre-existing-validator pre-check; the NEW-1 change only affects the both-backends-failing path which has no recording; the NEW-3 change short-circuits on GUID input, and every existing recording uses subscription GUIDs.
- Tool build (`Azure.Mcp.Tools.AzureBackup/src`) and server build (`servers/Azure.Mcp.Server`): clean, 0 warnings / 0 errors.

## Files

- `tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs` &mdash; NEW-1 + NEW-3 fixes (new `ResolveSubscriptionIdAsync` helper, new `BuildBothVaultListingsFailedException` helper, ctor now takes `ISubscriptionService`, one-line patches at the top of all 24 public async methods).
- `tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ProtectableItem/ProtectableItemListCommand.cs` &mdash; NEW-4 validator registration.
- `tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs` &mdash; NEW-4 helper (alias set mirrors RSV service-layer guard).
- `tools/Azure.Mcp.Tools.AzureBackup/tests/Azure.Mcp.Tools.AzureBackup.UnitTests/Services/AzureBackupServiceTests.cs` &mdash; new NEW-1 + NEW-3 tests, existing tests migrated to GUID subscriptions.
- `tools/Azure.Mcp.Tools.AzureBackup/tests/Azure.Mcp.Tools.AzureBackup.UnitTests/AzureBackupSetupTests.cs` &mdash; DI container registers `ISubscriptionService`.
- `tools/Azure.Mcp.Tools.AzureBackup/tests/Azure.Mcp.Tools.AzureBackup.UnitTests/ProtectableItem/ProtectableItemListCommandTests.cs` &mdash; NEW-4 tests.
- `servers/Azure.Mcp.Server/changelog-entries/1779703098177.yaml` &mdash; changelog entry (3 "Bugs Fixed" bullets).

## Invoking Livetests


tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
```diff
@@ -42,6 +93,7 @@ public async Task<VaultCreateResult> CreateVaultAsync(
         string location, string? sku, string? storageType, string? tenant,
         RetryPolicyOptions? retryPolicy, CancellationToken cancellationToken)
     {
+        subscription = await ResolveSubscriptionIdAsync(subscription, tenant, retryPolicy, cancellationToken);
         VaultTypeResolver.ValidateVaultType(vaultType);
```
alzimmermsft: nit: Perform validation that can be done without a possible network call first

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
```diff
@@ -17,9 +18,59 @@
...
+        AggregateException rsvFault,
+        AggregateException dppFault,
+        string operationDescription)
+    {
+        var rsvInner = rsvFault.Flatten().InnerExceptions.FirstOrDefault() ?? rsvFault;
+        var dppInner = dppFault.Flatten().InnerExceptions.FirstOrDefault() ?? dppFault;
+
+        if (rsvInner is RequestFailedException rsvReq &&
+            dppInner is RequestFailedException dppReq &&
+            rsvReq.Status == dppReq.Status)
```
alzimmermsft: When the status codes are the same is it safe to ignore one of them? For example, would it be possible for both to throw a 403 but have meaningfully different information on them that should be returned?

tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs
```diff
@@ -0,0 +1,61 @@
...
+internal static class WorkloadTypeNormalizer
+{
+    /// <summary>
+    /// Human-readable supported-token summary, matching the message the service layer
+    /// throws so that error text stays consistent regardless of which side rejects.
+    /// </summary>
+    public const string SupportedTokensDescription =
+        "SQL (or SQLDatabase), SQLInstance, SAPHana (or SAPHanaDatabase), SAPHanaSystem, " +
+        "SAPHanaDBInstance (or SAPHanaDBI), VM (or IaaSVM, VirtualMachine), " +
+        "FileShare (or AzureFileShare, AFS), SAPAse (or SAPAseDatabase, ASE, Sybase)";
```
alzimmermsft: Would it be better to remove the parentheses on aliases?


```suggestion
    public const string SupportedTokensDescription =
        "SQL or SQLDatabase, SQLInstance, SAPHana or SAPHanaDatabase, SAPHanaSystem, " +
        "SAPHanaDBInstance or SAPHanaDBI, VM, IaaSVM, or VirtualMachine, " +
        "FileShare, AzureFileShare, or AFS, or SAPAse, SAPAseDatabase, ASE, or Sybase";
```

tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs
```diff
@@ -0,0 +1,61 @@
...
+            _ => false
+        };
+    }
+
+    /// <summary>
+    /// Returns a human-readable error message naming the rejected value and the
+    /// supported tokens, suitable for a System.CommandLine validator.
+    /// </summary>
+    public static string FormatUnknownMessage(string? workloadType) =>
+        $"Unknown workload type '{workloadType}'. Supported values: {SupportedTokensDescription}.";
```
alzimmermsft:
```suggestion
        $"Unknown workload type '{workloadType}'. Supported values (case-insensitive): {SupportedTokensDescription}.";
```

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ProtectableItem/ProtectableItemListCommand.cs
```diff
@@ -39,6 +39,21 @@ protected override void RegisterOptions(Command command)
...
+            // from the service layer as a 500.
+            if (commandResult.HasOptionResult(AzureBackupOptionDefinitions.WorkloadType.Name))
+            {
+                var value = commandResult.GetValue<string>(AzureBackupOptionDefinitions.WorkloadType.Name);
+                if (!string.IsNullOrWhiteSpace(value) &&
+                    !WorkloadTypeNormalizer.IsSupported(value))
+                {
+                    commandResult.AddError(WorkloadTypeNormalizer.FormatUnknownMessage(value));
+                }
+            }
```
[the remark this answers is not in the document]
shrja-ms: Already addressed in c7098c5 - the validator now reads the value directly without a HasOptionResult gate, so whitespace-only inputs are surfaced and rejected by IsSupported (which treats whitespace as unknown).

tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs
```diff
@@ -0,0 +1,61 @@
...
+/// NEW-4 fix: centralised user-facing validation for the RSV protectable-item / policy
+/// <c>--workload-type</c> token, mirroring exactly the alias set accepted by
+/// <see cref="RsvBackupOperations"/>'s private <c>NormalizeWorkloadTypeForFilter</c>
+/// switch. Used by the command-layer validator so that unknown values are rejected
+/// at the boundary (400 ValidationError) instead of surfacing as a 500
+/// <see cref="System.ArgumentException"/> from the service layer.
+/// </summary>
+/// <remarks>
+/// Kept intentionally in sync (by hand) with the service-layer switch. If you add a new
+/// alias to either side, add it here too. A unit test asserts the two sets agree.
```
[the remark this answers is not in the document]
shrja-ms: Already addressed in c7098c5 - the misleading remarks line was removed.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
shrja-ms: Done in 944e560 - moved the synchronous VaultTypeResolver.ValidateVaultType ahead of ResolveSubscriptionIdAsync in CreateVaultAsync, and applied the same reordering to ListProtectableItemsAsync (the IsDpp(vaultType) check).

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
shrja-ms: Good catch - fixed in 944e560. We no longer drop the DPP exception when both backends fail with the same HTTP status. Both inner messages are always folded into the resulting InvalidOperationException's message, and the RSV exception is preserved as InnerException for diagnostics. The ListVaultsAsync_BothFailWithSameStatus_* test was updated to assert the new behavior.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs
shrja-ms: Done in 944e560 - parentheses removed from the supported-tokens description (applied verbatim from your suggestion).

tools/Azure.Mcp.Tools.AzureBackup/src/Services/WorkloadTypeNormalizer.cs
shrja-ms: Done in 944e560 - added (case-insensitive) to the unknown-workload-type message (applied verbatim from your suggestion).

alzimmermsft (approved):

PR #2773, opened 2026-05-29
Fix runtime options binding failures

g2vinay:
Fixed trimmed/package runtime option binding failures (e.g., Kusto query/sample) by preserving public parameterless constructors required for reflection-based option activation.

### DIFF
--- a/core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.Tests/Options/OptionBinderTests.cs
+++ b/core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.Tests/Options/OptionBinderTests.cs
@@ -3,6 +3,9 @@
 
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Microsoft.Mcp.Core.Commands;
 using Microsoft.Mcp.Core.Options;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


KarishmaGhiya (approved):

vcolin7 (commented):
The changes look good to me. I'm just curious about why the Communications command needed the standalone change. Is it because it doesn't extend from the other classes? Are there other commands that need the same treatment?

g2vinay:
> The changes look good to me. I'm just curious about why the Communications command needed the standalone change. Is it because it doesn't extend from the other classes? Are there other commands that need the same treatment?

It seems to be the only one using standalone annotations, something to consolidate in a follow up PR.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.Tests/Options/OptionBinderTests.cs
```diff
@@ -444,6 +447,39 @@ public void BindOptions_NestedRequired_BindsDeepValues()
...
+
+        Assert.True(annotations.HasFlag(DynamicallyAccessedMemberTypes.PublicProperties));
+        Assert.True(annotations.HasFlag(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor));
+    }
+
+    private static void AssertHasRequiredMemberAnnotations(MemberInfo member)
+    {
+        var attribute = member.GetCustomAttribute<DynamicallyAccessedMembersAttribute>();
+        Assert.NotNull(attribute);
+        Assert.True(attribute!.MemberTypes.HasFlag(DynamicallyAccessedMemberTypes.PublicProperties));
```
vcolin7, line 477: nit; we checked for nullness in the line above, we don't need to do it again here with `attribute!`.

vcolin7 (approved):

PR #2840, opened 2026-06-09
Change `GetResourceApiVersions.ResourceVisitor` in BicepSchema tools to throw `InvalidDataException` for unknown resource types

vcolin7:
`ResourceVisitor.GetResourceApiVersions()` threw a bare `Exception` (and an unhandled `KeyNotFoundException` when the provider existed but the resource type did not), which `AzApiDocsService` normalized via a broad catch-all that also misclassified real server-side faults as `400 BadRequest`.

This PR throw `InvalidDataException()` at the source instead, consistent with `ResourceVisitor.LoadSingleResource`, and drop the catch-all so unexpected exceptions surface as `500`.


alzimmermsft (approved):

liuwuliuyun (approved):
LGTM, Thanks~

PR #2889, opened 2026-06-16
Use csproj version for Fabric MCP VSIX instead of Major.0.X

AmosHersch:
## Description

The VS Code extension (VSIX) version is computed independently of the `.csproj` version in `New-BuildInfo.ps1` using a `Major.0.<next-marketplace-patch>` scheme. As a result, Fabric's `1.1.0` release published the extension as `1.0.1` while all other release targets (npm, NuGet, ADO feed, MCR) correctly showed `1.1.0`.

Fabric MCP Server follows a GA-only, minor-increment versioning strategy and drives its own explicit version numbers, so this change uses the `.csproj` version verbatim for the Fabric VSIX. All other servers keep the existing marketplace-derived `Major.0.X` patch behavior.

### Verified
- Fabric `1.2.0` -> VSIX `1.2.0`
- Fabric `1.1.0` -> VSIX `1.1.0`
- Azure (non-beta) -> VSIX `Major.0.<marketplace+1>` (unchanged)

Note: this does not retroactively change the already-published `1.0.1` VSIX; it applies from the next release (`1.2.0`).

## Invoking Livetests

### DIFF
--- a/eng/scripts/New-BuildInfo.ps1
+++ b/eng/scripts/New-BuildInfo.ps1
@@ -464,6 +464,15 @@ function Get-ServerDetails {
                 $vsixVersion = "$($version.Major).$($version.Minor).$($version.PrereleaseNumber)"
                 $vsixIsPrerelease = $true
             }
+            elseif ($serverName -eq 'Fabric.Mcp.Server') {
+                # Fabric MCP Server follows a GA-only, minor-increment versioning strategy and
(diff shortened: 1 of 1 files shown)


AmosHersch:
Combined into #2879, which now contains both the post-release auto-bump fix (Update-Version.ps1) and this VSIX version fix (New-BuildInfo.ps1). Closing in favor of that PR.

PR #2907, opened 2026-06-17
Fix case-sensitive tenant ID comparison and introduce StringComparisons helper

tmeschter:
## Summary

Fixes [#407](https://github.com/microsoft/mcp-pr/issues/407): the tenant ID comparison in `HttpOnBehalfOfTokenCredentialProvider` used an ordinal, case-sensitive `!=`, so a valid tenant supplied in mixed case (e.g. from config) was rejected even though the JWT `tid` claim (a lowercase GUID) referred to the same tenant.

The fix makes the comparison case-insensitive, then generalizes the pattern by introducing a small `StringComparisons` helper that documents the intended comparison semantics at each call site.

## Changes

- **Fix [#407](https://github.com/microsoft/mcp-pr/issues/407)**: `HttpOnBehalfOfTokenCredentialProvider` now compares the `tid` claim and configured tenant case-insensitively.
- **New `StringComparisons` helper** (`core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs`) with properties that make comparison intent explicit:
  - `TenantId`
  - `SubscriptionId`, `SubscriptionDisplayName`
  - `ResourceGroup`
  - `ResourceName`
- **Migrated existing comparisons** to use the helper (behavior unchanged — all were already `OrdinalIgnoreCase`):
  - Tenant ID: `TenantService`, AzureIsv test guard
  - Subscription: `SubscriptionService`, `SubscriptionListCommand`
  - Resource group: `ResourceGroupService`, `AppLensService`, `AzureBackupService`, `ResourceResolverService`, `FoundryExtensionsService`
  - Resource name: `Monitor`, `Cosmos`, `EventGrid`, `Search`, `Compute`, `FileShares`, `FoundryExtensions`

## Validation

- `dotnet build` succeeds for all affected projects (0 warnings, 0 errors).

## Invoking Livetests

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
+++ b/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
@@ -35,4 +35,10 @@ public static class StringComparisons
     /// Azure Resource Manager treats case-insensitively.
     /// </summary>
     public static StringComparison ResourceGroup => StringComparison.OrdinalIgnoreCase;
+
+    /// <summary>
+    /// The comparison to use when matching an Azure resource by its name. Azure
+    /// resource names are matched case-insensitively for lookup purposes.
+    /// </summary>
+    public static StringComparison ResourceName => StringComparison.OrdinalIgnoreCase;
 }
--- a/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
+++ b/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
@@ -29,4 +29,10 @@ public static class StringComparisons
     /// which are matched case-insensitively.
     /// </summary>
     public static StringComparison SubscriptionDisplayName => StringComparison.OrdinalIgnoreCase;
+
+    /// <summary>
+    /// The comparison to use when comparing Azure resource group names, which
+    /// Azure Resource Manager treats case-insensitively.
+    /// </summary>
+    public static StringComparison ResourceGroup => StringComparison.OrdinalIgnoreCase;
 }
--- a/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
+++ b/core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Mcp.Core.Helpers;
+
+/// <summary>
+/// Centralizes the <see cref="StringComparison"/> values to use when comparing
+/// well-known string values, making the intended comparison semantics explicit
+/// at each call site.
+/// </summary>
+public static class StringComparisons
+{
+    /// <summary>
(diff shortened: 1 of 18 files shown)


core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Mcp.Core.Helpers;
+
+/// <summary>
+/// Centralizes the <see cref="StringComparison"/> values to use when comparing
+/// well-known string values, making the intended comparison semantics explicit
+/// at each call site.
+/// </summary>
+public static class StringComparisons
```
tmeschter, line 11: The point of this type is to make it clear at the call site (`string.Equals(...)`, etc.) that the expected comparison is being used. Also, if we're using the _wrong_ comparison we can change it here to update it for all consumers.

I didn't move all usages of `StringComparison.Ordinal` or `StringComparison.OrdinalIgnoreCase` to this file in order to keep the size of the change manageable. More can be added in a future update.

core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// The comparison to use when comparing Azure resource group names, which
+    /// Azure Resource Manager treats case-insensitively.
+    /// </summary>
+    public static StringComparison ResourceGroup => StringComparison.OrdinalIgnoreCase;
+
+    /// <summary>
+    /// The comparison to use when matching an Azure resource by its name. Azure
+    /// resource names are matched case-insensitively for lookup purposes.
+    /// </summary>
+    public static StringComparison ResourceName => StringComparison.OrdinalIgnoreCase;
```
alzimmermsft, line 43: Are Azure resources always a case-insensitive comparison? This is the only one that gives me concern as it's a broad category.

alzimmermsft (approved):

core/Microsoft.Mcp.Core/src/Helpers/StringComparisons.cs
JasonYeMSFT: We may be doing case insensitive lookup/queries but I don't think we can assume two names differ only in case refer to the same resource.

https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

tmeschter: @JasonYeMSFT Per that page, resource names are case-insensitive with the exception of [Microsoft.OperationsManagement.solutions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftoperationsmanagement).

tmeschter: And we don't currently do anything with that resource type, as best I can tell.

PR #2920, opened 2026-06-22
Fix unbounded enumeration in `postgres list`

vcolin7:
## What does this PR do?

`postgres list databases` and `postgres list tables` previously returned every row from `pg_database` / `information_schema.tables` with no upper bound, so a server with hundreds of databases or thousands of tables would stream everything into memory with no cap and no truncation signal to the caller. This PR closes that gap by mirroring the existing `mysql list` row-cap pattern.

**Approach (parity with MySQL):**
- Added a `MaxRowCount = 10_000` constant in `PostgresService`.
- `ListDatabasesAsync`: query now uses `ORDER BY datname LIMIT @maxResults`. When the result hits the cap, a human-readable sentinel row is appended to the list (identical wording and behavior to `MySqlService.ListDatabasesAsync`).
- `ListTablesAsync`: now returns a new `TableListResult(List<string> Tables, bool IsTruncated)` sealed record (mirrors `MySql/.../TableListResult.cs`). Query uses `ORDER BY table_name LIMIT @maxResults + 1`; truncation is detected via `tables.Count > MaxRowCount`, then the extra row is trimmed.
- `PostgresListCommand` surfaces the flag via `bool? TablesTruncated` on `PostgresListCommandResult` (4th positional param, default `null`), matching `MySqlListCommandResult` exactly. `bool?` + `JsonIgnoreCondition.WhenWritingNull` keeps the field absent from non-truncated responses for backward compatibility.

**Notes:**
- The two methods deliberately use different SQL caps (`LIMIT cap` vs `LIMIT cap+1`) because they use different truncation signals (sentinel row vs structured flag). Inline comments on both methods explain the differences.
- I considered using MySQL's read-then-probe pattern for tables (no SQL `LIMIT`, post-loop `Read` to detect overflow) but it silently fails to detect truncation once a SQL `LIMIT` is added: the loop's exit iteration consumes the probe row. Switched to `LIMIT cap+1` + `Count > cap`, which keeps the public surface identical to MySQL.
- No `--max-results` CLI option was added. MySQL does not expose one, so Postgres does not either (parity, simpler, backward compatible).

## GitHub issue number?

Fixes [#472](https://github.com/microsoft/mcp-pr/issues/472)

## Pre-merge Checklist
- [x] Required for All PRs
    - [x] Commit history is clean with descriptive messages
    - [x] Created a changelog entry
- [x] For MCP tool changes:
    - [x] **One tool per PR**: only `azmcp postgres list` is modified
    - [ ] Updated `servers/Azure.Mcp.Server/README.md` and/or `servers/Fabric.Mcp.Server/README.md` documentation - N/A (tool description unchanged)
    - [ ] Validate `README.md` changes running the script `./eng/scripts/Process-PackageReadMe.ps1` - N/A
    - [ ] For new or modified tool descriptions, ran `ToolDescriptionEvaluator` - N/A (no tool description changes)
    - [ ] For tools with new names, including new tools or renamed tools, update `consolidated-tools.json` - N/A (no new/renamed tools)
    - [ ] For **renamed** tools, follow the [Tool Rename Checklist](https://github.com/microsoft/mcp/blob/main/docs/tool-rename-checklist.md) - N/A
    - [ ] For new tools associated with Azure services or publicly available tools/APIs/products, add URL to documentation - N/A
- [x] Extra steps for **Azure MCP Server** tool changes:
    - [x] Updated command list in `servers/Azure.Mcp.Server/docs/azmcp-commands.md`
    - [ ] Ran `./eng/scripts/Update-AzCommandsMetadata.ps1` to update tool metadata - N/A (no command surface or metadata change; output shape is backward compatible)
    - [ ] Updated test prompts in `servers/Azure.Mcp.Server/docs/e2eTestPrompts.md` - existing prompts still apply
    - [ ] 👉 For Community (non-Microsoft team member) PRs - N/A

## Invoking Livetests


servers/Azure.Mcp.Server/changelog-entries/1782105637872.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Bugs Fixed"
+    description: "Cap 'azmcp postgres list' database and table results at 10,000 entries to prevent unbounded enumeration on large servers, mirroring the existing 'azmcp mysql list' behavior. When the database list is truncated, a sentinel row is appended; when the table list is truncated, the response includes a 'tablesTruncated: true' flag."
```
alzimmermsft: I'd use a truncation `bool` on both as a sentinel row value doesn't have much meaning outside of internal usages where we explicitly know what to expect.

tools/Azure.Mcp.Tools.Postgres/src/Services/PostgresService.cs
```diff
@@ -91,15 +93,27 @@ public async Task<List<string>> ListDatabasesAsync(
...
         var dbs = new List<string>();
-        while (await reader.ReadAsync(cancellationToken))
+        var dbCount = 0;
+        while (await reader.ReadAsync(cancellationToken) && dbCount < MaxRowCount)
         {
             dbs.Add(reader.GetString(0));
+            dbCount++;
+        }
+
+        if (dbCount >= MaxRowCount)
```
alzimmermsft: In the current design we never trigger this case as we're having the query statement limit to 10,000 then check if we went over 10,000. Only time this would happen is if there is an error in running the query (which is a much bigger issue).

Make the query limit 10,001 for this to be properly handled.

tools/Azure.Mcp.Tools.Postgres/src/Services/PostgresService.cs
```diff
@@ -175,16 +189,27 @@ public async Task<List<string>> ListTablesAsync(
...
         var connectionString = BuildConnectionString(host, database, user, passwordToUse);
 
-        var query = "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public';";
+        var query = "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' ORDER BY table_name LIMIT @maxResults;";
         await using IPostgresResource resource = await _dbProvider.GetPostgresResource(connectionString, authType, cancellationToken);
         await using NpgsqlCommand command = _dbProvider.GetCommand(query, resource);
+        // Fetch cap+1 rows so we can detect truncation by observing whether an extra row exists, then trim it.
+        // Unlike ListDatabasesAsync (which signals truncation via a sentinel row appended to the list),
+        // tables return a structured TableListResult with an IsTruncated flag, so we need the extra row to set it.
+        command.Parameters.AddWithValue("maxResults", MaxRowCount + 1);
```
alzimmermsft, line 189: This case correctly handles the capping check with using 10,001

tools/Azure.Mcp.Tools.Postgres/tests/Azure.Mcp.Tools.Postgres.Tests/Services/PostgresServiceRowLimitTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using Azure.Mcp.Tools.Postgres.Tests.Services.Support;
+using Npgsql;
+using NSubstitute;
+using Xunit;
+
+namespace Azure.Mcp.Tools.Postgres.Tests.Services;
+
+public class PostgresServiceRowLimitTests
+{
+    private const int MaxRowCount = 10_000;
```
alzimmermsft: Make the const in source code `internal` so we don't duplicate it here, possibly having drift over time.

alzimmermsft (approved):

Issue #89, opened 2025-08-29
Standardize the cspell usage


weshaggard:
Update cspell to use a distributed model like we do in https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/cspell.yaml where each tool has its own unique words and imports a common file. The common file should only add words that applicable to the entire repo.

hallipr:
In the rest-api-specs repo, there doesn't seem to be any automation to run spelling checks.   Our code in eng/common looks for a root spelling config at .vscode/cspell.json, but this may still support additional configs per folder.  I don't know if each folder's config should point to the root config and how our eng/common cspell script would know to look at the per-folder configs,

If cspell already supports directory based hierarchical configs, maybe we can just declare a root spelling config without having to import it in each subfolder's more specific config.

hallipr:
see:
[eng/common/scripts/check-spelling-in-changed-files.ps1](https://github.com/Azure/azure-rest-api-specs/blob/a39b73b1c7d12a633805a3b2ed3177a8bfddd9e2/eng/common/scripts/check-spelling-in-changed-files.ps1#L59)

[eng/common/spelling/Invoke-Cspell.ps1](https://github.com/Azure/azure-rest-api-specs/blob/a39b73b1c7d12a633805a3b2ed3177a8bfddd9e2/eng/common/spelling/Invoke-Cspell.ps1)

danieljurek:
> In the rest-api-specs repo, there doesn't seem to be any automation to run spelling checks.

This is done in GitHub Actions 
* https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/spelling.yaml
* https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/spelling-all.yaml

> Our code in eng/common looks for a root spelling config at .vscode/cspell.json

Use `CspellConfigPath` to set the correct path. You don't need to nest in .vscode/cspell.json, that was only done to keep cspell JSON and YAML files from cluttering up the file list at the root of the repos (we already had a `.vscode` folder). 

> I don't know if each folder's config should point to the root config and how our eng/common cspell script would know to look at the per-folder configs,

This is handled by cspell itself. When it sees a file, it looks up the directory hierarchy until it finds a config file like `cspell.yaml`. Then it uses JUST the configs it finds in that file and no others (so you _must_ use an `import` to reference desired additional configs).  

[...]

PR #2958, opened 2026-06-26
Switch CSpell to distributed form

chidozieononiwu:
This pull request adds a spell checking workflow to the repository and introduces per-project spell check configuration files to improve spelling validation and maintainability across the codebase. The main changes are the addition of a GitHub Actions workflow for spell checking on pull requests, a compatibility configuration file, and new or updated `cspell.yaml` files for individual servers and tools.

**CI/CD and Spell Checking Infrastructure:**

* Added a new GitHub Actions workflow (`.github/workflows/spelling.yaml`) that runs a spell check on pull requests targeting the `main` branch, ensuring spelling errors are caught automatically in changed files.
* Added a `cspell.json` compatibility file at the root to support older spell check versions and specify files to lint, importing the main `cspell.yaml`.

**Per-Project Spell Check Configuration:**

* Introduced or updated `cspell.yaml` files for various servers and tools (e.g., `Azure.Mcp.Server`, `Fabric.Mcp.Server`, `Template.Mcp.Server`, and multiple under `tools/Azure.Mcp.Tools.*`). These files inherit from the root config and allow for project-specific spelling overrides, improving accuracy and maintainability of spell checking. [[1]](diffhunk://#diff-6b91269b2196d498d92faf84bb38278c507519b5444e79f0ffe724e646716a6aR1-R29) [[2]](diffhunk://#diff-a4cf49efc2d1c39c4b4091ff1edee6b56e2c7535e4f907b16f3ea26783caaf14R1-R10) [[3]](diffhunk://#diff-40e63dd8f976d6bf2e7e8b0664228b818a5a87c9e03a319f80874964b300116aR1-R9) [[4]](diffhunk://#diff-242b19a60c7340a68d2caeae6b97f3f1ee593990e897a950b2abcf7ac155db4cR1-R13) [[5]](diffhunk://#diff-af06e4f8da81bea1cb4655f49303845a28135f646bc99aa9f45359e4de49e2aeR1-R9) [[6]](diffhunk://#diff-bec2cf3fc41b9e04a36134ff80a6dc8227576e546654067e99198ffade009687R1-R18) [[7]](diffhunk://#diff-06fe88ed0fee21809a384eb5c5bf38e7b603f140de4b5b57fdc3f9b66c52092dR1-R10) [[8]](diffhunk://#diff-55f6e846d4a9afdb3258aea5f632745cb7842cbf668300142452c5d215f97e6aR1-R10) [[9]](diffhunk://#diff-e3ee54c056400c5d94d93951a9a065ec56c450c73d1a9e430a256c28280c2efbR1-R31) [[10]](diffhunk://#diff-06fe39acc048b43a1bc671b2314d8683945b674e7f703ca4eed23a2d8a2f28a0R1-R11) [[11]](diffhunk://#diff-5fc5e239665e3d7a3eda9225d7e05059d3b46294f5de0d880d8bfde637f699aaR1-R11) [[12]](diffhunk://#diff-77744e9db137b4f92cef49b531b77c496d96260e291593f63c7c9a7b0ba8f914R1-R12) [[13]](diffhunk://#diff-2167732997aed82cbce2160a2970c738db17839c53cdf910323d4a6767ed3c8eR1-R13)

**Project-specific Word Lists:**

* Populated the `words` section in each `cspell.yaml` with terms relevant to that specific project or tool, while leaving the section empty for projects without unique terms. This helps prevent false positives from domain-specific terminology. [[1]](diffhunk://#diff-6b91269b2196d498d92faf84bb38278c507519b5444e79f0ffe724e646716a6aR1-R29) [[2]](diffhunk://#diff-a4cf49efc2d1c39c4b4091ff1edee6b56e2c7535e4f907b16f3ea26783caaf14R1-R10) [[3]](diffhunk://#diff-40e63dd8f976d6bf2e7e8b0664228b818a5a87c9e03a319f80874964b300116aR1-R9) [[4]](diffhunk://#diff-242b19a60c7340a68d2caeae6b97f3f1ee593990e897a950b2abcf7ac155db4cR1-R13) [[5]](diffhunk://#diff-af06e4f8da81bea1cb4655f49303845a28135f646bc99aa9f45359e4de49e2aeR1-R9) [[6]](diffhunk://#diff-bec2cf3fc41b9e04a36134ff80a6dc8227576e546654067e99198ffade009687R1-R18) [[7]](diffhunk://#diff-06fe88ed0fee21809a384eb5c5bf38e7b603f140de4b5b57fdc3f9b66c52092dR1-R10) [[8]](diffhunk://#diff-55f6e846d4a9afdb3258aea5f632745cb7842cbf668300142452c5d215f97e6aR1-R10) [[9]](diffhunk://#diff-e3ee54c056400c5d94d93951a9a065ec56c450c73d1a9e430a256c28280c2efbR1-R31) [[10]](diffhunk://#diff-06fe39acc048b43a1bc671b2314d8683945b674e7f703ca4eed23a2d8a2f28a0R1-R11) [[11]](diffhunk://#diff-5fc5e239665e3d7a3eda9225d7e05059d3b46294f5de0d880d8bfde637f699aaR1-R11) [[12]](diffhunk://#diff-77744e9db137b4f92cef49b531b77c496d96260e291593f63c7c9a7b0ba8f914R1-R12) [[13]](diffhunk://#diff-2167732997aed82cbce2160a2970c738db17839c53cdf910323d4a6767ed3c8eR1-R13)

These changes together establish a robust, maintainable spell checking process that is both automated and customizable per project.

Addresses https://github.com/microsoft/mcp/issues/89


tools/Azure.Mcp.Tools.Advisor/cspell.yaml
```diff
@@ -0,0 +1,9 @@
+# Spell-check overrides for this toolset/server.
+# Inherits the global config at the repo root.
+#
+# Add words that only appear in this folder here, in alphabetical order.
+# Cross-cutting terms should go in the root cspell.yaml instead.
+version: "0.2"
+import:
+  - ../../cspell.yaml
+words: []
```
alzimmermsft, line 9: We should be consistent on how we handle cases when there are no additional words needed, we have both

```yaml
words:
```

and

```yaml
words: []
```

I prefer the first choice given this is a YAML style array.

wbreza (commented):
Solid migration and the word list carries over intact, but a dropped ignore path and stale contributor docs need fixing first.

### What stood out
- **Migration is functionally sound** — every word from the old config is accounted for, per-project configs do merge under an explicit `--config`, and the `HEAD^...HEAD` diff logic is correct.
- **Docs still name the deleted file** — five contributor and agent-facing surfaces point at `.vscode/cspell.json`, which this PR removes and the new config also ignores.
- **One ignore path lost to a rename** — the Fabric spec exclusion did not survive `PublicApi` becoming `Docs`, leaving 1,194 files unguarded.

### Top issues
- [F-001] Contributor and agent docs still reference the deleted config — `cspell.yaml:23`
- [F-002] Dropped ignore exposes 1,194 Fabric spec files to spell check — `cspell.yaml:32`
- [F-003] Three active toolsets did not receive a cspell.yaml — `tools/Azure.Mcp.Tools.KeyVault/cspell.yaml:1`

### Cross-cutting
- [F-001] **Spelling guidance is stale in five files** _(docs)_ — AGENTS.md, CONTRIBUTING.md, the rename checklist, the spelling README and a toolset skill all name the removed path.
- [F-004] **Two spell-check enforcement points coexist** _(process)_ — analyze.yml still includes the pipeline template, with no stated replace-or-supplement intent.

### Improvement opportunities
- [O-001] Enable npm caching for the cspell install _(Recommended)_ — `.github/workflows/spelling.yaml:33`
- [O-002] Cancel superseded spell-check runs per PR _(Recommended)_ — `.github/workflows/spelling.yaml:3`
- [O-003] Bound the job with timeout-minutes _(Optional)_ — `.github/workflows/spelling.yaml:22`
- [O-004] Drop per-project words already covered by the root baseline _(Recommended)_ — `servers/Azure.Mcp.Server/cspell.yaml:11`

cspell.yaml
```diff
@@ -0,0 +1,548 @@
...
+
+ignorePaths:
+  - "**/*.binlog"
+  - "**/.dist/**"
+  - "**/.editorconfig"
+  - "**/.gitattributes"
+  - "**/.git/"
+  - "**/.github/CODEOWNERS"
+  - "**/.gitignore"
+  - "**/.vscode/**"
```
wbreza, line 23: F-001 **Contributor and agent docs still point at the deleted `.vscode/cspell.json`.**

This PR removes `.vscode/cspell.json`, but five instruction surfaces still tell contributors and agents to add terms there:

- `AGENTS.md:730` — "Add new technical terms to .vscode/cspell.json if needed"
- `CONTRIBUTING.md:655`
- `docs/tool-rename-checklist.md:75`
- `eng/common/spelling/README.md:7, 20, 63, 71`
- `tools/Azure.Mcp.Tools.AzureBackup/skills/azurebackup-add-tool/SKILL.md:263`

The resulting failure is silent rather than loud. This line adds `**/.vscode/**` to `ignorePaths`, so a contributor who follows the stale instructions and recreates `.vscode/cspell.json` sees no error — the words are simply never applied, and CI keeps failing with no indication why. `AGENTS.md` is also the instruction contract this repo's AI agents follow, so the drift propagates into automated changes.

None of these five files appear in this diff, and the PR description does not mention the deletion at all.

Fix: update each reference to the new model — project-specific terms go in that folder's `cspell.yaml`, cross-cutting terms in the root `cspell.yaml` — which matches the header comment already present in every new per-project file. The `eng/common/spelling/README.md` examples also need rewriting from JSON to YAML.

cspell.yaml
```diff
@@ -0,0 +1,548 @@
...
+  - "**/.work/**"
+  - "**/NOTICE.txt"
+  - "**/TestResults/**"
+  - "**/bin/**"
+  - "**/eng/common/**"
+  - "**/obj/**"
+  - "**/tools/**/tests/test-resources-post.ps1"
+  - "cspell.yaml"
+  - "cspell.json"
```
wbreza, line 32: F-002 **The Fabric spec ignore was lost in a rename, exposing 1,194 files.**

The deleted config ignored `**/Fabric.Mcp.Tools.PublicApi/src/Resources/**`. That toolset was renamed to `Fabric.Mcp.Tools.Docs` in commit `4504034` (recorded as an `R100` rename), so the old pattern matches nothing — and no equivalent entry was carried into this `ignorePaths` block or into `tools/Fabric.Mcp.Tools.Docs/cspell.yaml`.

`tools/Fabric.Mcp.Tools.Docs/src/Resources/` currently holds 1,194 tracked files, mostly Fabric REST API spec `.json` and `.md`. They are tracked rather than ignored, so `useGitignore: true` on line 3 does not cover them, and no other pattern in this block matches. Running cspell over that directory yields 1,884 unknown-word hits (`Kqldatabase`, `kqldbsource`, `apacheAirflowJob`, and similar API nouns).

Because CI only checks changed files, this does not break every PR — but any PR that touches a Fabric spec file fails spell check outright.

The `**/Fabric.Mcp.Tools.OneLake/src/Resources/**` entry was dropped too; that directory no longer exists, so removing it is harmless.

Fix — restore the exclusion under the current name, either here:

```yaml
  - "**/Fabric.Mcp.Tools.Docs/src/Resources/**"
```

or, matching the distributed pattern this PR introduces, in `tools/Fabric.Mcp.Tools.Docs/cspell.yaml`:

```yaml
ignorePaths:
  - src/Resources/**
```

tools/Azure.Mcp.Tools.KeyVault/cspell.yaml
```diff
@@ -0,0 +1,10 @@
+# Spell-check overrides for this toolset/server.
```
wbreza, line 1: F-003 **Three active toolsets were skipped in the rollout.**

63 per-project configs were added, but these three did not receive one:

- `tools/Azure.Mcp.Tools.Insights`
- `tools/Azure.Mcp.Tools.IoTHub`
- `tools/Azure.Mcp.Tools.ResilienceManagement`

All three are active toolsets — each has `src/` and `tests/` trees and a project file (`tools/Azure.Mcp.Tools.IoTHub/src/Azure.Mcp.Tools.IoTHub.csproj`, and the equivalents for the other two). `Insights` and `IoTHub` would both sort immediately ahead of this file.

"No unique words yet" was not the skip criterion: 31 of the 63 configs that were added ship an empty `words:` list, so these three look like an accidental omission rather than a deliberate exclusion.

Consequence: new domain terms for those toolsets have no local home, so they will accumulate in the root config and re-centralize the word list this PR is decentralizing.

Fix — add the same file at each of the three toolset roots:

```yaml
# Spell-check overrides for this toolset/server.
# Inherits the global config at the repo root.
#
# Add words that only appear in this folder here, in alphabetical order.
# Cross-cutting terms should go in the root cspell.yaml instead.
version: "0.2"
import:
  - ../../cspell.yaml
words:
```

.github/workflows/spelling.yaml
```diff
@@ -0,0 +1,47 @@
...
+          fetch-depth: 2
+
+      - name: Setup Node 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+
+      - name: Run spelling check
+        run: |
+          ./eng/common/scripts/check-spelling-in-changed-files.ps1 `
```
wbreza, line 43: F-004 **The Azure DevOps spell check is left wired alongside this new workflow.**

`eng/pipelines/templates/jobs/analyze.yml:35` still includes `/eng/common/pipelines/templates/steps/check-spelling.yml` with no parameters, so it falls back to that template's default config path (`eng/common/pipelines/templates/steps/check-spelling.yml:23`). After this PR there are two spell-check enforcement points, and the description does not state whether this workflow replaces the pipeline step or supplements it during a transition.

That ambiguity determines how the config paths converge. The template's default still names the removed `.vscode/cspell.json`; if it is retargeted at `./cspell.json` while this workflow keeps `./cspell.yaml`, the two surfaces read different files and can drift apart on which words are errors, since only the shim imports the YAML today.

Fix — pick one and record it in the PR description: either drop the `check-spelling.yml` include from `analyze.yml` because this workflow supersedes it, or pass an explicit `CspellConfigPath` to the template that resolves to the same config this workflow uses.

.github/workflows/spelling.yaml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 2
+
+      - name: Setup Node 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
```
wbreza, line 37: O-001 **Enable npm caching for the cspell install.** _(Recommended)_

`eng/common/spelling/Invoke-Cspell.ps1:92` runs `npm --prefix $PackageInstallCache ci` on every invocation and then deletes the install directory (`:107-109`), so each run re-downloads the full cspell 10.x dependency tree. `actions/setup-node` can prime the runner's npm cache from the pinned lockfile at `eng/common/spelling/package-lock.json`.

```yaml
      - name: Setup Node 24
        uses: actions/setup-node@v6
        with:
          node-version: 24
          cache: npm
          cache-dependency-path: eng/common/spelling/package-lock.json
```

.github/workflows/spelling.yaml
```diff
@@ -0,0 +1,47 @@
+name: SpellCheck
+
+on:
```
wbreza, line 3: O-002 **Cancel superseded runs for the same PR.** _(Recommended)_

With `opened`, `synchronize`, `reopened` and `edited` all triggering, rapid pushes stack runs that each pay a fresh dependency install. `.github/workflows/onboarding-welcome.yml:17-19` already uses this pattern in-repo.

```yaml
concurrency:
  group: spellcheck-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

Worth noting that peer PR workflows such as `verify-links.yml` also omit a concurrency block, so treat this as a deliberate improvement rather than a consistency correction.

.github/workflows/spelling.yaml
```diff
@@ -0,0 +1,47 @@
...
+      # generate incorrect diff
+      - edited
+
+permissions:
+  contents: read
+
+jobs:
+  spellcheck:
+    name: SpellCheck
+    runs-on: ubuntu-24.04
```
wbreza, line 22: O-003 **Bound the job with `timeout-minutes`.** _(Optional)_

The job inherits the 360-minute default while performing a network install (`eng/common/spelling/Invoke-Cspell.ps1:92`). A short ceiling converts a hung registry call into a fast, legible failure instead of a long-running runner.

```yaml
  spellcheck:
    name: SpellCheck
    runs-on: ubuntu-24.04
    timeout-minutes: 15
```

servers/Azure.Mcp.Server/cspell.yaml
```diff
@@ -0,0 +1,29 @@
...
+# Inherits the global config at the repo root.
+#
+# Add words that only appear in this folder here, in alphabetical order.
+# Cross-cutting terms should go in the root cspell.yaml instead.
+version: "0.2"
+import:
+  - ../../cspell.yaml
+words:
+  - agentic
+  - amazonq
```
wbreza: O-004 **Some per-project words are already covered by the root baseline.** _(Recommended)_

The header on this file states that cross-cutting terms belong in the root config, but a few entries duplicate words already present in the root `baseline` dictionary — `amazonq` here, plus `azurecr` in `tools/Azure.Mcp.Tools.Acr/cspell.yaml` and `agentpool` / `azurepolicy` in `tools/Azure.Mcp.Tools.Aks/cspell.yaml`.

cspell accepts either location, so nothing is broken. Removing the duplicates now keeps the new convention unambiguous while it is still being established across 64 files.

Fix: delete the duplicated entries from the per-project files and let the root baseline cover them.

tools/Azure.Mcp.Tools.Advisor/cspell.yaml
chidozieononiwu: first choice is flagged by linter

wbreza: Checked this against cspell 10.0.1 (matching the `^10.0.0` pin in `eng/common/spelling/package.json`) — the linter concern is well founded, and the two forms are not equivalent.

In YAML a bare `words:` with no value parses to `null`, not to an empty sequence. The schema that ships with cspell types the field as a plain array with no null permitted (`node_modules/@cspell/cspell-types/cspell.schema.json`):

```json
"words": {
  "description": "List of words to be considered correct.",
  "items": { "type": "string" },
  "type": "array"
}
```

So `words: []` satisfies the schema and bare `words:` does not — which is what a schema-aware YAML validator reports.

At runtime the CLI tolerates both. I linted the same file through a root config plus a child config using `import:`, once with each form, and the results were identical — the imported words stayed in effect, so a null `words` does not break the import chain:

```
# child with bare `words:`      -> Issues found: 1 (only the intended bogus token)
# child with `words: []`        -> Issues found: 1 (identical)
```

That makes this an editor and schema-validation issue rather than a spell-check correctness one: no CI failure either way, but only `words: []` validates cleanly.

Standardizing on `words: []` satisfies both the consistency request and the schema. If the bare form is preferred for readability, the equivalent that still validates is to omit the `words` key entirely when a project has no local terms — the `import:` of the root config is what the file is really there for.

wbreza (commented):
Re-checked at the same commit — no code changes since the last pass, so the four open findings still stand. Settled the open question about the empty word-list form.

### What stood out
- **`words:` vs `words: []` is resolved** — the bare form parses as null and violates the shipped schema, which types the field as an array; the explicit empty array validates cleanly and neither form changes lint results. Detail is in that thread.
- **All four findings re-verified against current main** — the seven stale doc references, the 1,194-file Fabric spec directory, the three toolsets without a config, and the pipeline include are all still present as described.

### Top issues
- [F-001] Contributor and agent docs still reference the deleted config — `cspell.yaml:23`
- [F-002] Dropped ignore exposes 1,194 Fabric spec files to spell check — `cspell.yaml:32`
- [F-003] Three active toolsets did not receive a cspell.yaml — `tools/Azure.Mcp.Tools.KeyVault/cspell.yaml:1`

### Cross-cutting
- [F-004] **Two spell-check enforcement points coexist** _(process)_ — `eng/pipelines/templates/jobs/analyze.yml:35` still includes the pipeline template, with no stated replace-or-supplement intent.

cspell.yaml
```diff
@@ -0,0 +1,548 @@
...
+      - "roboto"
+      - "rollups"
+      - "scalesetpriority"
+      - "scriptable"
+      - "securestring"
+      - "segoe"
+      - "serverjson"
+      - "setdevversion"
+      - "setvariable"
+      - "skus"
```
wbreza, line 229: F-005 **Seven words from the old config landed in no new file, and the spell-check job is failing on this commit.**

[The run for `e0ea427f`](https://github.com/microsoft/mcp/actions/runs/31220693510/job/93004399063) reports 20 errors in 3 files:

```
AGENTS.md:97:26         - Unknown word (slnx)
CONTRIBUTING.md:103:112 - Unknown word (configfile)
CONTRIBUTING.md:278:107 - Unknown word (remotemcp)
CONTRIBUTING.md:775:22  - Unknown word (uniqueprefix)
SKILL.md:233:29         - Unknown word (slnx)
```

I compared the old `words` array (417 entries) against every word in all 67 `cspell.yaml` files on this branch. Seven are present in the old config and in none of the new ones:

`aiml`, `cloudhealth`, `feff`, `remotemcp`, `slnx`, `uniqueprefix`, and the literal `test_test_pmc2pc1.vmsr_uat_beta`

Only some surface today, because the job checks changed files: `slnx`, `remotemcp`, `configfile` and `uniqueprefix` sit in files this PR edits. The rest are latent and fail whenever their files are next touched — all seven are still in use (`aiml` in 5 files, `cloudhealth` in 9 including `Directory.Packages.props`, `feff` in 4, `slnx` in 19).

`configfile` is a fifth case: it was in the old config, and while it now appears in a per-project file, `CONTRIBUTING.md` sits at the repo root where only the root config applies.

These belong in the root `baseline` dictionary rather than a per-project file, since the files needing them (`AGENTS.md`, `CONTRIBUTING.md`, `.gitattributes`, `Directory.Packages.props`) are repo-root or cross-cutting. Added alphabetically, `slnx` sorts right after `"skus"` on this line:

```yaml
      - "skus"
      - "slnx"
```

and the rest at their alphabetical positions in the same block.

Two remaining errors are unrelated to the migration and want a different fix: `dislay` at `CONTRIBUTING.md:789:22` is a genuine typo for "display", and `worktree`/`worktrees` (10 hits in `AGENTS.md`) is new vocabulary this PR's own doc edits introduced — that one is a real addition to the baseline, not a restoration. `myop` at `SKILL.md:148:36` is a third such case.

wbreza (commented):
The four previous findings are all fixed — docs migrated, the Fabric ignore restored, the three missing configs added, and the DevOps step removed. One new issue: the spell-check job is red on this commit.

### What stood out
- **Every prior finding is resolved** — and each was fixed using the narrower, per-project approach rather than widening the root config.
- **The new job fails on this commit** — 20 errors across `AGENTS.md`, `CONTRIBUTING.md` and a skill doc, caused by words that existed in the old config and did not reach any new file.
- **Restoring seven words turns the job green** — the gap is small and fully enumerated below.

### Top issues
- [F-005] Seven words from the old config reach no new config, failing the job — `cspell.yaml:223`

eng/pipelines/templates/jobs/analyze.yml
```diff
@@ -32,8 +32,6 @@ jobs:
       runtime: 'linux-x64'
       sourceDirectory: $(Build.SourcesDirectory)
```
wbreza, line 34: C-001 **The remaining stale path references are correctly left alone.**

Five references to the deleted `.vscode/cspell.json` still exist under `eng/common/**`:

- `eng/common/pipelines/templates/steps/check-spelling.yml:23`
- `eng/common/scripts/check-spelling-in-changed-files.ps1:32, 59`
- `eng/common/spelling/Invoke-Cspell.ps1:15, 46`

Leaving them is the right call, and this note is a record of why rather than a request to change anything.

`eng/common/README.md` states that updates in that directory must be made in the `azure-sdk-tools` repo, since changes made elsewhere are overwritten by future syncs. So these defaults are not this repo's to edit.

Nothing executes with the stale default either. This line removed the only consumer — no file under `eng/pipelines/` references `check-spelling.yml` anymore — and the GitHub Actions job passes `-CSpellConfigPath './cspell.yaml'` explicitly rather than relying on the parameter default. The root config also lists `**/eng/common/**` under `ignorePaths`, so those files are not scanned.

The follow-up, when someone gets to it, belongs in `azure-sdk-tools`: retarget the template and script defaults away from `.vscode/cspell.json` so other repos adopting the distributed layout inherit a working default.

wbreza (approved):
All previously raised findings are resolved and the spell-check job is green on this commit. Approving.

### What stood out
- **Word parity fully restored** — comparing the old 417-entry list against every word across all 67 config files now yields zero missing entries, and `slnx` landed in its alphabetical slot after `"skus"`.
- **The two non-migration errors got the right treatment** — `dislay` was corrected to `display` rather than dictionary-suppressed, while `myop`, `worktree` and `worktrees` were added as genuine new vocabulary.
- **Earlier fixes hold at this commit** — contributor docs carry the root-versus-project rule, `Fabric.Mcp.Tools.Docs` scopes `src/Resources/**` locally instead of widening the root, the three missing toolset configs are present, and the pipeline no longer double-enforces.

### Cross-cutting
- [C-001] **Remaining stale config references are correctly out of scope** _(process)_ — the `.vscode/cspell.json` defaults left in `eng/common/**` are sync-managed from another repo, ignored by the root config, and no longer referenced by any pipeline.

cspell.yaml
wbreza: Verified fixed in `58f109a5` — all seven words are back in the `baseline` dictionary, `slnx` in its alphabetical slot after `"skus"`, and the spell-check job passes on this commit. Re-ran the comparison of the old 417-entry list against every word across all 67 config files: zero missing.

Good call correcting `dislay` to `display` at the source rather than adding it to the dictionary, and treating `myop` / `worktree` / `worktrees` as new vocabulary.

eng/pipelines/templates/jobs/analyze.yml
wbreza, line 34: C-002 **Reverting this directory was the right call, and it leaves one follow-up upstream.**

`eng/common/spelling/README.md` is now byte-identical to `main`, and the PR touches no file under `eng/common/` at all. That matches the contract in `eng/common/README.md`, which states updates there must be made in `azure-sdk-tools` because changes made elsewhere are overwritten by future syncs — so the distributed-layout guidance that briefly lived here would have been lost on the next sync anyway.

Nothing in this repo regresses as a result. The four in-repo surfaces still carry the new rule:

- `AGENTS.md`
- `CONTRIBUTING.md`
- `docs/tool-rename-checklist.md`
- `tools/Azure.Mcp.Tools.AzureBackup/skills/azurebackup-add-tool/SKILL.md`

None of them retain a `.vscode/cspell.json` reference, and the spell-check job passes on this commit.

The shared scripts still default to the deleted `.vscode/cspell.json` (`eng/common/scripts/check-spelling-in-changed-files.ps1:59`, `eng/common/spelling/Invoke-Cspell.ps1:46`, `eng/common/pipelines/templates/steps/check-spelling.yml:23`), but nothing here executes with those defaults: this line removed the last consumer, the workflow passes `-CSpellConfigPath './cspell.yaml'` explicitly, and the root config ignores `**/eng/common/**`.

The remaining work is an `azure-sdk-tools` change to retarget those defaults, so other repos moving to a distributed layout inherit a working default. Nothing blocking on this PR.

wbreza (approved):
The `eng/common` revert is the right call and everything else still holds. Re-approving.

### What stood out
- **`eng/common/spelling/README.md` is back to its upstream state** — byte-identical to `main`, so the PR now touches nothing under that sync-managed directory and its guidance will not be clobbered by the next sync.
- **Nothing regressed with it** — the four in-repo doc surfaces still carry the root-versus-project rule with no stale path references left, and the spell-check job stays green.
- **Word parity is intact** — comparing the old 417-entry list against every word across all 67 config files still yields zero missing entries.

### Cross-cutting
- [C-002] **Retargeting the shared defaults now belongs upstream** _(process)_ — with this directory untouched, the stale `.vscode/cspell.json` defaults in the shared scripts are inert here and are best fixed in the repo that owns them.

cspell.json
```diff
@@ -0,0 +1,7 @@
+{
+  "//purpose": "This file is needed for older spell check versions to specify the files to lint",
```
danieljurek: This file should not be needed in this repo. Everything can point to `cspell.yaml` at the root since you're not using `.vscode/cspell.json` anymore.

servers/Template.Mcp.Server/cspell.yaml
```diff
@@ -0,0 +1,9 @@
+# Spell-check overrides for this toolset/server.
+# Inherits the global config at the repo root.
+#
+# Add words that only appear in this folder here, in alphabetical order.
+# Cross-cutting terms should go in the root cspell.yaml instead.
+version: "0.2"
+import:
+  - ../../cspell.yaml
+words:
```
danieljurek, line 9: cspell.yaml files are not required if there's nothing in them.

cspell.json
```diff
@@ -0,0 +1,7 @@
...
+  "version": "0.2",
+  "import": [
```
wbreza: C-003 **This shim is unused on the branch, and removing it is safe.**

Verifying the point raised in the other thread — the file is genuinely redundant here:

**Nothing reads it.** Searching the branch for `cspell.json` outside `eng/common/` returns exactly one hit: line 32 of the root `cspell.yaml`, where it adds itself to `ignorePaths`. No workflow, script, or pipeline references it. The spell-check job passes `-CSpellConfigPath './cspell.yaml'` explicitly.

**The stated purpose no longer holds.** The `//purpose` key describes supporting "older spell check versions." `eng/common/spelling/package.json` pins `cspell: ^10.0.0`, and cspell 10 discovers a bare `cspell.yaml` with no JSON entry point — linting a file in a subdirectory of a tree containing only `cspell.yaml`, with no `--config` and no `cspell.json`, correctly picks up that config's words. The older consumer that would have needed the JSON path was the Azure DevOps template, and no file under `eng/pipelines/` references it anymore after the earlier commit in this PR.

So deleting `cspell.json` and its `ignorePaths` entry leaves behavior unchanged for both CI and editors.

One caveat if it is kept instead: only the shim carries a `files` key today, so the two root configs are not interchangeable, and a future edit to one will not be reflected in the other. Dropping it removes that divergence risk. Either way this is a redundancy cleanup rather than a correctness problem, and it does not affect the spell-check result.

wbreza (commented):
Checked the open question about the root `cspell.json` — it is unused on this branch, and removing it is safe. Deferring the final call to that thread rather than re-approving over it.

### What stood out
- **Nothing on this branch reads `cspell.json`** — no workflow, script, or pipeline references it; its only mention anywhere is its own self-ignore entry in the root config.
- **The compatibility rationale no longer applies** — the repo pins cspell `^10.0.0`, which discovers a bare `cspell.yaml` on its own, and the older consumer that motivated the shim was removed earlier in this PR.
- **Code state is otherwise unchanged since the last pass** — same commit, spell-check job still green, all earlier findings still resolved.

### Cross-cutting
- [C-003] **The root `cspell.json` shim is redundant on this branch** _(process)_ — safe to delete; verification and the one caveat worth keeping in mind are in that thread.

servers/Template.Mcp.Server/cspell.yaml
wbreza, line 9: C-004 **These files are behavioral no-ops, so this is a convention call.**

Confirming the point raised in the other thread, and adding two details that bear on the decision.

**They genuinely do nothing.** I linted the same content twice against a root config — once under a folder holding a word-less child config with `import`, once under a folder with no config at all. Identical output both times: only the intended unknown word, nothing else. So deleting the 33 word-less files changes no spell-check result, and keeping them changes none either.

**It is half the added files.** 33 of the 66 per-project configs carry no words today, so whichever way this goes it is worth applying uniformly rather than case by case.

**One file in this shape must stay.** `tools/Fabric.Mcp.Tools.Docs/cspell.yaml` has no words but does carry `ignorePaths: [src/Resources/**]`, which suppresses roughly 1,884 hits across the 1,194 tracked Fabric API spec files. It reads as "empty" by a words-only check while doing real work, so any sweep to delete word-less files needs to exclude it.

The trade-off is a fair one either way: deleting them removes 33 inert files, while keeping them gives each toolset an obvious place to add its first word, with the header comment already stating the root-versus-project rule. That second property is what makes the distributed model discoverable, which is why the earlier three additions were suggested. Both are defensible, and the spell-check result is unaffected either way.

wbreza (commented):
Checked the second point about word-less config files. They are exact no-ops, so this is a convention call rather than a correctness one — leaving it to that thread.

### What stood out
- **Empty configs change nothing at runtime** — linting a file under a folder with a word-less config and under one with no config at all produces identical results.
- **Half the added files are in this category** — 33 of the 66 per-project configs carry no words today.
- **One of them must stay regardless** — `tools/Fabric.Mcp.Tools.Docs/cspell.yaml` is the only word-less file carrying real configuration, in its case an `ignorePaths` entry.

### Cross-cutting
- [C-004] **Word-less configs are no-ops; keeping or deleting them is a convention choice** _(process)_ — the trade-off and the one file that must be retained are in that thread.

cspell.yaml
```diff
@@ -0,0 +1,559 @@
...
+  - "**/.vscode/**"
+  - "**/.work/**"
+  - "**/NOTICE.txt"
+  - "**/TestResults/**"
+  - "**/bin/**"
+  - "**/eng/common/**"
+  - "**/obj/**"
+  - "**/tools/**/tests/test-resources-post.ps1"
+  - "cspell.yaml"
+  - "cspell.json"
```
wbreza, line 32: C-005 **This entry now points at a file that no longer exists.**

`cspell.json` was deleted in `b4122379`, so this line is the only remaining reference to it anywhere outside `eng/common/`.

It is harmless: an `ignorePaths` glob that matches nothing is silently ignored, and the spell-check job passes on this commit. Nothing to fix for correctness.

Worth dropping the line whenever this block is next touched, so the ignore list keeps describing files that actually exist. The `"cspell.yaml"` entry above it is still doing real work and should stay — it keeps the config's own word list from being spell-checked against itself.

Not worth a commit on its own.

wbreza (approved):
Removing the root `cspell.json` is confirmed safe — the job stays green and nothing referenced it. Approving.

### What stood out
- **The shim removal is clean** — only that file was deleted, and the spell-check job passes on this commit, matching the expectation that nothing depended on it.
- **Everything earlier still holds** — word parity remains at zero missing entries across all 67 configs, the four in-repo doc surfaces carry no stale path references, the Fabric ignore and three added toolset configs are intact, and this directory stays untouched.
- **One leftover line** — the root config still ignores the file that was just deleted; harmless, and noted below for whenever this area is next edited.

### Cross-cutting
- [C-005] **Dead `ignorePaths` entry for the removed shim** _(process)_ — cosmetic only, not worth a commit on its own.

danieljurek (approved):

wbreza (approved):
Re-confirming approval — the code is unchanged since the last pass and all checks remain green.

### What stood out
- **Same commit as the previous approval** — no new commits, diff changes, or review comments to evaluate.
- **Checks are green** — the spell-check and workflow-lint jobs both pass, with no failing checks on this head.
- **Everything verified earlier still holds** — word parity across all 67 configs, the migrated documentation, the Fabric ignore path, the three added toolset configs, and the untouched shared engineering directory.

The only open thread is the cosmetic leftover ignore entry noted previously, which is not worth a commit on its own and does not block merge.

PR #2997, opened 2026-07-06
Add onboarding welcome action for [ONBOARD] issues

yunjchoi:
Adds a GitHub Action that welcomes new contributors when an **Azure MCP** onboarding issue is filed.

**Trigger:** `issues` (`opened`, `labeled`), gated on **both** the `onboarding` and `Azure.Mcp.Server` labels applied by the *Azure MCP - Service Onboarding Request* template. Requiring both labels scopes this to Azure MCP only — the generic *New Microsoft MCP Server Onboarding Request* template also uses `onboarding` but pairs it with `Template.Mcp.Server`, so it will not trigger this action.

**Behavior:** Posts a comment pointing contributors to the [CONTRIBUTING.md](https://github.com/microsoft/mcp/blob/main/CONTRIBUTING.md) guide and encouraging them to try the `@onboarding` Copilot agent. A hidden marker prevents duplicate comments across the multiple label events.

**Permissions:** `issues: write` only.

### DIFF
--- a/.github/workflows/onboarding-welcome.yml
+++ b/.github/workflows/onboarding-welcome.yml
@@ -11,8 +11,8 @@ on:
 
 permissions:
   issues: write
-
 jobs:
+  welcome:
     runs-on: ubuntu-latest
     concurrency:
       group: onboarding-welcome-${{ github.event.issue.number }}
--- a/.github/workflows/onboarding-welcome.yml
+++ b/.github/workflows/onboarding-welcome.yml
@@ -25,7 +25,7 @@ jobs:
       contains(github.event.issue.labels.*.name, 'Azure.Mcp.Server')
     steps:
       - name: Comment with onboarding guidance
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             const marker = '<!-- onboarding-welcome-comment -->';
--- a/.github/workflows/onboarding-welcome.yml
+++ b/.github/workflows/onboarding-welcome.yml
@@ -17,7 +17,6 @@ jobs:
     concurrency:
       group: onboarding-welcome-${{ github.event.issue.number }}
       cancel-in-progress: true
-    runs-on: ubuntu-latest
     # Run only for Azure MCP onboarding requests. The Azure MCP onboarding
     # template applies the labels: onboarding, needs-triage, Azure.Mcp.Server.
     # Other onboarding templates (e.g. the generic "New Microsoft MCP Server
--- a/.github/workflows/onboarding-welcome.yml
+++ b/.github/workflows/onboarding-welcome.yml
@@ -1,0 +1,58 @@
+name: Onboarding Issue Welcome
+
+# Posts a welcome comment on Azure MCP onboarding requests, pointing new
+# contributors to the contribution guide and the @onboarding Copilot agent.
+# Onboarding issues are created from the "Azure MCP - Service Onboarding
+# Request" template, which applies the labels below.
+
+on:
+  issues:
+    types: [opened, labeled]
+
+permissions:
+  issues: write
+
+jobs:
+  welcome:
+    runs-on: ubuntu-latest
+    # Run only for Azure MCP onboarding requests. The Azure MCP onboarding
+    # template applies the labels: onboarding, needs-triage, Azure.Mcp.Server.
+    # Other onboarding templates (e.g. the generic "New Microsoft MCP Server
+    # Onboarding Request") also use the "onboarding" label but NOT
+    # "Azure.Mcp.Server", so we require both to scope this to Azure MCP.
+    if: >-
+      contains(github.event.issue.labels.*.name, 'onboarding') &&
+      contains(github.event.issue.labels.*.name, 'Azure.Mcp.Server')
+    steps:
+      - name: Comment with onboarding guidance
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const marker = '<!-- onboarding-welcome-comment -->';
+            const { owner, repo } = context.repo;
+            const issue_number = context.issue.number;
+
+            // Avoid posting duplicate comments. The template applies multiple
+            // labels, so several "labeled" events can fire for one issue.
+            const comments = await github.paginate(
+              github.rest.issues.listComments,
+              { owner, repo, issue_number, per_page: 100 }
+            );
+            if (comments.some(c => c.body && c.body.includes(marker))) {
+              core.info('Onboarding welcome comment already present. Skipping.');
+              return;
+            }
+
+            const body = [
+              marker,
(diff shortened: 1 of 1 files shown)


.github/workflows/onboarding-welcome.yml
```diff
@@ -0,0 +1,58 @@
...
+    # template applies the labels: onboarding, needs-triage, Azure.Mcp.Server.
+    # Other onboarding templates (e.g. the generic "New Microsoft MCP Server
+    # Onboarding Request") also use the "onboarding" label but NOT
+    # "Azure.Mcp.Server", so we require both to scope this to Azure MCP.
+    if: >-
+      contains(github.event.issue.labels.*.name, 'onboarding') &&
+      contains(github.event.issue.labels.*.name, 'Azure.Mcp.Server')
+    steps:
+      - name: Comment with onboarding guidance
+        uses: actions/github-script@v7
```
jongio: The closest sibling workflow in this repo (`auto-milestone-bugbash.yml`) uses `actions/github-script@v8`, and `doc-gap-detector.lock.yml` pins to v9 by SHA. This references `@v7`, which is a couple major versions behind. Consider bumping to at least `@v8` for consistency with the rest of the repo.

```suggestion
        uses: actions/github-script@v8
```

jongio (commented):
One version consistency note flagged inline.

KarishmaGhiya (approved):

g2vinay (approved):

KarishmaGhiya (approved):

.github/workflows/onboarding-welcome.yml
```diff
@@ -0,0 +1,58 @@
...
+    # template applies the labels: onboarding, needs-triage, Azure.Mcp.Server.
+    # Other onboarding templates (e.g. the generic "New Microsoft MCP Server
+    # Onboarding Request") also use the "onboarding" label but NOT
+    # "Azure.Mcp.Server", so we require both to scope this to Azure MCP.
+    if: >-
+      contains(github.event.issue.labels.*.name, 'onboarding') &&
+      contains(github.event.issue.labels.*.name, 'Azure.Mcp.Server')
+    steps:
+      - name: Comment with onboarding guidance
+        uses: actions/github-script@v8
```
jongio: The dominant pattern in this repo is SHA-pinned with a version comment (see `event-processor.yml`, `doc-gap-detector.lock.yml`, `scheduled-event-processor.yml`). Consider aligning for supply-chain safety and consistency.

```suggestion
        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
```

jongio (commented):
Incremental review of the v7 to v8 fix. The repo's other workflows pin `actions/github-script` by SHA at v9. One inline suggestion below.

.github/workflows/onboarding-welcome.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  issues:
+    types: [opened, labeled]
+
+permissions:
+  issues: write
+
+jobs:
+  welcome:
```
jongio, line 15: Nit: when the issue template applies multiple labels at creation, GitHub fires the `opened` event plus separate `labeled` events per label. With 3 template labels, you can get up to 4 concurrent runs (all but the first are no-ops thanks to the duplicate check). Adding a concurrency group avoids queuing redundant runs:

```suggestion
    runs-on: ubuntu-latest
    concurrency:
      group: onboarding-welcome-${{ github.event.issue.number }}
      cancel-in-progress: true
```

Not blocking since the duplicate check makes this functionally safe either way.

jongio (commented):
Logic, permissions, and duplicate-prevention all look correct. The label gating (`onboarding` + `Azure.Mcp.Server`) properly scopes to Azure MCP onboarding only, and the hidden marker prevents re-posting.

The SHA-pinning concern is already covered by existing feedback. One minor suggestion below about redundant workflow runs.

.github/workflows/onboarding-welcome.yml
```diff
@@ -0,0 +1,61 @@
...
+    types: [opened, labeled]
+
+permissions:
+  issues: write
+
+jobs:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: onboarding-welcome-${{ github.event.issue.number }}
+      cancel-in-progress: true
```
jongio: This block is malformed and the workflow won't run in its current form. The `welcome:` job id got dropped, so `jobs:` now maps `runs-on`, `concurrency`, `if`, and `steps` directly instead of nesting them under a job. On top of that `runs-on` is duplicated (here and a few lines down), which is a duplicate-key error. Actions will reject this with something like `jobs.runs-on must be a mapping`. Restore the job key and drop the duplicate:

```suggestion
jobs:
  welcome:
    runs-on: ubuntu-latest
    concurrency:
      group: onboarding-welcome-${{ github.event.issue.number }}
      cancel-in-progress: true
```

The `if:` and `steps:` lines below are already indented 4 spaces, so they'll sit correctly under `welcome:` once the job key is back.

jongio (commented):
One blocking issue from the latest push. The rest of the changes (SHA-pinning `github-script` and adding `concurrency` to dedupe the multiple `labeled` events) look right.

- `.github/workflows/onboarding-welcome.yml`: the `jobs:` block lost its `welcome:` job id and has a duplicated `runs-on`, so it no longer parses as a valid job. Details inline.

jongio (commented):
Incremental review of the latest push. The `welcome:` job id is restored, which fixes half of what I flagged, but the duplicate `runs-on` is still present so the workflow still won't parse. One inline note.

.github/workflows/onboarding-welcome.yml
```diff
@@ -0,0 +1,61 @@
...
+
+permissions:
+  issues: write
+jobs:
+  welcome:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: onboarding-welcome-${{ github.event.issue.number }}
+      cancel-in-progress: true
+    runs-on: ubuntu-latest
```
jongio: The `welcome:` job id is back, but the duplicate `runs-on: ubuntu-latest` from my earlier note is still here. Line 16 already sets it under `welcome:`, so this second one is a duplicate map key. GitHub's workflow parser rejects duplicate keys, so drop this line:

```suggestion
```

jongio (commented):
Confirmed the latest commit drops the duplicate `runs-on`, so `.github/workflows/onboarding-welcome.yml` now parses as a single valid `welcome` job (checked with a YAML parse). The concurrency dedupe on the issue number and the SHA-pinned `github-script@3a2844b...` (v9) are both in place, and the label gate still requires both `onboarding` and `Azure.Mcp.Server`. Nothing outstanding from me on this workflow.

KarishmaGhiya (approved):

PR #3000, opened 2026-07-07
Add support for producing `outputSchema` and `structuredContent`

vcolin7:
## Summary

Adds opt-in MCP `outputSchema` and `structuredContent` support without changing the default content-only behavior.

- Adds `ResultTypeInfo`/`SetResult` infrastructure and generates output schemas from source-generated `JsonTypeInfo`.
- Pilots command-specific schemas on five App Configuration commands.
- Adds `--structured-output-mode duplicated|compact`; `duplicated` preserves full historical text, while `compact` returns a short pointer to the complete structured result.
- Supports every execution mode: eligible direct tools in `all`, stable tagged envelopes for in-process `namespace` and `consolidated` routers, and a `single` envelope containing the complete downstream `CallToolResult`.
- Keeps errors actionable in `content` and omits `structuredContent` for errors.
- Adds migration guidance, changelog documentation, and comprehensive loader/schema tests.

Object result payloads are set in `structuredContent` as-is. Array and scalar roots are wrapped under `value` so they validate against MCP's required object-root output schema.

## Validation

- Azure Core tests: 1,037 passed, 18 skipped, 0 failed
- `dotnet build`
- `Build-Local.ps1 -VerifyNpx`
- CSpell

## GitHub issue number?

Fixes: https://github.com/microsoft/mcp-pr/issues/237

## Invoking Livetests


core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
```diff
@@ -114,4 +114,35 @@ public static JsonObject CreateInputSchema(IReadOnlyList<Option> options)
...
     }
+
+    /// <summary>
+    /// Builds the <c>outputSchema</c> root <see cref="JsonObject"/> for the supplied result type.
+    /// MCP requires the root schema to be an <c>"object"</c>, so result types that export as a
+    /// non-object root (arrays or scalars) are wrapped under a single <c>value</c> property. Unlike the
+    /// input schema, <c>additionalProperties</c> is intentionally left unset for forward compatibility
+    /// with evolving result shapes.
+    /// </summary>
+    public static JsonObject CreateOutputSchema(JsonTypeInfo resultTypeInfo)
```
wbreza, line 140: [F-001] `outputSchema` marks nullable result fields `required` while the payload omits them when null

**Severity: MEDIUM.** `CreateOutputSchema` runs `JsonSchemaExporter` with `TreatNullObliviousAsNonNullable = true`, so a positional record parameter with no default is emitted in the schema `required` array even when its type is nullable. The migrated results serialize through `AppConfigJsonContext`, which sets `DefaultIgnoreCondition = WhenWritingNull`, so a null value is omitted from `structuredContent`. The two disagree for `KeyValueSetCommandResult` (`Key`, `Value`, `Label`), `KeyValueDeleteCommandResult` (`Key`, `Label`), and `KeyValueLockSetCommandResult` (`Label`).

Reproduced for `(string? Key, string? Value, string? Label)`: the generated schema `required` is `["key","value","label"]`, while serializing an instance with `Label = null` yields `{"key":"mykey","value":"myval"}` — `label` is absent. A `keyvalue set`/`delete`/`lock set` call without `--label` (the common case) therefore returns a success payload that fails validation against its own advertised `outputSchema`, so an MCP client that validates `structuredContent` rejects a valid result.

Fix: reconcile the generated `required` set with the omit-null serialization — drop omit-null-eligible members from `required`, or serialize the null members — and add a no-`label` schema-conformance test for the three commands.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
```diff
@@ -0,0 +1,148 @@
...
+            : root;
+
+        return CreateToolList(toolsNode);
+    }
+
+    internal static JsonElement CreateToolResult(string command, string jsonResponse)
+    {
+        using var document = JsonDocument.Parse(jsonResponse);
+        JsonNode? result = null;
+        if (document.RootElement.TryGetProperty("results", out var results))
```
wbreza: [F-002] Compact aggregate mode discards the response of message-only commands

**Severity: MEDIUM.** `CreateToolResult(command, jsonResponse)` extracts only the `results` property. A `CommandResponse` carries its payload in either `Results` or `Message`; when a command sets only `Message` (with no `Results`, which is omitted via `WhenWritingNull`), this builds an envelope with `result: null`. In `namespace`/`consolidated` compact mode, `CreateNamespaceCallResult` additionally replaces the text content with `CompactContentMessage`.

A routed child command that returns its content through `context.Response.Message` (many guidance/best-practices/validation commands do) therefore returns the generic pointer text plus `result: null` — the actual content is lost. This contradicts the documented behavior that compact "returns a short pointer to the complete structured result." `duplicated` mode is unaffected because it keeps the full text.

Fix: carry `message` in the aggregate envelope when `results` is absent, or do not compact the text when there is no structured `results` to point at. Add a message-only success test.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs
```diff
@@ -0,0 +1,1093 @@
...
+using Microsoft.Mcp.Core.Areas.Server.Options;
+using Microsoft.Mcp.Core.Commands;
+using Microsoft.Mcp.Core.Models.Command;
+using Microsoft.Mcp.Core.Services.Telemetry;
+using NSubstitute;
+using Xunit;
+
+namespace Azure.Mcp.Core.Tests.Areas.Server;
+
+public class ServiceStartCommandTests
```
wbreza: [F-003] New `ServiceStartCommandTests` duplicates the entire existing `ServerStartCommandTests` suite

**Severity: MEDIUM.** This new 1093-line file exercises the same class (`private readonly ServerStartCommand _command = new();`), and its 46 test methods include all 44 method names already present in the pre-existing `ServerStartCommandTests.cs`. Only `StructuredOutputModeOption_ParsesCaseInsensitiveValues` and `StructuredOutputModeOption_DefaultsToDisabled` are new. The 44 shared tests now run and must be maintained in two places, doubling their CI time and creating drift risk.

Fix: add the two new structured-output tests to the existing `ServerStartCommandTests.cs` and delete this file. If a rename to `ServiceStartCommandTests` was intended, delete the original instead so the suite is not duplicated.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -535,43 +561,47 @@ There was an error finding or calling tool and command.
                             Run again with the "learn=true" to get a list of available commands and their parameters.
                             """
                     }
-                ]
+                ],
+                IsError = true
```
wbreza, line 544: [F-004] Error path now sets `IsError = true` unconditionally — an undisclosed default-behavior change

**Severity: LOW.** The `InvokeChildToolAsync` catch fallback ("There was an error finding or calling tool and command.") now returns `IsError = true`; previously it left `IsError` unset (null/false). This changes the error wire-contract for every client independent of `--structured-output-mode`, while the PR states it adds structured output "without changing the default content-only behavior." It is also inconsistent with the sibling `SingleProxyToolLoader.CommandModeAsync` catch, which still returns no `IsError`. The new value is more correct, so real impact is narrow, but the change is unconditional and undisclosed.

Fix: apply the same `IsError = true` to the `SingleProxyToolLoader` error path for consistency, and note this error-contract change in the description/changelog.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
```diff
@@ -0,0 +1,148 @@
...
+
+namespace Microsoft.Mcp.Core.Areas.Server.Commands.ToolLoading;
+
+internal static class AggregateStructuredOutput
+{
+    internal static JsonElement NamespaceOutputSchema { get; } = CreateOutputSchema(requireTool: false);
+
+    internal static JsonElement SingleOutputSchema { get; } = CreateOutputSchema(requireTool: true);
+
+    internal static JsonElement CreateToolList(IEnumerable<Tool> tools)
```
wbreza: [F-005] Dead code: the `CreateToolList(IEnumerable<Tool>)` overload has no caller

**Severity: LOW.** `AggregateStructuredOutput.CreateToolList(IEnumerable<Tool> tools)` is never invoked. The namespace and single-proxy learn paths all use the `CreateToolList(string)` overload (`NamespaceToolLoader.cs:589`; `SingleProxyToolLoader.cs:267,302`), and no test references the typed overload.

Fix: remove the unused overload, or route a caller through it if typed input was the intent.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs
```diff
@@ -0,0 +1,1093 @@
...
+
+        // Assert
+        var hasToolOption = command.Options.Any(o => o.Name == "--tool");
+        Assert.True(hasToolOption, "Tool option should be registered");
+    }
+
+    [Theory]
+    [InlineData("duplicated", StructuredOutputMode.Duplicated)]
+    [InlineData("compact", StructuredOutputMode.Compact)]
+    public void StructuredOutputModeOption_ParsesCaseInsensitiveValues(
```
wbreza: [F-006] Test gaps around the new structured-output surface

**Severity: LOW.** Several new-code paths are under-verified:
- The duplicated-mode loader tests assert only that the text is not the compact sentinel (`text == StructuredOutputHelper.CompactContentMessage`), not that the historical text is preserved exactly — yet exact preservation is the compatibility guarantee of `duplicated` mode.
- `StructuredOutputModeOption_ParsesCaseInsensitiveValues` supplies only lowercase inputs, so a regression to case-sensitive parsing would still pass.
- `SetResult`'s guard paths (`ArgumentNullException` on null context; `InvalidOperationException` when `ResultTypeInfo` is null) and `CreateOutputSchema(null)` have no tests.

Fix: assert exact text equality for duplicated mode, add mixed/upper-case option inputs, and cover the `SetResult`/`CreateOutputSchema` guard paths.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
```diff
@@ -114,4 +114,35 @@ public static JsonObject CreateInputSchema(IReadOnlyList<Option> options)
...
+
+        if (schema is JsonObject rootObject && IsObjectRoot(rootObject))
+        {
+            return rootObject;
+        }
+
+        return new JsonObject
+        {
+            ["type"] = "object",
+            ["properties"] = new JsonObject { ["value"] = schema },
```
wbreza: [O-001] 💡 Improvement Opportunity (optional — not required to merge)

**Benefit:** The object-root "wrap a non-object result under `value`" rule and the envelope-kind / `value` string literals are implemented independently in `OptionSchemaGenerator.CreateOutputSchema`, `StructuredOutputHelper.TryBuildStructuredContent`, and `AggregateStructuredOutput`. A single shared helper plus named constants keeps the schema side and the payload side from drifting — the drift that produces F-001.
**Effort:** Small.
**Optionality:** Recommended.
**Sketch:** Extract one `WrapNonObjectRoot`/`IsObjectRoot` helper and `kind`/`value` constants, and call them from both the schema and content paths.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/StructuredOutputHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+
+internal static class StructuredOutputHelper
+{
+    internal const string CompactContentMessage = "Response successful. See structuredContent for details.";
+
+    /// <summary>
+    /// Extracts and shapes the result payload from a serialized <see cref="CommandResponse"/>.
+    /// Object payloads are used as-is; arrays and scalars are wrapped under <c>value</c>.
+    /// </summary>
+    internal static JsonElement? TryBuildStructuredContent(string jsonResponse)
```
wbreza: [O-002] 💡 Improvement Opportunity (optional — not required to merge)

**Benefit:** `TryBuildStructuredContent` and `AggregateStructuredOutput` re-parse an already-serialized `CommandResponse` string (`JsonDocument.Parse` + `GetRawText` + `JsonNode.Parse`/`DeepClone`), and compact mode serializes the full response to a string only to discard it for the pointer message. Serializing the typed result directly to a `JsonNode`/`JsonElement` removes the round-trips and the wasted large-string allocation on the call hot path.
**Effort:** Medium.
**Optionality:** Recommended.
**Sketch:** Add a `ResponseResult` path that writes its value to a `JsonElement`, pass a node to the shapers, and skip building the full `jsonResponse` string in compact mode.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -157,6 +160,29 @@ Sub commands are routed to MCP servers that require specific fields inside the "
         return ValueTask.FromResult(allToolsResponse);
     }
 
+    private CallToolResult CreateNamespaceCallResult(
```
wbreza: [O-003] 💡 Improvement Opportunity (optional — not required to merge)

**Benefit:** `CommandFactoryToolLoader.CallToolHandler` (inline), `NamespaceToolLoader.CreateNamespaceCallResult`, and `SingleProxyToolLoader.CreateAggregateCallResult` each repeat the same compact-vs-duplicated content selection and conditional `StructuredContent` assembly. One shared factory removes the triplication.
**Effort:** Small.
**Optionality:** Recommended.
**Sketch:** Move a `BuildCallToolResult(mode, text, structuredFactory, isError)` helper into `StructuredOutputHelper` and call it from all three sites.

wbreza (commented):
Well-documented opt-in structured-output support; resolve three schema/compatibility correctness issues before the feature is enabled.

### What stood out
- **Clean opt-in surface** — structured output is gated behind `--structured-output-mode`; when the option is omitted responses stay content-only, and a single source-generated `JsonTypeInfo` drives both the output schema and the payload.
- **Thorough documentation** — a step-by-step migration guide, a changelog entry, an `azmcp-commands.md` row, and a stable tagged-envelope contract (`tool-list`/`tool-result`/`message`) for the aggregate `namespace`/`consolidated`/`single` modes.
- **Broad test intent** — object-vs-array/scalar wrapping, compact-vs-duplicated content, and error omission are all exercised across the loaders.

### Top issues
- [F-001] Advertised `outputSchema` marks nullable result fields `required`, but the payload omits them when null — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs:125`
- [F-002] Compact aggregate mode drops the response of message-only commands — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs:37`
- [F-003] New `ServiceStartCommandTests` duplicates the entire existing `ServerStartCommandTests` suite — `core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs:20`

### Improvement opportunities
- [O-001] Centralize the object-root wrapping rule and envelope-kind constants shared by schema and payload _(Recommended)_ — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs:139`
- [O-002] Serialize the typed result straight to a node instead of serialize→parse→clone round-trips _(Recommended)_ — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/StructuredOutputHelper.cs:18`
- [O-003] Consolidate the three near-identical `CallToolResult` envelope factories _(Recommended)_ — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs:163`

wbreza (commented):
## Re-review — 2 commits since `df7b165`

**Still open (6):**
- 🟡 MEDIUM F-001 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs:125`
- 🟡 MEDIUM F-002 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs:37`
- 🟡 MEDIUM F-003 — `core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs:20`
- 🟢 LOW F-004 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs:565`
- 🟢 LOW F-005 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs:16`
- 🟢 LOW F-006 — `core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs:133`

No new issues in this delta — the 2 commits add `noschema` to the Azure Core cspell dictionary and reorder `using` directives in `NamespaceToolLoaderTests.cs`.

3 optional opportunities still available.

---
**To approve:** resolve F-001, F-002, F-003

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
vcolin7: Fixed on the schema side rather than changing serialization, preserving the existing omit-null payload shape.

After `JsonSchemaExporter` generates an object schema, `CreateOutputSchema` now reconciles its `required` array with the source-generated `JsonTypeInfo` serialization settings:

- `WhenWritingNull`: removes nullable properties from `required`.
- `WhenWritingDefault`: removes properties from `required`, since any default value may be omitted.
- Removes the `required` keyword entirely if it becomes empty.

Nullable detection handles `type: "null"`, type unions containing `null`, and `anyOf`/`oneOf` unions. Shared object-root and `value` wrapping logic was also centralized in `StructuredOutputJson` to prevent schema/payload drift.

Tests use the actual AppConfig set/delete/lock result types with `WhenWritingNull`, serialize no-label responses, and verify every property still advertised as required is present in the payload.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
vcolin7: `AggregateStructuredOutput.CreateToolResult` now accepts the typed `CommandResponse`. It uses `Results` when present; otherwise it places a non-empty `Message` into `result.message`. If neither exists, it returns no structured content, so compact mode retains the full text instead of replacing it with a pointer to nothing.

Before:

```json
{
  "content": "Response successful. See structuredContent for details.",
  "structuredContent": { "kind": "tool-result", "result": null }
}

```

After:

```json
{
  "content": "Response successful. See structuredContent for details.",
  "structuredContent": {
    "kind": "tool-result",
    "command": "read-cmd",
    "result": { "message": "Use zone-redundant storage." }
  }
}
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs
vcolin7: Removed duplicated code.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
vcolin7: In `SingleProxyToolLoader.CommandModeAsync`, the catch result now explicitly sets `IsError = true`.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
vcolin7: The unused `CreateToolList(IEnumerable<Tool>)` overload has been deleted.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/ServiceStartCommandTests.cs
vcolin7: 1. **Exact duplicated-mode compatibility.** Tests previously checked only that duplicated text was not the compact sentinel. They now assert the complete historical response exactly:

    - Namespace child calls equal the serialized `CommandResponse`.
    - Single-proxy child calls retain the downstream text exactly.
    - Namespace and single-proxy learn responses equal their full guidance text plus tool JSON.

2. **Real case-insensitivity coverage.** The option test now supplies `DuPlIcAtEd` and `COMPACT`; lowercase values could not detect a regression to case-sensitive parsing.

3. **`SetResult` guard coverage.** A small test-only wrapper exposes the protected helper and verifies:

    - A null `CommandContext` throws `ArgumentNullException`.
    - Calling it without overriding `ResultTypeInfo` throws `InvalidOperationException` with useful context.

4. **Schema-generator null guard.** `CreateOutputSchema(null)` is now verified to throw `ArgumentNullException`.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
vcolin7: Added a central class that contains these values to avoid drift.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/StructuredOutputHelper.cs
vcolin7: **Removed serialization round-trips (O-002):** `ResponseResult` gained an internal `ToJsonNode()` that serializes with its original source-generated `JsonTypeInfo`. The shapers now consume `ResponseResult`/`CommandResponse` directly instead of serializing the full response to a string and parsing it again.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
vcolin7: **Centralized compact/duplicated behavior (O-003):** `StructuredOutputHelper.CreateCallToolResult` now owns the shared rules: structured content is emitted only when enabled and successful; compact text is used only when structured content actually exists; otherwise historical full content is retained. The command-factory, namespace, and single-proxy loaders all use this helper.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
```diff
@@ -0,0 +1,153 @@
...
+
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using Microsoft.Mcp.Core.Models.Command;
+
+namespace Microsoft.Mcp.Core.Areas.Server.Commands.ToolLoading;
+
+internal static class AggregateStructuredOutput
+{
+    internal static JsonElement NamespaceOutputSchema { get; } = CreateOutputSchema(requireTool: false);
```
alzimmermsft, line 12: Based on my understanding of this change so far, this looks like were not actually including the outputSchemas of all tools in a namespace, but making a generic object type that just covers them. Basically, outputSchema = additionalProperites.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
```diff
@@ -0,0 +1,153 @@
...
+            toolResultProperties[StructuredOutputJson.ResultPropertyName] = new JsonObject { ["type"] = "object" };
+            toolResultRequired.Insert(1, StructuredOutputJson.ToolPropertyName);
+        }
+
+        var toolListSchema = new JsonObject
+        {
+            ["type"] = "object",
+            ["properties"] = new JsonObject
+            {
+                [StructuredOutputJson.KindPropertyName] = new JsonObject { ["const"] = StructuredOutputJson.ToolListKind },
```
alzimmermsft: Do these `const` properties actually gain us any benefit? In theory someone could write a parser for the different kinds but we don't publish model definitions anywhere (but I wouldn't want to make that a contract we guarantee).

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/StructuredOutputJson.cs
alzimmermsft, line 1: nit: Can this be merged into `StructuredOutputHelper`?

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
```diff
@@ -114,4 +114,85 @@ public static JsonObject CreateInputSchema(IReadOnlyList<Option> options)
...
+
+        return new JsonObject
+        {
+            ["type"] = "object",
+            ["properties"] = StructuredOutputJson.WrapValue(schema),
+            ["required"] = new JsonArray { (JsonNode)StructuredOutputJson.ValuePropertyName },
+        };
+    }
+
+    private static void ReconcileRequiredProperties(JsonObject schema, JsonIgnoreCondition defaultIgnoreCondition)
```
wbreza: [F-007] Per-property null omission remains in the advertised required set

`ReconcileRequiredProperties` exits unless the type-level `DefaultIgnoreCondition` is `WhenWritingNull` or `WhenWritingDefault`. `VmGetResult` has per-property `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` annotations while `ComputeJsonContext` has no default ignore condition. The migration guide directs a command to reuse its existing source-generated `JsonTypeInfo`; after that documented migration, the generated schema retains the omitted members in `required`, while result serialization excludes them when null. A validating MCP client rejects those successful result shapes.

Determine each property's effective ignore policy, including property-level `JsonIgnore`, before retaining it in `required`, and add a schema/payload conformance test for that result style.

wbreza (commented):
## Re-review — 8 commits since `2909a0a`

**Resolved (9):** F-001, F-002, F-003, F-004, F-005, F-006, O-001, O-002, O-003

**New (1):**
- 🟡 MEDIUM F-007 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs:145` — schema reconciliation ignores per-property null-omit configuration.

---
**To approve:** resolve F-007

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
vcolin7: Yeah, you could see it that way. For aggregate modes we're not including all the `outputSchemas` but instead showing the common fields we return on these like `kind`, `command`, and `result`, with the latter being opaque.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
vcolin7: I just thought that if we were going to establish a sort of "contract" with `outputSchema` for clients consuming `structuredContent` we could provide a bit more info on what the response can be interpreted as.

We can get rid of this if you think it's better for the time being.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/StructuredOutputJson.cs
vcolin7: Yeah, we can merge them :)

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
vcolin7: Good catch. Updated this to look at each property’s `JsonIgnore` setting instead of only the default setting from the serializer context. The same check now applies to nested result objects as well. I added coverage for `VmGetResult` and all `JsonIgnoreCondition` values.

*Written with the help of vcolin7-copilot.*

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/AggregateStructuredOutput.cs
vcolin7: Keeping for now after discussing it offline

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
```diff
@@ -114,4 +129,119 @@ public static JsonObject CreateInputSchema(IReadOnlyList<Option> options)
...
+        if (propertyInfo.ShouldSerialize is not null)
+        {
+            // Source-generated metadata uses ShouldSerialize for property-level JsonIgnore conditions.
+            // Read the attribute to distinguish Never from conditions that can omit the property.
+            ignoreAttribute = propertyInfo.AttributeProvider?
+                .GetCustomAttributes(typeof(JsonIgnoreAttribute), inherit: true)
+                .OfType<JsonIgnoreAttribute>()
+                .SingleOrDefault();
+
+            if (ignoreAttribute is null)
```
wbreza, line 205: [F-008] Custom `ShouldSerialize` handling has no regression coverage.

**Severity: LOW.** The new fallback removes a property from `required` when `JsonPropertyInfo.ShouldSerialize` is set but no `JsonIgnoreAttribute` is present. The added conformance tests exercise source-generated `JsonIgnore` conditions, but none supplies a custom predicate. This leaves the fallback's schema/payload contract unprotected against regression.

Add a test type-info modifier with a custom `ShouldSerialize` predicate that omits a property, then assert that member is absent from the generated `required` set.

wbreza (approved):
## Re-review — 6 commits since `7beb6f9`

**Resolved (10):** F-001, F-002, F-003, F-004, F-005, F-006, F-007, O-001, O-002, O-003

**New (1):**
- 🟢 LOW F-008 — `core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs:205` — custom `ShouldSerialize` coverage is absent

---
**Approved.**

alzimmermsft (approved):

PR #3144, opened 2026-07-23
Enforce code styles in build

alzimmermsft:
## What does this PR do?

Makes some changes to the build to enforce code style rules while building the project and enabling more diagnostics to be raised.

- Standardized verbiage for `CancellationToken` parameters.
- Require all CSharp files to have header `Copyright (c) Microsoft Corporation.\nLicensed under the MIT License.`.
- Made incorrect CSharp modifier order an error instead of a suggestion.
- Require `using` directives to be outside of `namespace` scope.
- Require `namespace` to be defined using file scoping instead of block scoping.
- Require `namespace` to match folder structure.
- Raise an error on any unused parameters, including public APIs as we ship as a compiled binary and not an SDK.
- Raise an error on unused `using` directives.
- Raise an error on invalid string formatting.
- Raise an error on unused and unread private members.
- Raise an error when a synchronous method is declared asynchronous.
- Validate doc string formatting (missing closing tags, unescaped special character) and require `cref` types to be declared in `using` directives correctly.

And fixed / suppressed (when valid to do so) all found issues due to the enforcement of the above.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:


core/Azure.Mcp.Core/src/Services/Azure/AzureServiceCollectionExtensions.cs
```diff
@@ -29,11 +29,12 @@ public static class TenantServiceCollectionExtensions
     /// <list type="bullet">
     /// <item>
     /// <description>
-    /// <see cref="IAzureTokenCredentialProvider"/> via <see cref="Authentication.AuthenticationServiceCollectionExtensions.AddSingleIdentityTokenCredentialProvider"/>.
-    /// This can be overridden using <see cref="Authentication.AuthenticationServiceCollectionExtensions.AddAzureTokenCredentialProvider"/>
```
vukelich, line 33: Are the lines about overriding no longer true? I believe I left those clues in the doc comments intentionally when adding this code due to some design issues at app start up.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -157,7 +157,9 @@ protected async Task<ResourceQueryResults<T>> ExecuteResourceQueryAsync<T>(
     /// <param name="subscription">The subscription ID or name</param>
     /// <param name="retryPolicy">Optional retry policy configuration</param>
     /// <param name="converter">Function to convert JsonElement to the target type</param>
+    /// <param name="tableName">Optional table name to query (default: "resources")</param>
```
vukelich, line 157: Do or don't: I'm torn on `(default: "resources")` in the doc comment. 

From a maintainability stance, this can easily diverge from the code. Other places solve this by defining a `static string` or `const string`, so that both the code and the doc comment resolve to the same `string` value. For example, see [`IOptionsSnapshot<TOptions>.Get(String)`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.ioptionssnapshot-1.get?view=net-11.0-pp#microsoft-extensions-options-ioptionssnapshot-1-get(system-string)). 

(Aside: If a `const string`, then the value can be used in the default parameter syntax because its compile-time. If `static string`, then it must be used in the method body BUT the value is a runtime value not a compiler-time value, which allows for dependent libraries to use the new value without being recompiled. The .NET [`Options.DefaultName`](https://source.dot.net/#Microsoft.Extensions.Options/Options.cs,7326828236b59598,references) is `static`.)

On the other hand, the `"resources"` table name does have direct ARG meaning, so it provides the reader some better understanding of the system. 

Though, using that table is the default behavior of the ARG API. We could be fine with a `null` default value and let ARG's query engine do what it does. Switching to letting ARG handle it would be more scope than this code change wants, but it can be beneficial to reduce our own business logic when it exists downstream already.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -225,6 +228,7 @@ protected static async Task<GenericResource> GetGenericResourceAsync(ArmClient a
     /// <param name="azureLocation">The Azure location for the resource.</param>
     /// <param name="content">The content to create or update the resource with.</param>
     /// <param name="jsonTypeInfo">The JSON type information for serialization.</param>
+    /// <param name="cancellationToken">Cancellation token</param>
```
vukelich: Minor: `A cancellation token.` fits a little better since the other parameters are also expressed as sentences. 

I won't repeat this comment for each place that it's applicable. I'm sure the repo has inconsistent language for this parameter.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -44,12 +45,16 @@ static BaseAzureService()
         s_defaultUserAgent = $"azmcp/{s_version} ({s_framework}; {s_platform})";
         s_sharedUserAgentPolicy = new UserAgentPolicy(s_defaultUserAgent);
 
-#if DEBUG
+        SetDefaultPollInterval();
+    }
+
+    [Conditional("DEBUG")]
```
vukelich: This is not necessarily equivalent to the previous code. Using `[Conditional(...)]` will include or omit invocations to the attached method, but it keeps the method body in the source code. If unused code paths are removed by other compliation configurations, then it'll be similar to what we had before.

What is the aim for this change? Was there an analyzers requiring it?

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ServiceCollectionExtensionsSerializedTests.cs
```diff
@@ -65,7 +61,7 @@ public void InitializeConfigurationAndOptions_HttpTransport()
         {
             Transport = TransportTypes.Http,
         };
-        var services = SetupBaseServices().AddSingleton(Options.Create(serviceStartOptions));
+        var services = SetupBaseServices().AddSingleton(Microsoft.Extensions.Options.Options.Create(serviceStartOptions));
```
vukelich: I dislike having to do this. Can we instead do...?

```csharp
using Options = Microsoft.Extensions.Options.Options;
```

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/AccessTokenHandlerTests.cs
```diff
@@ -4,6 +4,7 @@
 using System.Net;
 using System.Net.Http.Headers;
 using Azure.Core;
+using Microsoft.Mcp.Core;
```
vukelich, line 7: To other reviewers: this new `using` is because [AccessTokenHandler.cs‎](https://github.com/microsoft/mcp/pull/3144/changes#diff-d3467525adde30a3350e89bf0b77daee7806d9f1a0507ecfb568d2f2dbe0e62e) changed namespaces in this PR.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
```diff
@@ -23,13 +23,7 @@ namespace Microsoft.Mcp.Core.Areas.Server.Commands;
 /// into build errors. Option value types come from <c>System.CommandLine</c> as
 /// runtime <see cref="Type"/> values, so this helper has to use
 /// <see cref="DefaultJsonTypeInfoResolver"/> and
-/// <see cref="JsonStringEnumConverter"/>, both annotated. The
-/// <see cref="UnconditionalSuppressMessageAttribute"/> attributes below are used
```
vukelich, line 27: Please don't entirely get rid of the extra context. This sound helpful for any future maintainer.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ServiceCollectionExtensions.cs
```diff
@@ -46,7 +42,7 @@ public static IServiceCollection AddAzureMcpServer(this IServiceCollection servi
 
         // Register options for service start
         services.AddSingleton(serviceStartOptions);
-        services.AddSingleton(Options.Create(serviceStartOptions));
+        services.AddSingleton(Microsoft.Extensions.Options.Options.Create(serviceStartOptions));
```
vukelich: Same as before about aliasing rather than fully qualifying, please.

core/Microsoft.Mcp.Core/src/Areas/Tools/Commands/ToolsListCommand.cs
```diff
@@ -78,15 +78,11 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
                 }
 
                 // If --name-only is also specified, return only the names
-                if (options.NameOnly)
-                {
-                    var namespaceNames = namespaceCommands.Select(nc => nc.Command).ToList();
-                    context.Response.Results = ResponseResult.Create(new(null, namespaceNames), ModelsJsonContext.Default.ToolsListResult);
-                    return context.Response;
-                }
+                context.Response.Results = options.NameOnly
```
vukelich: Couple things:

**Harder to read than before** 

I misread the new line 82 because of amount of code now jammed into a single line. Using a proper `if/else` here is totally fine. We don't need to nor would I encourage making one line harder to read. By being one line with lots going on, exceptions are harder to debug. The exception information will often include the line number, which in this case could be many different subexpression method calls. This one line has seven method invocations (`Create`, a constructor, `Select`, `ToList`, and three properties which are compiled to `get_PropertyNameHere()`).

If we really need to do a single expression, then please explain why and then make a change to put the method invocations on new lines with indentation based on parameter positioning. For example:

```csharp
context.Response.Results = options.NameOnly
    ? ResponseResult.Create(
        new(
            null,
            namespaceCommands.Select(nc => nc.Command).ToList()),
        ModelsJsonContext.Default.ToolsListResult)
    : ResponseResult.Create(
        new(namespaceCommands, null),
        ModelsJsonContext.Default.ToolsListResult);
```

I struggle to read method invocations with many parameters on the same line and especially when a parameter has subexpressions (e.g., method calls). It also makes looking at diff history more annoying when a single parameter is added or removed. This repo has a lot of this writing style due to AI.

** Bonus points**

If readability is a goal, then using named parameters for `null` arguments provides value.

I'd like us to enable the analyzer that would require us to use the full type name at constructor invocations except when already clear on the LHS of a statement. Just reading `new()` here does not tell me anything as a static code reviewer. Don't do that in this PR. I'm sure you'll get lots of noise, and I don't want to muddy up the great progress you've already made.

core/Microsoft.Mcp.Core/src/Commands/ICommandFactory.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.CommandLine;
+using Microsoft.Mcp.Core.Models.Command;
```
vukelich, line 5: Why needed?

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
```diff
@@ -125,13 +108,35 @@ private static void EnableAzureMonitor(this IServiceCollection services)
...
     /// Configures OpenTelemetry to use Azure Monitor exporters with Microsoft's Application Insights instance.
     /// </summary>
     /// <param name="otelBuilder">The OpenTelemetry builder to configure.</param>
     /// <param name="appInsightsConnectionString">The Application Insights connection string for Microsoft's telemetry instance.</param>
+    [Conditional("RELEASE")]
     private static void ConfigureMicrosoftAzureMonitorExporter(OpenTelemetry.OpenTelemetryBuilder otelBuilder, string appInsightsConnectionString)
     {
+        // This environment variable can be used to disable Microsoft telemetry collection.
+        // By default, Microsoft telemetry is enabled.
+        var microsoftTelemetry = Environment.GetEnvironmentVariable("AZURE_MCP_COLLECT_TELEMETRY_MICROSOFT");
```
vukelich: In the future, when making changes like this to change the compilation conditionality handling, please add your own PR comment to explain why this is equivalent or equivalent enough. I don't expect you're thoroughly testing debug vs release in remote mode for this change, so this is putting onerous risk on me, your reviewer, to make sure this is correct.

vukelich: TODO: Steven, return to this later with at minimum his static analysis reading of this in more depth.

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/IAzureTokenCredentialProvider.cs
```diff
@@ -14,8 +14,7 @@ namespace Microsoft.Mcp.Core.Services.Azure.Authentication;
 /// </summary>
 /// <remarks>
 /// <para>
-/// <b>Callers</b> can either directly depend on this interface or indirectly depend on it through
-/// <see cref="ITenantService"/>.
+/// <b>Callers</b> can either directly depend on this interface or indirectly depend on it through ITenantService.
```
vukelich: Incorrect change. We should add a `using` statement so that `ITenantService` resolves as a symbol.

core/Microsoft.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -431,7 +431,7 @@ public ProcessStreamReader(Process process, bool isErrorStream, ILogger<External
 
         /// <summary>
         /// Begins asynchronous reading of the associated stream.
-        /// Must be called only after <see cref="Process.Start"/> has successfully completed.
+        /// Must be called only after <see cref="Process.Start()"/> has successfully completed.
```
vukelich, line 434: Useful fact: I often omit the parentheses because the compiler handles it very well, sometimes better than if I were to include the parameters. This is especially helpful when the method you're describing to a reader has overloads, but you don't want/need to be specific about a single overload.

Might also just be a feature of Visual Studio to handle it very well and I'm mistaking it for a compiler feature.

core/Microsoft.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
```diff
@@ -604,9 +604,9 @@ public static ExitCheckResult CheckExitState(this Process process, ILogger logge
         {
             // Official docs: "No process is associated with this object." - treat as "already gone".
             logger.LogDebug(
-            checkException,
-            "Process.HasExited reported no associated process. Treating as already exited. " +
-            "Process: {ProcessName}, PID: {Pid}", process.SafeName(), process.SafeId());
+                checkException,
+                "Process.HasExited reported no associated process. Treating as already exited. " +
+                "Process: {ProcessName}, PID: {Pid}", process.SafeName(), process.SafeId());
```
vukelich: Good catch about indentation, but there's a little more to go for a good multi-line method invocation. The beginning of line 609 is the continuation of the expression (string concat) started on 608. It should be indented further. The last two parameters of `LogDebug` should be on their own lines, too.

```csharp
logger.LogDebug(
    checkException,
    "Process.HasExited reported no associated process. Treating as already exited. "
        + "Process: {ProcessName}, PID: {Pid}",
    process.SafeName(),
    process.SafeId());
```

The above is a requested fix. Below is a subjective rant on how to handle the string concatenation operator:

You'll notice that I put the string concatenation (`+`) on the indented line. I like putting expression operators on the next line because I like reading it like the following:

```text
myNewString is "Some bit of text"
    and " with some more bits of text"
    and " ended with this last bit of text."
```

In this way, all the information about the expression being split is on the left side: the extra indentation and the introduction of subexpression operands. 

It's also how I write KQL queries:

```kql
myTable
| where someCol = 1
    and otherCol = 2
    and anotherCol = 3
```

With saved and ad-hoc KQL queries, this style of writing has the added benefit of making commenting-out individual subexpressions easier. I can comment out any line starting with `and` to still be syntactically valid:

```kql
myTable
| where someCol = 1
    and otherCol = 2
//    and anotherCol = 3
```

In rare cases, I've caught logic issues due to unintentional mixing and matching operators, like with the following. I'm quicker to notice and ask if `or` was intentional, and react accordingly:

```kql
myTable
| where someCol = 1
    and otherCol = 2
    and anotherCol = 3
    or fourCol = 4   // Easy to see the "or" stick out in a line of "and" operators
    and fiveCol = 5
```

This operator location can also come into play when considering the dot operator (`.`) which I follow the same rule as the above. This is less common, but it's consistent!

```csharp
SomeTopLevelMethod(
    UnfinishedFirstArg
        .SomePropertyWeLike
        .TransformItALittle(),
    ReadyToGoSecondArg)
```

I don't consider the comma of a method invocation an operator. The method is the operator, and the comma is syntax to separate the operands (parameters) of the operator (method). Some languages use whitespace for separation, like PowerShell. For this reason, I don't tend to put the comma at the beginning of each line, but I can see why others may.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Generated/MicrosoftClientModelTestFrameworkModelFactory.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#pragma warning disable CS1574
```
vukelich, line 4: I was going to comment to ask for an in-line comment about why this is being suppressed, but then I noticed this file claims to be generated.

1. Why were we getting CS1574 errors?
2.  Should we be requiring generated code to follow CS1574?
3. If we expect CS1574 to be honored, what is generating bad doc comments?
4. If we don't need CS1574 to be honored, do we need to add this disabling in a way that survives regeneration?

eng/tools/ToolDescriptionEvaluator/src/Program.cs
```diff
@@ -825,7 +825,9 @@ await Task.WhenAll(
         }
     }
 
+#pragma warning disable IDE0060 // Should use isCiMode parameter, but it's not used in this method.
```
vukelich: Is this a real TODO or not? Should we just remove the parameter from the method signature? If it's a real TODO but we don't want to create a work item, I'd prefer a comment that shows where it'd be used rather than in the method signature. When I read, "Should use isCiMode parameter", my first thought is, "where? For what method invocation?" When fulfilling that usage, it's at that time we pick if the value should be coming from a parameter, a class member, or somewhere else entirely. Putting it in a parameter can limit someone's (or AI's) creative thinking with an expectation that's not real.

eng/tools/ToolMetadataExporter/src/Program.cs
```diff
@@ -89,7 +88,7 @@ private static void ConfigureAzureServices(IServiceCollection services)
 
             return credential;
         });
-        services.AddSingleton<ICslQueryProvider>(sp =>
+        services.AddSingleton(sp =>
```
vukelich: Subjective: I prefer the type arguments. It makes the type of the delegate clearer to me, especially when the delegate's return value is an expression, like here with `return KustoClientFactory.CreateCslQueryProvider(connectionStringBuilder);`.

I'd guess this line change is due to the analyzer that says something like "name can be simplified". I often dislike that analyzer. Explicit typing improves my confidence when I review code.

tools/Azure.Mcp.Tools.Aks/src/Services/AksService.cs
```diff
@@ -18,12 +16,10 @@ namespace Azure.Mcp.Tools.Aks.Services;
 public sealed class AksService(
     ISubscriptionService subscriptionService,
     ITenantService tenantService,
-    ICacheService cacheService,
-    ILogger<AksService> logger) : BaseAzureResourceService(subscriptionService, tenantService), IAksService
```
vukelich: I'm confused why logging was never used in this type. Seems like there are plenty of helpful places for at least debug-level logging. It's logging I like putting in places where there's a branch in code, like early `return`s, due to externally queried data.

Fine to remove for this PR but concerning nonetheless. My mind wanders on lurking issues that lack vital troubleshooting details.

tools/Azure.Mcp.Tools.AppConfig/src/Models/KeyValueSetting.cs
```diff
@@ -3,15 +3,13 @@
...
-using ETag = Microsoft.Mcp.Core.Models.ETag;
-
 public class KeyValueSetting
 {
     public string Key { get; set; } = string.Empty;
     public string Value { get; set; } = string.Empty;
     public string Label { get; set; } = string.Empty;
     public string ContentType { get; set; } = string.Empty;
-    public ETag ETag { get; set; } = new();
+    public Microsoft.Mcp.Core.Models.ETag ETag { get; set; } = new();
```
vukelich, line 12: I'll let this fully qualified type slide this time. It's the only one in the file, but I'm letting you know I saw it 👀

vukelich: lol very next file removed the statement `using ETag = Microsoft.Mcp.Core.Models.ETag;`

tools/Azure.Mcp.Tools.AppConfig/src/Services/AppConfigService.cs
```diff
@@ -8,20 +8,16 @@
...
 using Microsoft.Mcp.Core.Models.Identity;
 using Microsoft.Mcp.Core.Options;
 using Microsoft.Mcp.Core.Services.Azure.Authentication;
 
 namespace Azure.Mcp.Tools.AppConfig.Services;
 
-using ETag = Microsoft.Mcp.Core.Models.ETag;
-
-public sealed class AppConfigService(ISubscriptionService subscriptionService, ITenantService tenantService, ILogger<AppConfigService> logger, IHttpClientFactory httpClientFactory)
+public sealed class AppConfigService(ISubscriptionService subscriptionService, ITenantService tenantService, IHttpClientFactory httpClientFactory)
```
vukelich: Mentioned before in a different comment, but this is another case where I really like parameters on separate lines. GitHub's UX doesn't make clear that you removed the `ILogger` parameter because UX isn't clever enough to see the after-line18 as before-line21.  If the class default constructor parameters were on separate lines. I'd expect the delta and diff visualization to be more aligned with the semantic change to this file.

tools/Azure.Mcp.Tools.AppService/src/Models/DiagnosisResult.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.Expand commentComment on line R1Resolved
```
vukelich, line 1: Lol that text. How'd that get there?

tools/Azure.Mcp.Tools.Extension/src/ExtensionSetup.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Mcp.Core.Areas.Server.Options;
 using Microsoft.Mcp.Core.Commands;
 using Microsoft.Mcp.Core.Extensions;
+using Microsoft.Mcp.Core.Services.ProcessExecution;
```
vukelich, line 11: How'd this become needed?

tools/Azure.Mcp.Tools.Functions/src/Services/Helpers/GitHubUrlValidator.cs
```diff
@@ -163,7 +163,7 @@ public static bool IsValidRepositoryUrl(string? url)
     /// Enforces the limit in bytes and then decodes to UTF-8 string.
     /// </summary>
     /// <param name="content">The HTTP content to read.</param>
-    /// <param name="maxSizeBytes">Maximum allowed size in bytes (must be <= int.MaxValue).</param>
+    /// <param name="maxSizeBytes">Maximum allowed size in bytes (must be &lt;= int.MaxValue).</param>
```
vukelich: `<see cref="int.MaxValue />` or something like that.

tools/Azure.Mcp.Tools.KeyVault/src/Services/IKeyVaultService.cs
```diff
@@ -137,9 +143,11 @@ Task<List<string>> ListCertificates(
     /// List all keys in a Key Vault.
     /// </summary>
     /// <param name="vaultName">Name of the Key Vault.</param>
+    /// <param name="includeManagedKeys">Whether to include managed keys in the results.</param>
     /// <param name="subscriptionId">Subscription ID containing the Key Vault.</param>
     /// <param name="tenantId">Optional tenant ID for cross-tenant operations.</param>
     /// <param name="retryPolicy">Optional retry policy for the operation.</param>
+    /// <param name="cancellationToken">A cancellation token</param>
```
vukelich: This file wavers between ending `param` node text with periods and not. Idk if you care to fix. All the additional `cancellationToken` nodes added are consistent to each other but not necessarily to the other params of the same method.

vukelich: I do like `A cancellation token` with the leading article. I'm probably just used to it across some API surface that I used a lot. Quickly looking at some .NET types, there's no consistency on the text used, including if methods say `cancellation token` or `<see cref="CancellationToken"/>` or allude to `CancellationToken.None`.

tools/Azure.Mcp.Tools.Monitor/src/Tools/Instrumentation/OrchestratorTool.cs
```diff
@@ -130,7 +130,9 @@ public string Start(string workspacePath)
         });
     }
 
+#pragma warning disable IDE0060 // Remove unused parameter
```
vukelich, line 133: Why the disable rather than remove `string completionNote`?

vukelich (changes requested):
Some just questions. Some subjective. A couple actual issues.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
vukelich: Update: the language used is inconsistent in the repo. in .NET, and across lines added in this PR. Idk what to suggest or say.

core/Azure.Mcp.Core/src/Services/Azure/AzureServiceCollectionExtensions.cs
alzimmermsft: `AuthenticationServiceCollectionExtensions.AddAzureTokenCredentialProvider` doesn't exist as an API anymore, so unless this was something external I missed, this doesn't exist anymore.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
alzimmermsft: I may have left this to an agent to decide what to do. I'll work on making this consistent across what we own.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
alzimmermsft: The analyzers and IDEs get a bit weird on preprocessor blocks on properly conveying when code is still used. This was making it so they do understand this code does get used. The other option here was to have `#pragma` statements to disable analyzer rules that flag the code as unused.

No strong opinion on which to use here.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/AccessTokenHandlerTests.cs
alzimmermsft: Yeah, that class still was in `Azure.Mcp.Core` even though it had been moved into the `Microsoft.Mcp.Core` namespace.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/OptionSchemaGenerator.cs
alzimmermsft: Will add this back sans the CREF part to `UnconditionalSuppressMessageAttribute`, as that class is internal and outside the reachability for this spot.

tools/Azure.Mcp.Tools.Monitor/src/Tools/Instrumentation/OrchestratorTool.cs
alzimmermsft: This ties back to the only usage of the required tool parameter `completion-note`. This needs a deeper follow up than just deleting it (in my opinion).

tools/Azure.Mcp.Tools.Aks/src/Services/AksService.cs
alzimmermsft: This likely stems from a design I changed a while back where `*Service` implementation were try-catching `Exception` and logging in the same way the tool calling the API would. Ex:

**Tool**

```dotnet
try
{
    service.Something();
}
catch (Exception ex)
{
    _logger.LogError(ex, "Something failed");
    HandleException(context, ex);
}
```

**Service**

```dotnet
public void Something()
{
    try
    {
        CallTheService();
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Something failed");
        throw ex;
    }
}
```

If the service method was only used by a single tool and the error logging matched exactly, I changed the design to stop try-catch and logging in the service, as it was redundant. So, the design became:

**Tool**

```dotnet
try
{
    service.Something();
}
catch (Exception ex)
{
    _logger.LogError(ex, "Something failed");
    HandleException(context, ex);
}
```

**Service**

```dotnet
public void Something()
{
    CallTheService();
}
```

eng/tools/ToolMetadataExporter/src/Program.cs
alzimmermsft: This might have just been me doing that. That analyzer is still an informational message and not a warning. Can revert it

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
alzimmermsft: In this PR, I'd say let's leave it as-is, as it documents current behaviors and won't introduce scope creep. But I also agree with this comment, and I'm in the opinion that something this meaningful should require explicit interaction by the caller to indicate the table to query and we provide some canned `const`s to help. This has been a spot where we've seen a few bugs as the default was used but the needed ARG table was something else (hence why this is a parameter in the first place).

core/Microsoft.Mcp.Core/src/Commands/ICommandFactory.cs
alzimmermsft: Doc comments

core/Microsoft.Mcp.Core/src/Services/Azure/Authentication/IAzureTokenCredentialProvider.cs
alzimmermsft: This is in the `Microsoft.Mcp.Core` package where `ITenantService` is in `Azure.Mcp.Core`, so we don't have access to `ITenantService` here. This is a spot we need to follow-up on where `Microsoft.Mcp.Core` still has some weird spots where it owns Azure-related code even though it isn't meant to.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ServiceCollectionExtensionsSerializedTests.cs
alzimmermsft: Yeah, though need to use a slightly different alias as we also have an `Options` namespace that conflicts with this recommendation. Going to use `ExtensionOptions`.

core/Microsoft.Mcp.Core/src/Services/ProcessExecution/ExternalProcessService.cs
alzimmermsft: The compiler is flagging this as an ambiguous reference as there is the `Process.Start()` instance method and multiple `Process.Start(...)` static methods.

tools/Azure.Mcp.Tools.Extension/src/ExtensionSetup.cs
alzimmermsft: Doc references

tools/Azure.Mcp.Tools.AppService/src/Models/DiagnosisResult.cs
alzimmermsft: I assume either a copy and paste error or something agentic, or a combination of both

tools/Azure.Mcp.Tools.AppConfig/src/Models/KeyValueSetting.cs
alzimmermsft: This is a spot where we have another namespace and class name collision with `Azure.ETag`.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Generated/MicrosoftClientModelTestFrameworkModelFactory.cs
alzimmermsft: Something does seem off with the generated code here, and I'm not sure why.

The issue is that the crefs aren't pointing to real references, they're off. Ex, `TestProxy.TestProxyStartInformation` should just be `TestProxyStartInformation` as it's trying to reference that class in the `Models` namespace.

For this PR I'd prefer this just ignore those warnings for now and circle back to what's going wrong with codegen to have invalid references to other generated code.

core/Microsoft.Mcp.Core/src/Extensions/OpenTelemetryExtensions.cs
alzimmermsft: I've reverted this change.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
alzimmermsft: Reverted this change.

conniey:
Based on the description of the PR, I like the enforcement of these rules.. however, I'm having a hard time finding the file you changed to enforce these rules because it also includes the fixes.  Can you point them out? :(

Directory.Build.props
```diff
@@ -20,6 +20,8 @@
       See: https://learn.microsoft.com/nuget/reference/errors-and-warnings/nu1901-nu1904
     -->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902</WarningsNotAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
LarryOsterman, line 27: @conniey I think this is the magic change to add style enforcement. @alzimmermsft can confirm though.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -7,7 +7,9 @@
 using Azure.Core.Pipeline;
 using Azure.Mcp.Core.Services.Azure.Tenant;
 using Azure.ResourceManager;
+#pragma warning disable IDE0005 // using isn't used in release builds.
```
LarryOsterman: Does it make sense to wrap this in a `#if debug` conditional rather than unconditionally disabling the warning?

LarryOsterman (approved):
LGTM but it's going to make life unpleasant for awhile.

Directory.Build.props
alzimmermsft: Yeah, this enables checks to run as part of the build, rather than pushing everything off to `dotnet format`.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
alzimmermsft: That makes a lot more sense to me, will make that change.

LarryOsterman (approved):
Oh my, that most recent commit made this even bigger.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -4,7 +4,9 @@
 using Azure.Core;
 using Azure.Mcp.Core.Services.Azure.Helpers;
 using Azure.ResourceManager;
+#if DEBUG
 using Microsoft.Mcp.Core.Helpers;
```
conniey: Fully-qualifying the one statement that uses this would make this easier to digest. I was expecting a slew of usages since we imported the statement.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Helpers/MockMcpClientBuilder.cs
```diff
@@ -1,14 +1,13 @@
-#pragma warning disable MCP9003 // Obsolete RequestContext constructor - migrating during Phase 1
-#pragma warning disable MCP9005 // Deprecated Sampling/Logging APIs - backward compat during Phase 1
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#pragma warning disable MCP9003 // Obsolete RequestContext constructor - migrating during Phase 1
```
conniey, line 4: For all these, do we have any intention of fixing these? (i.e no longer using the requestcontext ctor?)

alzimmermsft: I believe @g2vinay is owning this work, but if he isn't I'm happy to follow up on it.

conniey (approved):

conniey (approved):

PR #3184, opened 2026-07-30
Stop the npm wrapper's platform-package auto-install from writing to the caller's package.json

### LINKED ISSUE #3519: [BUG] npm wrapper writes @azure/mcp-<platform>-<arch> into an unrelated project's package.json and lockfile
### Describe the bug

`@azure/mcp-darwin-arm64` appeared in the `dependencies` of a Next.js app of mine that has nothing to do with Azure, and in its lockfile. I hadn't installed it. It propagated to CI and to every other developer on the repo through the lockfile.

The source is the npm wrapper's auto-install fallback in `eng/npm/wrapper/index.js#L48-L52`. When the wrapper can't `require` its platform package it shells out to a bare `npm install`:

```js
execSync(`npm install ${platformPackageName}@${packageVersion}`, {
  stdio: ['inherit', 'inherit', 'pipe'],
  timeout: 60000
})
```

There is no `--no-save` and no `--prefix`, and `cwd` is inherited from whoever spawned the wrapper. For an MCP client, that's the user's own project directory, so npm walks up, finds their `package.json`, and saves the platform package into it. `--no-save` is only applied on the second attempt (L58), which never runs because the first one succeeds.

Two smaller problems in the same path:

- The install inherits **stdout**. Under `server start` that's the JSON-RPC channel, so npm output written there can corrupt the protocol stream.
- The troubleshooting output doesn't print the Node version, which is the likeliest cause (below) and otherwise invisible.

### Why the fallback fires at all

npm silently skips `optionalDependencies` whose `engines` don't match the running Node, and `3.0.0-beta.27` raised `engines.node` to `>=22.0.0` on both the wrapper and the platform packages (`eng/scripts/Pack-Npm.ps1` L102 and L156). On Node 20 the platform package is omitted from the install with no error:

```console
$ node -v && npm i @azure/mcp@3.0.0-beta.30 --ignore-scripts && ls node_modules/@azure/
v20.19.4
mcp

$ node -v && npm i @azure/mcp@3.0.0-beta.30 --ignore-scripts && ls node_modules/@azure/
v22.22.0
mcp  mcp-darwin-arm64
```

The wrapper's own engines mismatch is only an `EBADENGINE` warning, since it's a direct dependency, so the install looks clean and the failure surfaces at runtime. Whether `>=22.0.0` is the right floor is a separate question I haven't touched.

### Expected behavior

Running the Azure MCP server should never modify the `package.json` or lockfile of the project it happens to be launched from, and should never write to stdout during `server start`.

### Actual behavior

`@azure/mcp-<platform>-<arch>` is added to the caller's `dependencies` and lockfile:

```diff
-{"name":"ctrl","version":"1.0.0","dependencies":{"@azure/mcp":"^3.0.0-beta.30"}}
+{"name":"ctrl","version":"1.0.0","dependencies":{"@azure/mcp":"^3.0.0-beta.30","@azure/mcp-darwin-arm64":"^3.0.0-beta.30"}}
```

### Reproduction Steps

1. On Node 20, in any project directory with a `package.json`:
   `npm i @azure/mcp@3.0.0-beta.30`
2. Confirm the platform package was skipped: `ls node_modules/@azure/` shows only `mcp`.
3. Run the wrapper from that directory: `node node_modules/@azure/mcp/index.js --version`
4. `git diff package.json package-lock.json` — both now contain `@azure/mcp-darwin-arm64`.

`--version` is just the smallest command that triggers it. The same fallback runs on `server start`, which is how users actually hit it — and there npm's output lands on stdout, i.e. on the JSON-RPC channel itself.

### Environment

* macOS (Darwin 25.6.0) arm64, npm 11.6.1, Node 20.19.4
* `@azure/mcp@3.0.0-beta.30`

### Fix

PR #3184 is open with a fix: install into a private `.platform` directory inside the wrapper package with `--no-save` and an explicit `--prefix`, resolve the platform package from that path, stop inheriting stdout, and add the Node version to the troubleshooting output.

Two notes for whoever triages this:

- CODEOWNERS routes `%packages-npx` to @KarishmaGhiya and @hallipr, but that's a `ServiceLabel`, which only applies to issues. There is no `PRLabel:` entry covering `/eng/`, so a PR touching the npm wrapper gets no path label and no owner, which may be why #3184 hasn't found its way to anyone yet. A `PRLabel: %packages-npx` entry for `/eng/npm/` would route future wrapper PRs automatically; happy to send that as a separate one-liner if it'd be useful.
- The PR has no test coverage because `eng/npm/wrapper/index.js` has none today. Glad to add some if you have a preferred location.

djdmbrwsk:
### Problem

When the wrapper can't `require` its platform package, it auto-installs it with a bare `npm install` ([`eng/npm/wrapper/index.js#L48-L52`](https://github.com/microsoft/mcp/blob/main/eng/npm/wrapper/index.js#L48-L52)):

```js
execSync(`npm install ${platformPackageName}@${packageVersion}`, {
  stdio: ['inherit', 'inherit', 'pipe'],
  timeout: 60000
})
```

There's no `--no-save` and no `--prefix`, and `cwd` is inherited from whoever spawned the wrapper. For an MCP client that's an unrelated user project, so npm walks up from there, finds the user's `package.json`, and **saves `@azure/mcp-<platform>-<arch>` into their dependencies and lockfile**. `--no-save` is only applied on the second attempt (L58), which never runs because the first one succeeds.

I hit this as an unexplained `@azure/mcp-darwin-arm64` entry in the `dependencies` of a Next.js app with no Azure involvement — it then propagates to CI and to every other developer on the repo via the lockfile.

### Why the fallback fires at all

npm skips `optionalDependencies` whose `engines` don't match the running Node, and `3.0.0-beta.27` raised `engines.node` to `>=22.0.0` on the wrapper and the platform packages. On Node 20 the platform package is silently omitted from the install:

```console
$ node -v && npm i @azure/mcp@3.0.0-beta.30 --ignore-scripts && ls node_modules/@azure/
v20.19.4
mcp

$ node -v && npm i @azure/mcp@3.0.0-beta.30 --ignore-scripts && ls node_modules/@azure/
v22.22.0
mcp  mcp-darwin-arm64
```

With npm, the wrapper's postinstall check then fails the install ("Missing required package"), so npm/npx users on Node 20 get an install error before they ever reach this fallback. The fallback runs when install scripts are skipped: pnpm and bun by default, or npm with `--ignore-scripts` as above, which is how I hit it. Whether the postinstall should warn instead and let the fallback handle Node <22, and whether `>=22.0.0` is the intended floor, are separate questions I've left to you.

### Fix

- Install into a private `.platform` directory inside the wrapper package, with `--no-save` and an explicit `--prefix`, and resolve the platform package from that path. The directory intentionally has no `package.json`: pointing npm at the wrapper's own manifest is a no-op, because the platform package is declared there as an `optionalDependency` that npm is already skipping.
- When `.platform` isn't writable (root-owned global installs, read-only filesystems), fall back to a per-user cache directory: `~/Library/Caches`, `%LOCALAPPDATA%`, or `$XDG_CACHE_HOME`/`~/.cache`, under `microsoft-mcp/<package>/<version>`.
- Reuse an earlier install from either directory instead of running npm on every start.
- Serialize concurrent installs with a lock file in the install directory, so several MCP clients starting at once install once instead of racing.
- Stop inheriting **stdout** for the install. Under `server start` that's the JSON-RPC channel, so npm output written there can corrupt the protocol stream. The wrapper's own messages already go to stderr.
- Add the Node version to the troubleshooting output, since the engines skip is the likeliest cause and is otherwise invisible.

### Verification

macOS arm64, npm 11.6.1, against `@azure/mcp@3.0.0-beta.30`.

**Before** (Node 20, wrapper run from a project directory):

```diff
-{"name":"ctrl","version":"1.0.0","dependencies":{"@azure/mcp":"^3.0.0-beta.30"}}
+{"name":"ctrl","version":"1.0.0","dependencies":{"@azure/mcp":"^3.0.0-beta.30","@azure/mcp-darwin-arm64":"^3.0.0-beta.30"}}
```

**After** (same setup, patched wrapper):

```console
$ node node_modules/@azure/mcp/index.js --version > stdout.txt 2> stderr.txt; echo "exit=$?"
exit=0

$ cat stdout.txt          # clean - no npm output on the protocol channel
3.0.0-beta.30+bd96686782eb9b62c47662b941de4c4b9eb0aaf7

$ cat stderr.txt
Installing missing platform package: @azure/mcp-darwin-arm64
✅ Successfully installed and loaded @azure/mcp-darwin-arm64

$ ls node_modules/@azure/mcp/.platform/node_modules/@azure/
mcp-darwin-arm64

$ diff package.json.before package.json && diff package-lock.json.before package-lock.json
# no output - both unchanged
```

**Happy path** (Node 22, platform package present): unchanged — the server runs, the fallback never fires, and no `.platform` directory is created.

### Notes

- Tests: `eng/npm/test/wrapper.test.js` (`node:test`, no new dependencies), run with `node --test eng/npm/test/wrapper.test.js`. They live outside `eng/npm/wrapper/` because Pack-Npm.ps1 publishes that whole directory. Nothing in CI runs them yet.
- Fixes #3519.

### DIFF
--- a/eng/npm/wrapper/index.js
+++ b/eng/npm/wrapper/index.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 
+const fs = require('fs')
 const os = require('os')
+const path = require('path')
 const packageJson = require('./package.json')
 
 // Check if DEBUG environment variable is set
@@ -44,21 +46,35 @@ try {
 
     console.error(`Installing missing platform package: ${platformPackageName}`)
 
+    // Install into a private directory owned by this package, never the
+    // caller's. cwd is inherited from whoever spawned us - for an MCP client
+    // that is an unrelated user project - and a bare `npm install` resolves the
+    // nearest package.json from there and saves the platform package into it.
+    //
+    // The directory deliberately has no package.json of its own: pointing npm
+    // at this package's manifest is a no-op, because the platform package is
+    // declared there as an optionalDependency that npm is already skipping
+    // (that skip is usually why we are in this code path at all).
+    const installDir = path.join(__dirname, '.platform')
+    fs.mkdirSync(installDir, { recursive: true })
+
+    // stdout is not inherited: for `server start` it carries the JSON-RPC
+    // protocol stream, and npm output written there would corrupt it.
+    const installOptions = {
+      cwd: installDir,
+      stdio: ['ignore', 'pipe', 'pipe'],
+      timeout: 60000 // 60 second timeout
+    }
+    const installCommand = `npm install ${platformPackageName}@${packageVersion} --no-save --no-audit --no-fund --prefix "${installDir}"`
+
     // Try to install the platform package
     try {
-      execSync(`npm install ${platformPackageName}@${packageVersion}`, {
-        stdio: ['inherit', 'inherit', 'pipe'], // Only pipe stderr to capture errors
-        timeout: 60000 // 60 second timeout
-      })
+      execSync(installCommand, installOptions)
     } catch (npmErr) {
-      // If npm install fails, try with --no-save and different install strategies
+      // If npm install fails, try again against the registry
       debugLog(`npm install failed, trying alternative installation methods: ${npmErr.message}`)
 
-      // Try with --no-save and --prefer-online
-      execSync(`npm install ${platformPackageName}@${packageVersion} --no-save --prefer-online`, {
-        stdio: ['inherit', 'inherit', 'pipe'],
-        timeout: 60000
-      })
+      execSync(`${installCommand} --prefer-online`, installOptions)
     }
 
     // Clear module cache and try to require again after installation
@@ -68,7 +84,7 @@ try {
       }
     })
 
-    platformPackage = require(platformPackageName)
+    platformPackage = require(path.join(installDir, 'node_modules', platformPackageName))
 
     console.error(`✅ Successfully installed and loaded ${platformPackageName}`)
 
@@ -87,8 +103,11 @@ try {
     console.error(`   npx -y ${packageName}@latest --version`)
     console.error(`\n4. Manually install the platform package to check compatibility:`)
     console.error(`   npm install ${platformPackageName}@latest`)
-    console.error(`\n5. Check your internet connection and try again`)
-    console.error(`\n6. If the issue persists, please report it at:`)
+    console.error(`\n5. Check your Node version. npm silently skips the optional platform`)
+    console.error(`   package when Node does not satisfy engines (${(packageJson.engines || {}).node}).`)
+    console.error(`   You are running ${process.version}.`)
+    console.error(`\n6. Check your internet connection and try again`)
+    console.error(`\n7. If the issue persists, please report it at:`)
     console.error(`   https://github.com/microsoft/mcp/issues`)
     console.error(`\nOriginal error: ${err.message}`)
     console.error(`Install error: ${installErr.message}`)
--- a/eng/npm/wrapper/index.js
+++ b/eng/npm/wrapper/index.js
@@ -48,10 +48,19 @@ const platformPackageName = `${packageName}-${platform}-${arch}`
 // package's manifest is a no-op, because the platform package is declared
 // there as an optionalDependency that npm is already skipping (that skip is
 // usually why we are in this code path at all).
-const installDirs = [
-  path.join(__dirname, '.platform'),
-  path.join(userCacheDir(), 'microsoft-mcp', platformPackageName, packageVersion)
-]
+//
+// This only runs once the normal require has failed, so a working install never
+// touches the home directory: os.homedir() throws when HOME is unset and the
+// user has no passwd entry.
+function getInstallDirs() {
+  const dirs = [path.join(__dirname, '.platform')]
+  try {
+    dirs.push(path.join(userCacheDir(), 'microsoft-mcp', platformPackageName, packageVersion))
+  } catch (err) {
+    debugLog(`No per-user cache directory available: ${err.message}`)
+  }
+  return dirs
+}
 
 function userCacheDir() {
   if (platform === 'win32') {
@@ -60,7 +69,10 @@ function userCacheDir() {
   if (platform === 'darwin') {
     return path.join(os.homedir(), 'Library', 'Caches')
   }
-  return process.env.XDG_CACHE_HOME || path.join(os.homedir(), '.cache')
+  // The XDG spec says to ignore a relative XDG_CACHE_HOME. Honoring one would
+  // put the cache under the caller's working directory.
+  const xdgCacheHome = process.env.XDG_CACHE_HOME
+  return xdgCacheHome && path.isAbsolute(xdgCacheHome) ? xdgCacheHome : path.join(os.homedir(), '.cache')
 }
 
 function requireFromInstallDir(installDir) {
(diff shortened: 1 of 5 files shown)


djdmbrwsk:
@microsoft-github-policy-service agree

djdmbrwsk:
Rebased onto current `main` and moved the changelog entry to a `changelog-entries/*.yml` file to match the current convention. CI is green / re-running against `main`.

@KarishmaGhiya @hallipr — CODEOWNERS lists you as the owners for `%packages-npx`, so flagging this one directly. It's a small fix to `eng/npm/wrapper/index.js`: the platform-package auto-install runs a bare `npm install` with an inherited `cwd`, so it saves `@azure/mcp-<platform>-<arch>` into the calling project's `package.json` and lockfile. I hit it as a mystery dependency in an unrelated Next.js app. Filed #3519 with the full repro.

@joshfree — could you add the `packages-npx` label? PRs under `/eng/` don't match any `PRLabel:` entry in CODEOWNERS, so this one was never labeled or routed to anyone.

eng/npm/wrapper/index.js
```diff
@@ -44,21 +46,35 @@ try {
...
+    // caller's. cwd is inherited from whoever spawned us - for an MCP client
+    // that is an unrelated user project - and a bare `npm install` resolves the
+    // nearest package.json from there and saves the platform package into it.
+    //
+    // The directory deliberately has no package.json of its own: pointing npm
+    // at this package's manifest is a no-op, because the platform package is
+    // declared there as an optionalDependency that npm is already skipping
+    // (that skip is usually why we are in this code path at all).
+    const installDir = path.join(__dirname, '.platform')
+    fs.mkdirSync(installDir, { recursive: true })
```
KarishmaGhiya, line 59: `installDir`  lives inside the wrapper package's own install location. That's fine for local/npx-cached installs, but breaks two common cases:
• Global installs ( `npm install -g @azure/mcp` ) where the global  `node_modules`  is root-owned (e.g. system Node on Linux) — the previous  `cwd` -based approach at least usually landed in a writable directory; this one may now hard-fail with  `EACCES` .
• Read-only container filesystems, which are common for hardened deployments.

Suggest catching the permission failure and falling back to a per-user cache directory instead of failing outright:
```
function resolveInstallDir() {
  const primary = path.join(__dirname, '.platform')
  try {
    fs.mkdirSync(primary, { recursive: true })
    fs.accessSync(primary, fs.constants.W_OK)
    return primary
  } catch {
    const fallback = path.join(os.homedir(), '.azure-mcp', 'platform', packageVersion)
    fs.mkdirSync(fallback, { recursive: true })
    return fallback
  }
}
```
Worth a debug log noting which path was used, so support/troubleshooting output can surface it.

KarishmaGhiya:
@djdmbrwsk 
This file has no tests today, but you could add some since you offered earlier— this PR is a good place to start since it's changing the exact logic that had the regression. Suggest a minimal test that mocks  `child_process.execSync`  and  `fs` / `os`  to assert:
1. The install command includes  --no-save ,  --no-audit ,  --no-fund , and  --prefix <installDir>  (regression guard for the original bug).
2.  cwd / stdio[1]  (stdout) is never inherited from the parent process (regression guard for the JSON-RPC corruption issue).
3. When  installDir/node_modules/<platformPackage>  already exists,  execSync  is not called at all (guards comment #1 above once implemented).
Even a lightweight  node:test / assert  based spec without new dependencies would close this gap and prevent silent regressions on future edits to this file.

eng/npm/wrapper/index.js
```diff
@@ -44,21 +46,35 @@ try {
...
+    // Install into a private directory owned by this package, never the
+    // caller's. cwd is inherited from whoever spawned us - for an MCP client
+    // that is an unrelated user project - and a bare `npm install` resolves the
+    // nearest package.json from there and saves the platform package into it.
+    //
+    // The directory deliberately has no package.json of its own: pointing npm
+    // at this package's manifest is a no-op, because the platform package is
+    // declared there as an optionalDependency that npm is already skipping
+    // (that skip is usually why we are in this code path at all).
+    const installDir = path.join(__dirname, '.platform')
```
KarishmaGhiya, line 58: Nice fix for the caller's- `package.json`  mutation. One gap: this always shells out to  `npm install`  on every invocation where the top-level  `require(platformPackageName)` fails — which, on Node <22, is every run, since the platform package is permanently excluded by `engines` . Before this PR, the installed package ended up in the caller's own  `node_modules`, so plain node resolution found it on subsequent runs without reinstalling; now that side effect is gone (correctly), but we lost that "caching" for free. Suggest checking for an existing install before invoking npm:
```
const installedPath = path.join(installDir, 'node_modules', platformPackageName)
if (fs.existsSync(installedPath)) {
  debugLog(`Found cached platform package at ${installedPath}`)
  platformPackage = require(installedPath)
} else {
  fs.mkdirSync(installDir, { recursive: true })
  // ...existing install logic...
}
```
This avoids a network round-trip + npm spawn on every single server start for anyone on Node <22, which otherwise makes every launch slower and dependent on registry availability.

eng/npm/wrapper/index.js
djdmbrwsk: Done in 1f279c05. If `.platform` isn't writable, the install goes to a per-user cache directory (`~/Library/Caches` on macOS, `%LOCALAPPDATA%` on Windows, `$XDG_CACHE_HOME` or `~/.cache` elsewhere) under `microsoft-mcp/<package>/<version>`. It isn't `~/.azure-mcp` because the Fabric and Template servers ship the same wrapper. With `DEBUG=true` the wrapper logs which directory it used.

I left out the OS temp directory as a last resort: on a shared machine another user could plant a package there for the wrapper to load.

eng/npm/wrapper/index.js
djdmbrwsk: Good catch, fixed in 1f279c05: the wrapper now checks both install directories before running npm. On Node 20 with real npm, the second start went from about 2.3s to 0.1s.

db132760 also adds a lock, since every launch now shares one install directory. With real npm, 4 simultaneous first starts used to fail intermittently (exit codes like `0 0 1 1`); now they all start, with one install.

djdmbrwsk:
@KarishmaGhiya Added `eng/npm/test/wrapper.test.js`: 12 tests using `node:test`, no new dependencies. It covers the three checks you listed (install flags and `--prefix`, stdout not inherited, no npm call when already installed), plus concurrent launches, a missing home directory, the per-user cache fallback, and the failure paths.

Because the wrapper does its work as soon as it loads, the tests run a copy of it as a child process with a fake `npm` on `PATH`, rather than mocking `execSync`. They live outside `eng/npm/wrapper/` because Pack-Npm.ps1 publishes that whole directory. Nothing in CI runs them yet (`node --test eng/npm/test/wrapper.test.js`). Which pipeline would you like them in?

I also updated the PR description. While testing I found that with npm, the wrapper's postinstall check fails a Node 20 install before this fallback can run, so the fallback mainly covers installs that skip scripts (pnpm, bun, `--ignore-scripts`). Whether the postinstall should warn instead is your call; I've left it unchanged.

KarishmaGhiya:
@djdmbrwsk Thanks for adding these — 12 tests covering the regression scenarios is great coverage for a dependency-free spec.

To answer where they should run: `eng/pipelines/templates/jobs/analyze.yml` (the `Analyze` job in the `Build` stage)

`analyze.yml` is a better fit because:
- It already runs on every PR unconditionally (no `PublishTarget`/stage gating like `SignAndPack`/`pack-npm.yml` has, which is skipped for normal PR validation since `pullrequest.yml` doesn't set `PublishTarget`).
- It already installs Node.js 24 via `UseNode@1`.
- It's already the home for other lightweight, no-new-dependency checks (`Analyze-Code.ps1`, AOT compat, README validation) — a `node --test` step fits the same pattern.

Something like this, added after the "Run source analysis" step:

```yaml
- pwsh: node --test eng/npm/test/wrapper.test.js
  displayName: "Run npm wrapper tests"
  workingDirectory: $(Build.SourcesDirectory)
```

Could you wire it in there? Also happy to have you fold in the Node-20-postinstall-check finding as a follow-up issue rather than blocking this PR on it.

Also, please add a changelog.yml for this PR. 

Will be ready to merge once you have these updates (add above step in analyze.yml and add changelog), thanks!

KarishmaGhiya (approved):

msalaman (approved):

msalaman (approved):

PR #3228, opened 2026-08-07
Add progress notifications during EventHubs namespace update LRO wait

LarryOsterman:
The EventHubs namespace update tool blocks silently while ARM completes the SKU/capacity update, which can genuinely take several minutes. With no MCP activity emitted during that wait, MCP clients (including the Vally eval harness) can conclude the session is idle and time out even though the tool is still legitimately working.

- Add an optional IProgress<string> parameter to BaseAzureService.WaitForLroCompletionAsync (both overloads). When supplied, the LRO poll loop reports a heartbeat status message every 15 seconds while waiting, instead of delegating to the SDK's opaque WaitForCompletionAsync/WaitForCompletionResponseAsync. This is purely additive - existing callers that don't pass progress are unaffected.
- Thread an optional progress parameter through IEventHubsService.CreateOrUpdateNamespaceAsync / EventHubsService.CreateOrUpdateNamespaceAsync.
- NamespaceUpdateCommand now builds an IProgress<string> that forwards to context.McpServer.NotifyProgressAsync (mirroring the existing pattern in InsightsGetCommand), so clients receive periodic progress notifications - resetting client-side inactivity timeouts - while a namespace update is in flight.
- Update NamespaceUpdateCommandTests to account for the new trailing IProgress<string>? parameter on CreateOrUpdateNamespaceAsync.
- Add a changelog entry.

## GitHub issue number?

#3229 

## Pre-merge Checklist
- [ ] Required for All PRs
- [ ] For MCP tool changes:

### DIFF
--- a/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
@@ -409,13 +409,13 @@ public abstract class BaseAzureService
             Response response = await operation.UpdateStatusAsync(cancellationToken).ConfigureAwait(false);
             if (operation.HasCompleted)
             {
-                return operation.GetRawResponse();
+                return response;
             }
 
-            elapsed += pollInterval;
             progress.Report($"Still waiting for the operation to complete ({elapsed:mm\\:ss} elapsed)...");
 
             await Task.Delay(pollInterval, cancellationToken).ConfigureAwait(false);
+            elapsed += pollInterval;
         }
     }
 }
--- a/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
@@ -397,54 +397,48 @@ public abstract class BaseAzureService
     /// </remarks>
     private static async Task<Response> WaitForLroCompletionInternalAsync(Operation operation, IProgress<string>? progress, CancellationToken cancellationToken)
     {
-        // Base case: No progress, no default poll interval. Just wait for the operation to complete.
-        if (progress == null && !s_defaultPollInterval.HasValue)
+        if (s_defaultPollInterval.HasValue)
+        {
+            // Loop polling the async operation status at a fixed interval until it completes.
+            // This is used during playback testing to avoid the Azure SDK's default polling delay.
+            while (true)
+            {
+                Response response = await operation.UpdateStatusAsync(cancellationToken).ConfigureAwait(false);
+                if (operation.HasCompleted)
+                {
+                    return response;
+                }
+
+                await Task.Delay(s_defaultPollInterval.Value, cancellationToken).ConfigureAwait(false);
+            }
+        }
+
+        if (progress is null)
         {
             return await operation.WaitForCompletionResponseAsync(cancellationToken);
         }
         else
-            if (s_defaultPollInterval.HasValue)
+        {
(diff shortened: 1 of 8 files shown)


core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -297,6 +297,14 @@ protected static void ValidateRequiredParameters(params (string name, string? va
         }
     }
 
+    /// <summary>
+    /// Interval at which <see cref="WaitForLroCompletionAsync{T}(Operation{T}, IProgress{string}, CancellationToken)"/> polls
+    /// the operation and reports progress. This keeps the caller (and any MCP client watching for session activity)
+    /// informed while a genuinely long-running Azure operation is still in progress, instead of appearing idle for the
+    /// entire duration of the wait.
+    /// </summary>
+    private static readonly TimeSpan s_progressPollInterval = TimeSpan.FromSeconds(15);
```
alzimmermsft, line 306: Any reason for progress-based LROs to have a different default polling interval? Why not use the built-in interval used by the library being used?

If we're worried about over reporting, instead time interval when we report to this specific delay. Right now, if the operation takes 5 seconds were forcing in an additional 10 seconds of delay until we complete the operation here.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -351,4 +398,24 @@ private static async Task<Response> WaitForLroCompletionInternalAsync(Operation
...
+
+        while (true)
+        {
+            Response response = await operation.UpdateStatusAsync(cancellationToken).ConfigureAwait(false);
+            if (operation.HasCompleted)
+            {
+                return response;
+            }
+
+            progress.Report($"Still waiting for the operation to complete ({elapsed:mm\\:ss} elapsed)...");
```
alzimmermsft: Given the possibility that the initial operation and polling operations won't always be the exact `TimeSpan` specified, should we be reporting a diff on the real wall clock? Ex, at the start of this we capture the current time and every call here is `current - begin`.

For example, could see the server requesting backoff during these operations which would change the 15 second polling delay to something larger as it'd be 15 seconds + backoff delay until we report again.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
LarryOsterman: Progress based LROs have the same polling interval, I believe. `WaitForLroCompletionInternalAsync` uses `s_defaultPollInterval` if it's configured. `WaitForLroCompletionWithProgressAsync` uses `s_defaultPollInterval if it's configured otherwise it uses `s_progressPollInterval`. Essentially `s_progressPollInterval` is a backstop for when the `s_defaultPollInterval` isn't set.

LarryOsterman: To be 100% clear: `WaitForLroCompletionAsync` is somewhat weird. If you're in playback mode we never delay, otherwise we fall back to `operation.WaitForCompletionResponseAsync` which handles polling on a per SDK basis using the SDK's default strategy.

But in this case we don't have access to the polling strategy so we don't have a great way of providing polling updates to the client. 

I'm open to other suggestions w.r.t. how to minimize the traffic to the server - possibly spawning a separate task which does the progress callback and then cancels the task when the Wait completes?

PR #3288, opened 2026-08-14
Add MCP output size measurement tooling

LarryOsterman:
## Why

MCP clients benefit from understanding the protocol cost of tool discovery and learn-mode responses. This adds a repeatable, opt-in measurement workflow to compare consolidated and namespace server modes.

This tool is written as a test to take advantage of the existing MCP test infrastructure that invokes the MCP tool so that the metrics can be generated based on the actual output of the MCP server.

## Approach

- Adds an opt-in stdio integration test that captures initialize, discovery, top-level learn, and per-command learn responses across both modes.
- Writes structured reports and response artifacts, including decoded top learn descriptions and extracted inner-command schemas.
- Adds PowerShell scripts to summarize JSON reports as console, JSON, and Markdown output, and to run build, measurement, and summarization end to end.

## Validation

Ran the end-to-end measurement script successfully.


eng/tools/McpOutputSizeMeasurer/src/McpOutputSizeMeasurer.cs
```diff
@@ -0,0 +1,405 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
+using Xunit;
+
```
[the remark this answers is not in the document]
LarryOsterman: FWIW, this seems to compile just fine in the CI pipeline.

servers/Azure.Mcp.Server/tests/Azure.Mcp.Server.Tests/Infrastructure/McpOutputSizeTests.cs
```diff
@@ -0,0 +1,405 @@
...
+            measurements.Add(await MeasureModeAsync(mode, Path.GetDirectoryName(reportPath)!));
+        }
+
+        var report = JsonSerializer.Serialize(
+            new
+            {
+                transport = "stdio",
+                threeStepToolDiscovery = ServerTestOptions.IsThreeStepToolDiscoveryEnabled(),
+                generatedAtUtc = DateTimeOffset.UtcNow,
+                reportPath,
```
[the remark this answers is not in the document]
LarryOsterman: This was because of a bad merge, fixed seprately.

servers/Azure.Mcp.Server/tests/Azure.Mcp.Server.Tests/Infrastructure/McpOutputSizeTests.cs
```diff
@@ -0,0 +1,405 @@
...
+
+        return commands;
+    }
+
+    private static bool IsEnabled()
+    {
+        var configured = Environment.GetEnvironmentVariable("MCP_OUTPUT_SIZE_ENABLED");
+        return !string.IsNullOrWhiteSpace(configured) &&
+            (configured.Equals("true", StringComparison.OrdinalIgnoreCase) || configured == "1");
+    }
```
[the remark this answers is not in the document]
LarryOsterman: This is not relevant any more, it was removed.

servers/Azure.Mcp.Server/tests/Azure.Mcp.Server.Tests/Infrastructure/McpOutputSizeTests.cs
```diff
@@ -0,0 +1,405 @@
...
+
+namespace Azure.Mcp.Server.Tests.Infrastructure;
+
+/// <summary>
+/// Measures MCP responses from a client perspective for the server's exposed tool surfaces.
+/// This test is opt-in: set <c>MCP_OUTPUT_SIZE_ENABLED=true</c> to run it, otherwise it is
+/// skipped so it does not slow down normal test runs.
+/// Set <c>MCP_OUTPUT_SIZE_REPORT</c> to change the JSON report path.
+/// Set <c>AZURE_MCP_TEST_THREE_STEP_TOOL_DISCOVERY=true</c> to start the server with
+/// three-step tool discovery; omit it or set it to <c>false</c> for the default two-step mode.
```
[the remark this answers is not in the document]
LarryOsterman: no longer relevant.

eng/scripts/Measure-McpOutputSizes.ps1
```diff
@@ -0,0 +1,127 @@
...
+#Requires -Version 7
+
+<#
+.SYNOPSIS
+    Builds and runs the MCP output size measurement test, then summarizes the results.
+
+.DESCRIPTION
+    Runs the full measurement workflow end to end:
+
+    1. Builds the Azure.Mcp.Server.Tests project (which also builds the azmcp server).
```
alzimmermsft: Long term would be nice to expand this out to support any MCP server in this repo

eng/scripts/Measure-McpOutputSizes.ps1
```diff
@@ -0,0 +1,127 @@
...
+    3. Runs Summarize-McpOutputSizes.ps1 to produce console, JSON, and Markdown
+       summaries, extract readable description text, and split each top tool's inner
+       commands into individual files.
+
+    The measurement test is opt-in and normally skipped, so this script sets
+    MCP_OUTPUT_SIZE_ENABLED for the duration of the run.
+
+.PARAMETER OutputDirectory
+    Directory for the measurement report and its artifacts.
+    Defaults to `<repo-root>/TestResults`.
```
alzimmermsft: Most other scripts have defaulted to `/.work`, would be good to continue that pattern as the default

eng/scripts/Measure-McpOutputSizes.ps1
```diff
@@ -0,0 +1,127 @@
...
+}
+$OutputDirectory = [IO.Path]::GetFullPath($OutputDirectory)
+
+if ($Clean -and (Test-Path -LiteralPath $OutputDirectory)) {
+    Write-Host "Removing existing output directory $OutputDirectory"
+    Remove-Item -LiteralPath $OutputDirectory -Recurse -Force
+}
+
+New-Item -ItemType Directory -Path $OutputDirectory -Force | Out-Null
+$reportPath = Join-Path $OutputDirectory 'mcp-output-size.json'
```
alzimmermsft, line 479: Do we need `$Clean` if our implicit behavior is overwriting any previous result we captured?

eng/scripts/Measure-McpOutputSizes.ps1
```diff
@@ -0,0 +1,127 @@
...
+}
+
+Write-Host "Running MCP output size measurement test..."
+$previousEnabled = $env:MCP_OUTPUT_SIZE_ENABLED
+$previousReport = $env:MCP_OUTPUT_SIZE_REPORT
+try {
+    $env:MCP_OUTPUT_SIZE_ENABLED = 'true'
+    $env:MCP_OUTPUT_SIZE_REPORT = $reportPath
+
+    & $testExecutable --filter-class Azure.Mcp.Server.Tests.Infrastructure.McpOutputSizeTests --no-progress
```
alzimmermsft: Bit strange that this is calling a unit test with special behaviors whereas most other things like this are standalone applications, like Vally work

eng/scripts/Summarize-McpOutputSizes.ps1
alzimmermsft, line 1: Does this need to be a separate script? Should the other script and where we land on how this works just write this report always?

alzimmermsft (commented):
Will review the CSharp portion after we determine if we want it as a test or a standalone application like many of our other outer loop tooling (like Vally)

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: Agreed.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: I didn't see we had standalone applications that hosted the MCP server directly, but we had a significant amount of infrastructure in the test collateral to handle this case. Rather than invent a new set of paradigms, I figured I'd leverage the existing infrastructure.

The Vally stuff has independent tools to generate the vally output, but not to actually run vally. 

However  I'm 100% open to changing this to a standalone application. But it will require surgery in the test infrastructure to move the logic which invokes the MCP server into a standalone assembly that can be referenced by both the new tool and the test infrastructure.

eng/scripts/Summarize-McpOutputSizes.ps1
LarryOsterman: No it doesn't, but in other code reviews people objected to the large size of the scripts so I broke it up into a script runner and an analysis/output generator.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: Fixed.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: I added the -clean because the tool outputs to the same directory all the time and sometimes there are artifacts from previous runs (this happened yesterday while I was preparing for office hours - I had some artifacts from last week which confused me). 

So when things change, there can be artifacts left over from previous runs that no longer apply.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: Unit test removed and replaced with a standalone application.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: Fixed.

eng/scripts/Summarize-McpOutputSizes.ps1
LarryOsterman: FWIW, it's now a single script.

LarryOsterman:
@alzimmermsft It's now a standalone tool, and a single script :).

LarryOsterman:
Ping :). I'd like to get this in if possible.

eng/scripts/Measure-McpOutputSizes.ps1
```diff
@@ -0,0 +1,627 @@
...
+        $modeResults[$mode.mode] = $mode
+    }
+
+    foreach ($modeName in @('consolidated', 'namespace')) {
+        if (!$modeResults.ContainsKey($modeName)) {
+            throw "The report does not contain the '$modeName' mode."
+        }
+    }
+
+    function Get-PercentDifference([double] $value, [double] $baseline) {
```
alzimmermsft: Feels odd that we are defining functions inside a function, was that intentional?

eng/tools/McpOutputSizeMeasurer/src/McpOutputSizeMeasurer.cs
```diff
@@ -0,0 +1,466 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        if (!File.Exists(executablePath))
+        {
+            throw new FileNotFoundException(
+                $"Executable not found at {executablePath}. Build the server project first.",
+                executablePath);
+        }
+
+        using var process = Process.Start(new ProcessStartInfo
```
alzimmermsft: May want to look at what we're doing in testing for MCP servers so you don't need to manually handle JSON RPC things:

https://github.com/microsoft/mcp/blob/main/core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Client/Helpers/McpTestUtilities.cs#L163

Should also make it easier to handle new MCP protocol versions. E.g., the initialization you're using here no longer exists in newer MCP versions.

eng/scripts/Measure-McpOutputSizes.ps1
LarryOsterman: It's the contents of the original script just embedded. But you're right, it's weird. Let me fix it.

eng/tools/McpOutputSizeMeasurer/build-info.json
alzimmermsft, line 1: Was this an accidental commit?

alzimmermsft (approved):

eng/tools/McpOutputSizeMeasurer/build-info.json
LarryOsterman: It showed up when I started using `Build-Code.ps1` to build the binaries rather than `dotnet build`.

PR #3289, opened 2026-08-14
Add three-step tool discovery mode

LarryOsterman:
## Summary
- add opt-in three-step tool discovery mode for namespace/consolidated learn responses
- keep default two-step behavior unchanged
- cover default and enabled learn-mode behaviors with focused tests

## Related
- stacked on #3288


core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -130,15 +131,25 @@ public override ValueTask<ListToolsResult> ListToolsHandler(RequestContext<ListT
...
 
-            var tool = new Tool
-            {
-                Name = namespaceName,
-                Description = group.Description + """
+            var toolRouterDescription = _options.Value.ThreeStepToolDiscovery
+                ? """
+                    This tool is a hierarchical MCP command router.
+                    Sub commands are routed to MCP servers that require specific fields inside the "parameters" object.
+                    To invoke a command, set "command" to the exact command name (e.g. "appconfig_account_list") and wrap its args in "parameters".
```
alzimmermsft, line 137: Something we need to stop doing is having explicit command names in descriptions. Often the MCP client modifies the name we provide to prevent collisions across MCP servers, so we don't get a nice `appconfig_account_list` but more a `<guid-like-string>_appconfig_account_list`. Aside from that, it also just adds another maintenance spot we're likely to miss in a refactoring.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -130,15 +131,25 @@ public override ValueTask<ListToolsResult> ListToolsHandler(RequestContext<ListT
...
+                    Do not guess a "command" name from natural language. First call this tool with "learn=true" and no "command" to get the exact list of command names and descriptions.
+                    Then call again with "learn=true" and the matching "command" from that list to get its full input schema before invoking it.
```
alzimmermsft, line 139: This is really good guidance but also may be overly prescriptive and forcing the agent to learn again when it's possible it can just skip a step as it had enough context

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
```diff
@@ -541,14 +557,94 @@ There was an error finding or calling tool and command.
...
+                    {
+                        Text = $"""
+                            Here are the available commands for '{namespaceName}' tool.
+                            The "tool" field of each entry below is the exact "command" value to use -- do not guess or invent a
+                            different name. Pick the entry matching your task, then run again with "learn=true" and "command" set
+                            to that exact "tool" value to get its full input schema.
+
+                            {learnToolsJson}
+                            """
+                    }
```
alzimmermsft, line 626: Hmm, something we could change here, but would explicitly require three turns is to simplify `learnToolsJson` returned here to just the tool name and description. Right now, we're returning the input schema as well. So, this current implementation is more a 2.5 turn type implementation.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/NamespaceToolLoader.cs
LarryOsterman: Interesting. The LLM suggested adding this text to prevent the host LLM from guessing the tool names, it proposed adding text to the instructions to explain how to construct the command name.

I'm surprised that the MCP host would change the tool name, but it does make sense in the face of tool ambiguity. So would a better pattern be something like `<tool_name>_add_kv`?

PR #3296, opened 2026-08-17
Correctly capture tool parameters telemetry in multi-step server modes

alzimmermsft:
## What does this PR do?

`ToolParameters` telemetry wasn't being captured correctly for MCP server modes that have multiple phases before a tool is called. Modes such as `single` and `namespace` need to update the parameters telemetry when the tool arguments are inferred from the `parameters` input.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:


core/Microsoft.Mcp.Core/src/Helpers/McpHelper.cs
```diff
@@ -74,16 +74,31 @@ public static CallToolResult InjectToolIdMetadata(CallToolResult result, string?
     /// <summary>
     /// Creates a telemetry-friendly string array representation of the tool's parameter names.
     /// </summary>
-    /// <param name="request">The tool call request parameters.</param>
+    /// <param name="parameters">The tool call parameters.</param>
     /// <returns>A string array representing the parameter names, or null if there wasn't parameters.</returns>
-    public static string? CreateToolParametersTelemetry(RequestContext<CallToolRequestParams> request)
+    public static string? CreateToolParametersTelemetry(IDictionary<string, JsonElement>? parameters)
```
vukelich: For safety's sake: I'd recommend not passing an object type that includes the parameter values. Don't let someone or AI mistakenly think that by passing a dictionary, we want to log both the keys and values. Make the contract to this method clear that we ONLY care about parameter names/keys. Providing a data structure allowing more than that leads to ambiguity.

alzimmermsft: Done

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -3,23 +3,45 @@
...
 using Xunit;
 
 namespace Azure.Mcp.Core.Tests.Areas.Server.Commands.ToolLoading;
 
-public class CommandFactoryToolLoaderTests
+public class CommandFactoryToolLoaderTests : IDisposable
 {
+    private readonly Activity _activity;
+
+    public CommandFactoryToolLoaderTests()
```
conniey: Do we need this for every test case in the class? If it's a subset, feels better to scope it to the test cases it matters.

alzimmermsft: Nope, can make this a test-level thing.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.Mcp.Tests.Helpers;
+
+/// <summary>
+/// Test helper class for validating telemetry.
+/// </summary>
+public static class TestTelemetryHelpers
+{
+    public static object GetAndAssertTagKeyValue(Activity activity, string tagName)
+    {
+        var matching = activity.TagObjects.SingleOrDefault(x => string.Equals(x.Key, tagName, StringComparison.OrdinalIgnoreCase));
```
vukelich: Couple recommendations:

**Case sensitivity**
We may want to assert by case-sensitivity on the tag name. Changing the casing is a breaking change for any consumers of the telemetry, e.g., real App Insights usage by remote self-hosters and 1P teams. 

For our own usability, we can add helpful logic that will detect incorrect casing and report the mismatch directly. Rather than a "tag didn't exist" we have a "tag was in the wrong casing"

**Plurality**
`Activity` tags are not maintained as a dictionary. It's possible to have duplicates of the same tag name. Using `SingleOrDefault` will throw an `InvalidOperationException` when more than one matching element is found. To make our tests clear, we should get all matching tag key-value pairs, then use some xunit `Assert` that tests the size of the collection is our singular item. IIRC, there are asserts for both size and positional elements, and the latter's Assert exception will print out the actual and expected collection elements. Pretty nifty in cases like this to see the violating data in the exception details, but you may need to have the expected value passed as a parameter to this method, which is probably good to do anyway.

In fact, we have this duplicate scenario happening. `CommandFactoryToolLoader.CallToolHandler` sets the same tag twice: [first to `false`](https://github.com/alzimmermsft/mcp/blob/FixToolParametersTelemetryCapture/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs#L91) and [potentially a second time to `true`](https://github.com/alzimmermsft/mcp/blob/7f2baa1a858ff10d12202392025f2209960676e9/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs#L226).

If the tests aren't having issues with this using `SingleOrDefault`, then that may be a sign that we're not thoroughly testing our telemetry changes to `CommandFactoryToolLoader`.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -1314,4 +1322,266 @@ public async Task CallToolHandler_UnknownParameters_RejectsToolCall()
...
+        commandFactory.AllCommands.Returns(new Dictionary<string, IBaseCommand>
+        {
+            ["nevercalled"] = Substitute.For<IBaseCommand>()
+        });
+        var options = Microsoft.Extensions.Options.Options.Create(new ToolLoaderOptions());
+        var logger = Substitute.For<ILogger<CommandFactoryToolLoader>>();
+
+        using var activity = new Activity("test-activity");
+        activity.Start();
+
```
vukelich: Nit: remove extra spacing

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -1314,4 +1322,266 @@ public async Task CallToolHandler_UnknownParameters_RejectsToolCall()
     }
 
     #endregion
+
+    #region Telemetry Tests
+
+    [Fact]
+    public async Task CommandFactoryToolLoader_EmitsErrorTelemetry_IfToolIsFiltered()
```
vukelich: Where is the actual "filtering" happening in this test? Can you please add a comment that makes clear how that setup is happening? It looks like `tool` is not a tool added to the mock command factory, which makes me think this is the same as `CommandFactoryToolLoader_EmitsErrorTelemetry_IfCommandDoesNotExist`

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -1314,4 +1322,266 @@ public async Task CallToolHandler_UnknownParameters_RejectsToolCall()
...
+
+        using var activity = new Activity("test-activity");
+        activity.Start();
+
+        var mcpRuntime = CreateRuntime(new CommandFactoryToolLoader(commandFactory, options, logger), activity);
+        var request = CreateToolCallRequest(mcpServer, toolName);
+
+        await mcpRuntime.CallToolHandler(request, TestContext.Current.CancellationToken);
+
+        Assert.Equal(ActivityStatusCode.Error, activity.Status);
```
vukelich: A recurring problem in the new tests is the inability of the test to prove the expected failure case occurred. The new tests are only validating the details of the request but and that the request failed, but none of this proves the _why_. Do we have a way any error code or message system that can help us? It'd be great to validate through that. Without being able to pinpoint to a root cause in these tests, then we're not truly verifying our behavior. Failure to call a tool could be a test code bug as much as a real product bug, as I indicated with a separate comment asking if this test is correctly written for its intent.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -1314,4 +1322,266 @@ public async Task CallToolHandler_UnknownParameters_RejectsToolCall()
...
+    {
+        var toolName = "tool";
+        var toolArea = "area";
+        var toolId = Guid.NewGuid().ToString();
+        var mcpServer = Substitute.For<McpServer>();
+        var commandFactory = Substitute.For<ICommandFactory>();
+        commandFactory.GetServiceArea(Arg.Is(toolName)).Returns(toolArea);
+
+        var toolCommand = Substitute.For<IBaseCommand>();
+        toolCommand.ExecuteAsync(Arg.Any<CommandContext>(), Arg.Any<ParseResult>(), Arg.Any<CancellationToken>())
```
vukelich: If we keep telemetry test separate from the functionality tests, we may also want to assert the `toolCommand` mock was invoked. It's redundant with the functional tests, but it avoids any weirdness that could come from telemetry passing but functionality failing.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -1314,4 +1322,266 @@ public async Task CallToolHandler_UnknownParameters_RejectsToolCall()
     }
 
     #endregion
+
+    #region Telemetry Tests
```
vukelich: Subjective: I support putting telemetry testing within the same tests for functionality. We're repeating a lot of boilerplate in these telemetry tests that are likely covered by existing functional tests. Duplicating setup and type references makes any refactoring that much more difficult down the line.

vukelich: Separation can also lead to unexpected gaps. A functional unit test might be correctly set up for invoking a tool but somehow fail to send telemetry. Or a telemetry unit test might correctly cause telemetry code paths to hit but the actual functionality gets skipped, shortcut, or silently fails.

vukelich: Actually, let me know if you're going to follow this suggestion. It's going to affect how I review the newly added tests or save me time from doing that.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/NamespaceToolLoaderTests.cs
```diff
@@ -881,6 +884,105 @@ public async Task GetChildToolList_WithIsHttpOption_DoesNotReturnLocalRequiredTo
         });
     }
 
+    // Telemetry tests
+
+    [Fact]
+    public async Task NamespaceToolLoader_HasNamespaceToolParameters_WhenToolDoesNotGetCalled()
```
vukelich, line 890: Back to the "separate functional and telemetry or combined" topic: if these tests are kept separate, please update these telemetry tests to reflect their telemetry intent. The added tests `CommandFactoryToolLoaderTests` do well at identifying themselves. This new test name confused me wondering, "what is the subject of the clause that 'has' the tool parameters in `HasNamespaceToolParameters`?"

core/Azure.Mcp.Core/cspell.yaml
```diff
@@ -8,8 +8,8 @@ import:
   - ../../.vscode/cspell.json
 words:
   - azuer
+  - nevercalled
```
vukelich: Do we really need to add a spelling exception? Can we hyphenate or some other term delimiter instead for the test that uses that phrase? This is a good way to accidentally introduce a typo in the product just because a hardcoded string in a unit test.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -336,7 +327,7 @@ public async Task ListToolsHandler_WhenToolLoaderThrows_PropagatesException()
         mockTelemetry.Received(1).StartActivity(ActivityName.ListToolsHandler, Arg.Any<Implementation?>(), Arg.Any<RequestParams?>());
         Assert.Equal(ActivityStatusCode.Error, activity.Status);
 
-        GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
```
vukelich: Add an assert for the tag value

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -372,10 +363,10 @@ public async Task CallToolHandler_WhenToolLoaderThrows_PropagatesException()
         mockTelemetry.Received(1).StartActivity(ActivityName.ToolExecuted, Arg.Any<Implementation?>(), Arg.Any<RequestParams?>());
         Assert.Equal(ActivityStatusCode.Error, activity.Status);
 
-        var actualToolName = GetAndAssertTagKeyValue(activity, TagName.ToolName);
+        var actualToolName = TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ToolName);
         Assert.Equal(toolName, actualToolName);
 
-        GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
```
vukelich: Add an assert for the tag value

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -499,7 +490,7 @@ public async Task CallToolHandler_WithNullParameters_ReturnsError()
 
         mockTelemetry.Received(1).StartActivity(ActivityName.ToolExecuted, Arg.Any<Implementation?>(), Arg.Any<RequestParams?>());
         Assert.Equal(ActivityStatusCode.Error, activity.Status);
-        GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
```
vukelich: Add an assert for the tag value

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -800,7 +791,7 @@ public async Task CallToolHandler_WithToolLoaderError_ShouldReturnErrorAndSetTel
...
```
vukelich: Add an assert for the tag value

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/SingleProxyToolLoaderTests.cs
```diff
@@ -646,4 +650,101 @@ public async Task CallToolHandler_WithoutModeRestrictions_AllowsNonReadOnlyComma
...
+        request.Params.Arguments?.Add("learn", JsonDocument.Parse("true").RootElement);
+
+        // Act
+        var result = await toolLoader.CallToolHandler(request, TestContext.Current.CancellationToken);
+
+        var toolParameters = TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ToolParameters);
+        Assert.NotNull(toolParameters);
+        var parametersList = JsonSerializer.Deserialize(toolParameters.ToString()!, ModelsJsonContext.Default.ListString);
+        Assert.NotNull(parametersList);
+        Assert.Equal(4, parametersList.Count);
```
vukelich, line 678: Might be able to simplify this with https://api.xunit.net/v3/3.2.2/v3.3.2.2-Xunit.Assert.Equivalent.html. Just make sure to have `strict: true` as per the documentation for counting number of elements.

vukelich: As in, use that method for testing two collections, which the documentation says is supported.

core/Microsoft.Mcp.Core/src/Helpers/McpHelper.cs
```diff
@@ -74,16 +74,16 @@ public static CallToolResult InjectToolIdMetadata(CallToolResult result, string?
...
-    /// <returns>A string array representing the parameter names, or null if there wasn't parameters.</returns>
-    public static string? CreateToolParametersTelemetry(RequestContext<CallToolRequestParams> request)
+    /// <param name="parameterNames">The tool call parameter names.</param>
+    /// <returns>A JSON string array representing the parameter names, or null if there wasn't parameters.</returns>
+    public static string? CreateToolParametersTelemetry(IEnumerable<string>? parameterNames)
     {
-        if (request?.Params?.Arguments == null || request.Params.Arguments.Count == 0)
+        if (parameterNames == null || !parameterNames.Any())
         {
             return null;
```
vukelich, line 82: It might be a breaking change, but I tend to prefer an empty collection over `null` in cases like this. A common problem in telemetry is trusting if `null` means "did not get to the code that sets this value" or "the value is truly `null`." Some telemetry exporters could also choose to drop `null` values entirely, which would lead to a further issue of lacking the property in its entirety.

KQL has semantics for testing each of these cases if we do keep it as-is. You can test for "does this `dynamic` have a key by this name" vs "is the value of this key `null`" vs "is the value of this key a non-null empty"?

vukelich: Only change if someone using this would like to differentiate more clearly or notices that a `null` value causes the property to be dropped entirely.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+        Assert.False(matching.Equals(default(KeyValuePair<string, object?>)), $"Tag '{tagName}' was not found in activity tags.");
+        Assert.NotNull(matching.Value);
+
+        return matching.Value;
+    }
+
+    public static void AssertTagDoesNotExist(Activity activity, string tagName)
+    {
+        var matching = activity.TagObjects.SingleOrDefault(x => string.Equals(x.Key, tagName, StringComparison.OrdinalIgnoreCase));
+        Assert.True(matching.Equals(default(KeyValuePair<string, object?>)), $"Tag '{tagName}' was found in activity tags but should not exist.");
```
vukelich: Use `Assert.Equal()` or `Assert.NotEqual()`. Using xunit's Assert methods provide helpful information about the `expected` and `actual` arguments to the methods which makes troubleshooting much easier in build pipeline failures.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <summary>
+/// Test helper class for validating telemetry.
+/// </summary>
+public static class TestTelemetryHelpers
+{
+    public static object GetAndAssertTagKeyValue(Activity activity, string tagName)
+    {
+        var matching = activity.TagObjects.SingleOrDefault(x => string.Equals(x.Key, tagName, StringComparison.OrdinalIgnoreCase));
+
+        Assert.False(matching.Equals(default(KeyValuePair<string, object?>)), $"Tag '{tagName}' was not found in activity tags.");
```
vukelich: Use `Assert.Equal()` or `Assert.NotEqual()`. Using xunit's Assert methods provide helpful information about the `expected` and `actual` arguments to the methods which makes troubleshooting much easier in build pipeline failures.

vukelich (changes requested):
Various comments mostly around the new or moved tests.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
alzimmermsft: Will harden the testing behaviors here.

Luckily this:

> In fact, we have this duplicate scenario happening. CommandFactoryToolLoader.CallToolHandler sets the same tag twice: [first to false](https://github.com/alzimmermsft/mcp/blob/FixToolParametersTelemetryCapture/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs#L91) and [potentially a second time to true](https://github.com/alzimmermsft/mcp/blob/7f2baa1a858ff10d12202392025f2209960676e9/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoader.cs#L226).

Is using `SetTag` which is an overwrite operation if the tag already exists. I believe we did have a bug here at one point when we were using `AddTag`.

core/Microsoft.Mcp.Core/src/Helpers/McpHelper.cs
alzimmermsft: This should be left as-is as the incoming `IEnumerable<string>` flows in from code in the ModelContextProtocol SDK, where `Arguments` can be null, therefore names can be null. Would prefer this over having `nullableEnumerable ?? []` or conditional calling on the enumerable being null (both being the same issue just pushed up to the caller to handle)

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -499,7 +492,8 @@ public async Task CallToolHandler_WithNullParameters_ReturnsError()
 
         mockTelemetry.Received(1).StartActivity(ActivityName.ToolExecuted, Arg.Any<Implementation?>(), Arg.Any<RequestParams?>());
         Assert.Equal(ActivityStatusCode.Error, activity.Status);
-        GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        var exceptionType = TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        Assert.Equal("InvalidParameters", exceptionType);
```
vukelich, line 496: Is there really an exception type called `InvalidParameters`? I'd expect the type to be suffixed with the commonplace `Exception` substring. Where is this type coming from? I noticed it's not being referenced but instead hardcoded.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
```diff
@@ -800,14 +794,15 @@ public async Task CallToolHandler_WithToolLoaderError_ShouldReturnErrorAndSetTel
 
         mockTelemetry.Received(1).StartActivity(ActivityName.ToolExecuted, Arg.Any<Implementation?>(), Arg.Any<RequestParams?>());
         Assert.Equal(ActivityStatusCode.Error, activity.Status);
-        GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        var exceptionType = TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.ExceptionType);
+        Assert.Equal("ToolCallError", exceptionType);
```
vukelich, line 798: Similar question as https://github.com/microsoft/mcp/pull/3296/changes#r3816097280

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Services/Azure/Authentication/CustomChainedCredentialTests.cs
```diff
@@ -439,14 +439,14 @@ public void NonProdPinnedCredential_WithForceBrowserFallback_CreatesCredentialSu
...
-    /// the default credential chain. This catches typos such as "AzuerCliCredential".
+    /// the default credential chain.
     /// </summary>
     [Fact]
     public void UnknownCredentialValue_LogsWarning_AndFallsBackToDefaultChain()
     {
         // Arrange
         using var env = new EnvironmentScope("AZURE_TOKEN_CREDENTIALS");
-        Environment.SetEnvironmentVariable("AZURE_TOKEN_CREDENTIALS", "AzuerCliCredential"); // intentional typo
+        Environment.SetEnvironmentVariable("AZURE_TOKEN_CREDENTIALS", "MadeUpCredential"); // intentional typo
```
vukelich: Comment of `// intentional typo` is no longer accurate.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+/// </summary>
+public static class TestTelemetryHelpers
+{
+    public static object GetAndAssertTagKeyValue(Activity activity, string tagName)
+    {
+        var matching = activity.TagObjects.Where(x => string.Equals(x.Key, tagName));
+
+        Assert.Single(matching);
+        Assert.NotNull(matching.First().Value);
+        return matching.First().Value!;
```
vukelich: Since `Assert.NotNull` uses `[NotNull]` on the first argument, the compiler will know the value is non-null if `Assert.NotNull` returns (i.e., doesn't throw an exception).

For your code to take advantage of this, you need to use a local variable. `First()` and `Value` are both methods, so the complier can't leverage these nullability hints.

```csharp
var matching = activity.TagObjects
    .Where(x => string.Equals(x.Key, tagName))
    // We call ToList() here because it'll "materialize" the IEnumerable.
    .ToList();

Assert.Single(matching);

// We have array indexing, so use that instead of IEnumerable.First().
var first = matching[0].Value;
Assert.NotNull(first);

// Since Assert.NotNull has the [NotNull] attribute on the first argument,
// the compiler knows that the reference is not null. We can avoid
// having to use the null-forgiveness operator (!) like "return first!;"
return first; 
```

[Postconditions: `MaybeNull` and `NotNull`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull)

[`Assert.NotNull` source code](https://github.com/xunit/assert.xunit/blob/main/NullAsserts.cs#L28)

In the suggestion above, you'll noticed I also played with the `IEnumerable` calls. If you're unfamiliar with how `IEnumerable` extension methods like `Where` and `First` are chained and have deferred execution, then I'd recommend reading up and playing with those .NET concepts.

Summary doc: [Deferred execution and lazy evaluation (LINQ to XML)](https://learn.microsoft.com/en-us/dotnet/standard/linq/deferred-execution-lazy-evaluation)
- Another way to think about it: there's a difference between building a query and running a query. Some extension methods consuming an `IEnumerable` chain together operations of a query, while other extension methods actually run the query. Calling `Where()` doesn't perform anything, but instead it just chains another step and represents the full chain of operations as its returned `IEnumerable`. Calling `First()` executes everything cobbled together and returns an actual value rather than another `IEnumerable`. Therefore, if you call `First()` twice on the same `IEnumerable`, you're causing all chained operations to evaluate twice. In fact, the `Where` is being executed by `Assert.Single` as well for it to evaluate the number of collection elements, making the `Where` executed a total of three times in the current form.

Example doc: [Deferred execution example (LINQ to XML)](https://learn.microsoft.com/en-us/dotnet/standard/linq/deferred-execution-example)
- Run this example yourself and modify it to iterate over `q` twice. Just copy and paste that `foreach` a second time.

vukelich (commented):
Still reviewing. Current items.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
alzimmermsft: Telemetry's `exception.type` is either the fully qualified name of the C# exception handled or a custom string for when we have failures that aren't exceptions. In this case the exception is `InvalidParameters` as the passed input didn't include any parameters (required) but wasn't an exception.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/Runtime/McpRuntimeTests.cs
alzimmermsft: This is a generic fallback when `tools/call` returned with `CallToolResult.IsError` being true but we don't have more information available. There is a TODO around this on whether we can parse information from `CallToolResult` to get more details, but this can happen when proxying tool calls for `registry.json` where everything is opaque to us.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Diagnostics;
+using Xunit;
+
+namespace Microsoft.Mcp.Tests.Helpers;
+
+/// <summary>
+/// Test helper class for validating telemetry.
+/// </summary>
+public static class TestTelemetryHelpers
```
vukelich, line 12: For future changes, I'd recommend calling a class with only xunit-style asserts something like `TelemetryAsserts` or `ActivityAsserts` or `AssertXyz`. As I type this, I forget which format is the conventional preference of users for the library. Other test libraries use fluid notation, in which I'd encourage something like `ActivityAssertExtensions` as the static class name, but I'm not sure if mixing fluid and non-fluid is common for Xunit users.

With `Assert` or `Asserts` somewhere the class name (assuming a non-fluid convention), then `Assert` in method names becomes redundant. We can then name these something like `SingleTagByName` and `TagDoesNotExist`. (Note that I didn't say `SingleTagByName` with the `Get`. Some Xunit `Assert` static methods return `void` and others return a value. I'm not sure why the inconsistency. Returning a value is good for fluid conventions, but Xunit isn't itself structured that way from how I've used it. I'd prefer always having a return so that we can write assignment statements like `var tagValue = AssertActivity.SingleTagByName(activity, "foo");`.)

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -32,20 +38,54 @@ private static (CommandFactoryToolLoader toolLoader, ICommandFactory commandFact
...
+        var commandMapField = typeof(CommandFactory).GetField("_commandMap", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+        var commandMap = (Dictionary<string, IBaseCommand>)commandMapField!.GetValue(commandFactory)!;
+        commandMap[fakeCommand.GetCommand().Name] = fakeCommand;
+    }
+
+    private static RequestContext<ListToolsRequestParams> CreateToolListRequest() =>
+        new(Substitute.For<McpServer>(), new() { Method = RequestMethods.ToolsList }, new());
+
+    private static RequestContext<CallToolRequestParams> CreateToolCallRequest(string toolName, McpServer? mcpServer = null) =>
+        new(mcpServer ?? Substitute.For<McpServer>(), new() { Method = RequestMethods.ToolsCall }, new()
```
vukelich, line 78: Optional for time:
The variety of syntax on this line is hard to visually read. I'd generally advocate for putting parameters on separate lines in pretty much all cases (ex: source code is generally line-based, so diffs/blame are more convenient), but I'm much stronger on that recommendation when there are expressions for individual parameters and especially when we start using curly or square braces.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -325,32 +365,44 @@ public async Task CallToolHandler_WithValidTool_ExecutesSuccessfully()
...
+        var mcpRuntime = CreateRuntime(toolLoader, activity);
+
+        var result = await mcpRuntime.CallToolHandler(request, TestContext.Current.CancellationToken);
 
         Assert.NotNull(result);
         Assert.NotNull(result.Content);
         Assert.NotEmpty(result.Content);
+
+        // Validate telemetry
+        Assert.Equal(result.IsError == true ? ActivityStatusCode.Error : ActivityStatusCode.Ok, activity.Status);
```
vukelich, line 382: Why is the expected value being computed? I would expect we know upfront what value we want to have here.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
```diff
@@ -325,32 +365,44 @@ public async Task CallToolHandler_WithValidTool_ExecutesSuccessfully()
...
+        Assert.Equal(true, TestTelemetryHelpers.GetAndAssertTagKeyValue(activity, TagName.IsServerCommandInvoked));
```
vukelich, line 383: Future PRs: consider making a custom assert method that can validate the entirety of an `Activity` or other telemetry. It's hard because there are a lot of things that _could_ be validated, but now with so many tests in place, we may see straight-forward patterns.

vukelich (approved):
Still going to do some more reviewing, but approving to unblock desire to ship this week.

core/Azure.Mcp.Core/tests/Azure.Mcp.Core.Tests/Areas/Server/Commands/ToolLoading/CommandFactoryToolLoaderTests.cs
alzimmermsft: This test needs to be updated as it just pull a random tool from what exists in Azure MCP and tries to use it. If the tool has required parameters it will fail. So, it's a non-deterministic test.

core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Tests/Helpers/TestTelemetryHelper.cs
vukelich: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#extension-members

oh wait, C#14 added the ability for us to do the following. It means we can create extension methods that are static on `Assert`. 

image

PR #3322, opened 2026-08-19
Azure Backup: resourceguard command group + MUA disable safety fix

shrja-ms:
# Azure Backup: resourceguard command group + MUA disable safety fix

This PR delivers two related improvements from `azurebackup-rsv-mcp-improvements-plan.md` in a single change:

## PR 3 — New `azurebackup resourceguard` command group

Adds `create`, `get`, and `delete` operations for `Microsoft.DataProtection/resourceGuards`, the resource that backs Multi-User Authorization (MUA) on Recovery Services vaults and Backup vaults.

- `azurebackup resourceguard create` — provision a new Resource Guard with optional `--excluded-operations` and `--tags`
- `azurebackup resourceguard get` — retrieve details for a Resource Guard
- `azurebackup resourceguard delete` — remove a Resource Guard

The service layer defaults `s_mandatoryResourceGuardOps` (`disableSoftDelete`, `disableMultiUserAuthorization`, `removeMUAProtection`, `disableSecurityFeatures`) so critical destructive operations are always gated by MUA.

## PR 6 — Fix `security configure-mua` safety bug + add `security disable-mua`

**Problem:** `azurebackup security configure-mua` accepted an empty `--resource-guard-id`, in which case the underlying SDK call *silently disabled* MUA on the vault instead of failing. This meant a benign-looking "configure" invocation with a typo or missing argument could remove a key security control without any confirmation.

**Fix:**

- `--resource-guard-id` is now **required** on `configure-mua`. Missing/empty values return `BadRequest` with a message that also points users at `disable-mua`.
- Added a new `azurebackup security disable-mua` command that explicitly disables MUA and **requires `--force`**. It is marked `Destructive`.
- Added `mua.action` telemetry tag (`enable` / `disable`) so both commands can be tracked separately.

Behavior migration:

| Before | After |
|---|---|
| `configure-mua --resource-guard-id <id>` | Same — enables MUA |
| `configure-mua` (no id) → silently disables MUA | Returns `BadRequest`, directs user to `disable-mua` |
| (n/a) | `disable-mua --force` — new explicit disable command |

## Files changed

- **New commands**: `Commands/ResourceGuard/{Create,Get,Delete}Command.cs`, `Commands/Security/SecurityDisableMuaCommand.cs`
- **New options**: `Options/ResourceGuard/*`, `Options/Security/SecurityDisableMuaOptions.cs`
- **New models**: `Models/ResourceGuardInfo.cs`
- **Service**: added `DisableMultiUserAuthorizationAsync` + 3 resource-guard methods to `IAzureBackupService` / `AzureBackupService`
- **Telemetry**: `AzureBackupTelemetryTags.AddMuaActionTag`, `AddResourceGuardOperationTag`
- **Setup**: `AzureBackupSetup.cs` registers the new group and commands
- **Docs**: `azmcp-commands.md` (rewrote MUA section, added Resource Guard section), `e2eTestPrompts.md` (12 new prompts), `consolidated-tools.json` (mapped new tools)
- **Changelog entry**: `shrja-azurebackup-resource-guard-mua-safety.yml`
- **cspell**: added `resourceguard` to workspace dictionary

## Tests

- New unit tests for all three `resourceguard` commands and both `security` commands
- New integration test methods `SecurityDisableMua_RsvVault_Successfully` and `SecurityDisableMua_DppVault_Successfully` (`[LiveTestOnly]`) with tolerant error accept set
- Full test suite: **749 total / 740 passed / 9 skipped (`[LiveTestOnly]`) / 0 failed**

## Tool Description Evaluator

I confirm each new tool description scores at least `0.4` on `ToolDescriptionEvaluator`.

## Invoking Livetests


servers/Azure.Mcp.Server/docs/azmcp-commands.md
```diff
@@ -1140,6 +1151,39 @@ azmcp azurebackup security configure-encryption --subscription <subscription> \
                                                 [--user-assigned-identity-id <user-assigned-identity-id>]
 ```
 
+#### Resource Guard
+
+```bash
+# Creates a Resource Guard for Multi-User Authorization (MUA). Once a vault is linked to this
+# Resource Guard, protected operations (disable soft delete, remove immutability, stop protection,
+# disable MUA) will require approval from a security admin with Backup MUA Admin role on the guard.
+# ✅ Destructive | ❌ Idempotent | ❌ OpenWorld | ❌ ReadOnly | ❌ Secret | ❌ LocalRequired
```
alzimmermsft, line 1160: Is this actually destructive? Will multiple create calls overwrite each other?

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardCreateCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+        var dict = new Dictionary<string, string>(StringComparer.Ordinal);
+        foreach (var pair in tags.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries))
+        {
+            var idx = pair.IndexOf('=');
+            if (idx < 0)
+            {
+                throw new ArgumentException($"Invalid --tags value '{pair}'. Expected format: 'key1=value1,key2=value2'.");
+            }
+            var key = pair[..idx].Trim();
+            var value = pair[(idx + 1)..].Trim();
```
alzimmermsft, line 103: Is an empty tag value invalid on the service side? And does the service side retain leading and trailing whitespace characters?

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardDeleteCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+        RequestFailedException reqEx => reqEx.Message,
+        _ => base.GetErrorMessage(ex)
+    };
+
+    protected override HttpStatusCode GetStatusCode(Exception ex) => ex switch
+    {
+        ArgumentException => HttpStatusCode.BadRequest,
+        RequestFailedException reqEx => (HttpStatusCode)reqEx.Status,
+        _ => base.GetStatusCode(ex)
+    };
```
alzimmermsft: Can be removed as this is just a subset of the base implementation

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardGetCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+        ArgumentException argEx => argEx.Message,
+        RequestFailedException reqEx when reqEx.Status == (int)HttpStatusCode.NotFound =>
+            "Resource Guard not found. Verify the name and resource group.",
+        RequestFailedException reqEx when reqEx.Status == (int)HttpStatusCode.Forbidden =>
+            $"Authorization failed. Ensure you have Reader role on the resource group or subscription. Details: {reqEx.Message}",
+        RequestFailedException reqEx => reqEx.Message,
+        _ => base.GetErrorMessage(ex)
+    };
+
+    protected override HttpStatusCode GetStatusCode(Exception ex) => ex switch
```
alzimmermsft: Same as the other command where this is just a subset of the base class

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityConfigureMuaCommand.cs
```diff
@@ -17,11 +17,10 @@ namespace Azure.Mcp.Tools.AzureBackup.Commands.Security;
     Name = "configure-mua",
```
alzimmermsft: Does this configure anything else other than enabling mua? If not, let's just rename this to `enable-mua`.

Also, did we look at improving the handling of this tool rather than split it?

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityDisableMuaCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+        UnauthorizedAccessException => "Authorization failed. Verify your RBAC permissions on the vault, or specify --vault-type to skip auto-detection.",
+        RequestFailedException reqEx when reqEx.Status == (int)HttpStatusCode.NotFound =>
+            "Vault not found, or MUA is not enabled for this vault. Verify the vault name and resource group. If MUA is not configured on this vault there is nothing to disable.",
+        RequestFailedException reqEx when reqEx.Status == (int)HttpStatusCode.Forbidden =>
+            $"Authorization failed. Disabling MUA requires the Backup MUA Operator role on the linked Resource Guard. Details: {reqEx.Message}",
+        RequestFailedException reqEx => reqEx.Message,
+        _ => base.GetErrorMessage(ex)
+    };
+
+    protected override HttpStatusCode GetStatusCode(Exception ex) => ex switch
```
alzimmermsft: Another spot of this being a subset of the base implementation

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityDisableMuaCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+    private readonly IAzureBackupService _azureBackupService = azureBackupService;
+
+    public override void ValidateOptions(SecurityDisableMuaOptions options, ValidationResult validationResult)
+    {
+        base.ValidateOptions(options, validationResult);
+
+        if (!options.Force)
+        {
+            validationResult.Errors.Add("--force is required to disable Multi-User Authorization. Disabling MUA removes the Resource Guard's protection from critical vault operations. Pass --force to confirm.");
+        }
```
alzimmermsft: Something we should think about is how valuable is this kind of design. Destructive tools already require elicitation by default and an LLM should be smart enough to always pass `--force` after parsing the description.

So, does this parameter meaningfully do something or is it just noise.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/ResourceGuard/ResourceGuardCreateOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+public sealed class ResourceGuardCreateOptions : ISubscriptionOption
+{
+    [Option(Description = "Name of the Resource Guard to create (Microsoft.DataProtection/resourceGuards).")]
+    public required string ResourceGuard { get; set; }
+
+    [Option(Description = "Azure region where the Resource Guard is created. Must match the region of vaults that will link to it (e.g., 'eastus2', 'westeurope').")]
+    public required string Location { get; set; }
+
+    [Option(Description = "Comma-separated list of critical operations to EXCLUDE from Resource Guard protection. Valid RSV values: deleteProtection, getSecurityPIN, updatePolicy, updateProtection. Cannot include mandatory operations (disableSoftDelete, disableMultiUserAuthorization, removeMUAProtection, disableSecurityFeatures). Omit to protect all default critical operations.")]
```
alzimmermsft:
```suggestion
    [Option(Description = "Comma-separated list of critical operations to EXCLUDE from Resource Guard protection. Valid RSV values: deleteProtection, getSecurityPIN, updatePolicy, updateProtection. Omit to protect all default critical operations.")]
```

Do we need to discuss the invalid options here is we're already discussion which options are allowed?

Also, does DPP have a set of valid options? Or is this an RSV-only feature?

tools/Azure.Mcp.Tools.AzureBackup/src/Options/Security/SecurityDisableMuaOptions.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Mcp.Core.Options;
+
+namespace Azure.Mcp.Tools.AzureBackup.Options.Security;
+
+public sealed class SecurityDisableMuaOptions : BaseAzureBackupOptions
+{
+    [Option(Description = "Required confirmation flag. Set --force to acknowledge that disabling Multi-User Authorization removes the Resource Guard's protection from critical vault operations (disable soft delete, remove immutability, stop protection). Without this flag the command fails.")]
```
alzimmermsft: Just bringing up the same question in another spot. An LLM would likely infer that `--force` is required when it is going to disable MUA, rendering this option effectively pointless.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
```diff
@@ -873,6 +875,173 @@ public async Task<OperationResult> ConfigureEncryptionAsync(
             : await dppOps.ConfigureEncryptionAsync(vaultName, resourceGroup, subscription, keyVaultUri, keyName, identityType, keyVersion, userAssignedIdentityId, tenant, retryPolicy, cancellationToken);
     }
 
+    // ---------------------------------------------------------------------
+    // Resource Guard (Microsoft.DataProtection/resourceGuards) operations.
+    // Resource Guards are subscription/RG-scoped, not vault-scoped. They are
+    // referenced by vaults (RSV or DPP) via Multi-User Authorization (MUA).
+    // ---------------------------------------------------------------------
+
+    private static readonly HashSet<string> s_mandatoryResourceGuardOps = new(StringComparer.OrdinalIgnoreCase)
```
alzimmermsft, line 886: High-level design question, will maintenance and correct behavior be better handled by a deny list (what you have here) or an allow list?

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
```diff
@@ -873,6 +875,173 @@ public async Task<OperationResult> ConfigureEncryptionAsync(
...
+    private static ResourceGuardInfo ToResourceGuardInfo(ResourceGuardData data)
+    {
+        var id = data.Id;
+        var name = data.Name;
+        var location = data.Location.ToString();
+        // ARM ID: /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.DataProtection/resourceGuards/{name}
+        var resourceGroup = id?.ResourceGroupName ?? string.Empty;
+
+        var exclusions = data.Properties?.VaultCriticalOperationExclusionList is { } excl
+            ? (IReadOnlyList<string>)[.. excl]
```
alzimmermsft: Is using `IReadOnlyList` helpful in the model? We're an application and should know not to modify elements in the result types and just to pass them on to serialization for creating the response payload.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
```diff
@@ -873,6 +875,173 @@ public async Task<OperationResult> ConfigureEncryptionAsync(
...
+        {
+            var conflicts = excludedOperations
+                .Where(o => s_mandatoryResourceGuardOps.Contains(o))
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToList();
+            if (conflicts.Count > 0)
+            {
+                throw new ArgumentException(
+                    $"The following operations cannot be excluded from a Resource Guard because they are mandatory: {string.Join(", ", conflicts)}. " +
+                    $"Remove them from --excluded-operations. Typical valid RSV exclusion values are: deleteProtection, getSecurityPIN, updatePolicy, updateProtection.");
```
alzimmermsft, line 947: Does the service also treat the excluded operations case-insensitively?

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardDeleteCommand.cs
shrja-ms: Removed. The base BaseCommand.GetStatusCode already maps ArgumentException, InvalidOperationException, and RequestFailedException the same way.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardGetCommand.cs
shrja-ms: Removed here too.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityDisableMuaCommand.cs
shrja-ms: Removed.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityConfigureMuaCommand.cs
shrja-ms: Renamed the command to azurebackup security enable-mua (docs, tests, e2e prompts, changelog all updated). Kept the split with disable-mua so the two operations can be gated with different Azure RBAC: enable-mua only needs Backup Contributor on the vault, while disable-mua also requires Backup MUA Admin on the linked Resource Guard. Merging into a single muaState flag would either request unnecessarily broad permissions for enable, or fail on disable when the caller only has the enable-side role.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/Security/SecurityDisableMuaCommand.cs
shrja-ms: Fair point. Dropped --force from disable-mua. The destructive metadata plus elicitation already covers the confirmation flow, and any LLM will happily pass --force after reading the description, so the option was not adding signal.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/Security/SecurityDisableMuaOptions.cs
shrja-ms: Removed --force from disable-mua (and from resourceguard delete) for the same reason.

tools/Azure.Mcp.Tools.AzureBackup/src/Options/ResourceGuard/ResourceGuardCreateOptions.cs
shrja-ms: Applied the suggestion. Description now lists valid RSV values only and drops the mandatory-ops enumeration. DPP resource guards currently only support the same four operations at the ARM API level (they share the Microsoft.DataProtection/resourceGuards type); the service applies the same allow list.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
shrja-ms: Changed ResourceGuardInfo to expose List<string> and Dictionary<string, string>. The service builder now materializes to concrete collections directly.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
shrja-ms: Went with a deny list because the set of mandatory operations is short, well-defined by the backup service, and unlikely to shrink (removing an op from the mandatory set would be a service-side breaking change). An allow list would require us to enumerate the full valid set here and keep it in sync with every new operation the service adds. Happy to invert if you prefer the stricter posture.

tools/Azure.Mcp.Tools.AzureBackup/src/Services/AzureBackupService.cs
shrja-ms: The ARM API accepts the four values case-insensitively, but the service normalizes them to camelCase on the response. Using HashSet<string>(StringComparer.OrdinalIgnoreCase) on our end keeps the client validation tolerant of what the LLM emits so we surface a client-side error instead of a round-trip 400.

tools/Azure.Mcp.Tools.AzureBackup/src/Commands/ResourceGuard/ResourceGuardCreateCommand.cs
shrja-ms: The service accepts empty tag values but preserves whitespace verbatim, which means a caller passing 'key= ' ends up with a tag whose value is a single space (impossible to unset without a follow-up call). I kept the current trimming/rejection behavior so the tool fails loudly for these degenerate inputs rather than silently persisting them.

servers/Azure.Mcp.Server/docs/azmcp-commands.md
shrja-ms: Yes. A second resourceguard create with the same name is a PUT that overwrites the existing guard's tags and vaultCriticalOperationExclusionList. Kept Destructive = true in the metadata to force elicitation on the overwrite path. Left Idempotent = false because the operation is not state-preserving when the guard already exists (tags/excluded-ops are replaced, not merged), which matches the rest of the create_azure_backup_resources group.

alzimmermsft (approved):

alzimmermsft (approved):

PR #3368, opened 2026-08-25
Add Azure operation-plane metadata

### LINKED ISSUE #1616: Feature request for Azure MCP CLI: add custom annotation for Data, Control, Both (plane)
Because there isn't a control plane and data plane MCP servers, it should be clear to users which a tool or tool family for an Azure service offers control plane or both. I assume that most tool families for an Azure service are data plane. 

In order to not cause an issue, provide the annotation and set it to data plane by default. 

Annotations = ToolMetadata

```
    public override ToolMetadata Metadata => new()
    {
        Destructive = false,
        Idempotent = false,
        OpenWorld = false,
        ReadOnly = false,
        LocalRequired = false,
        Secret = false
    };
```

Another option is to provide this at the service level

```
public class EventGridService(ISubscriptionService subscriptionService, ITenantService tenantService, ILogger<EventGridService> logger, IHttpClientFactory httpClientFactory)
    : BaseAzureService(tenantService), IEventGridService
{
    private readonly ISubscriptionService _subscriptionService = subscriptionService ?? throw new ArgumentNullException(nameof(subscriptionService));
    private readonly ILogger<EventGridService> _logger = logger ?? throw new ArgumentNullException(nameof(logger));
    private readonly IHttpClientFactory _httpClientFactory = httpClientFactory ?? throw new ArgumentNullException(nameof(httpClientFactory));

    public async Task<List<EventGridTopicInfo>> GetTopicsAsync(
        string subscription,
        string? resourceGroup = null,
        string? tenant = null,
        RetryPolicyOptions? retryPolicy = null,
        CancellationToken cancellationToken = default)
    {
        var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken);
        var topics = new List<EventGridTopicInfo>();

        if (!string.IsNullOrEmpty(resourceGroup))
        {
            // Get topics from specific resource group
            var resourceGroupResource = await subscriptionResource
                .GetResourceGroupAsync(resourceGroup, cancellationToken);

            await foreach (var topic in resourceGroupResource.Value.GetEventGridTopics().GetAllAsync(cancellationToken: cancellationToken))
            {
                topics.Add(CreateTopicInfo(topic.Data));
            }
        }
        else
        {
            // Get topics from all resource groups in subscription
            await foreach (var topic in subscriptionResource.GetEventGridTopicsAsync(cancellationToken: cancellationToken))
            {
                topics.Add(CreateTopicInfo(topic.Data));
            }
        }

        return topics;
    }
```

Ultimately, provide the information in a way that makes it to the CLI output JSON.  This is necessary to docs generation to accurately indicate to the customer the breadth of coverage the tool or tool family provides.

KarishmaGhiya:
## What does this PR do?

Adds an `OperationPlane` value to `CommandMetadata` so every tool declares whether it acts against a control-plane (ARM) or data-plane API, and classifies all tools in the repo.

`ToolOperationPlane` is a new enum in `Microsoft.Mcp.Core` with five values:

| Value | Meaning |
| --- | --- |
| `Control` | The tool performs its action(s) against a resource-management API. |
| `Data` | The tool performs its action(s) against a resource's own workload API. |
| `Both` | The tool performs two genuine user-facing actions, one on each plane. |
| `NotApplicable` | The tool calls no remote service (embedded docs, local content, server introspection). |
| `Unspecified` | Not declared. Rejected by an enforcement test. |

**Classification rule.** A tool's plane is the plane of the API it acts against to produce the result the user asked for. Calls made only as *setup* — subscription or tenant resolution, or a resource lookup done solely to read an endpoint — never contribute a plane. Without this rule nearly every data-plane tool would resolve its target through ARM first and collapse into `Both`, and the value would stop discriminating.

`EventsPublishCommand` is the worked example: it reads the topic endpoint from ARM, then publishes through `EventGridPublisherClient`, so it is `Data`, not `Both`. `EventsPublishCommandTests.Command_OperationPlane_IgnoresArmLookupUsedOnlyForSetup` pins that.

**Coverage.** All tools are classified in this PR rather than migrated incrementally, per review feedback — an `Unspecified` interim state would have shipped a value consumers could not rely on. `CommandOperationPlaneTests` walks every registered tool and fails on `Unspecified`, with no allowlist.

**Serialization.** `ToolMetadataJsonConverter.ToJsonValue` enumerates every value explicitly and throws on an unrecognized one, so a future enum value cannot silently serialize as unspecified. The read path stays lenient.

Design notes are in `docs/design/operation-plane-metadata.md`.

Follow-ups agreed during review and tracked separately: #3541 (an `Internal` plane for server-bound tools), #3542 (making `CommandMetadata` values explicitly required rather than defaulted), #3543 (a design document covering `CommandMetadata` as a whole).

## GitHub issue number?

Fixes #1616

## Pre-merge Checklist

- [x] Required for All PRs
    - [x] Added or updated unit tests (`CommandOperationPlaneTests`, `EventsPublishCommandTests`, `ToolMetadataJsonConverter` coverage)
    - [x] Added a changelog entry (`servers/Azure.Mcp.Server/changelog-entries/kaghiya-operation-plane-metadata.yaml`)
    - [x] Ran `dotnet format`
    - [x] Ran the spell checker
- [x] Not applicable
    - [x] New tool docs (`azmcp-commands.md`, `e2eTestPrompts.md`, `consolidated-tools.json`) — no tools are added, removed, or renamed, and no tool descriptions change, so no `ToolDescriptionEvaluator` score applies.
    - [x] Live test infrastructure — this change adds no service calls.

## Invoking Livetests


docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Classification
+
+| Value | Meaning |
+|---|---|
+| `Data` | The command operates against an Azure service data-plane API. |
+| `Control` | The command operates against Azure Resource Manager or another management-plane API. |
+| `Both` | The command directly uses both control-plane and data-plane APIs. |
+| `NotApplicable` | The command does not perform an Azure service-plane operation, such as `tools list`, `server start`, or a local-only utility. |
+| `Unspecified` | The command has not yet been reviewed and classified. This is a migration state, not a final classification. |
```
alzimmermsft: I'd prefer we don't have this and just migrate everything at once, if possible.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,125 @@
...
+    LocalRequired = false)]
+```
+
+### Classification
+
+| Value | Meaning |
+|---|---|
+| `Data` | The command operates against an Azure service data-plane API. |
+| `Control` | The command operates against Azure Resource Manager or another management-plane API. |
+| `Both` | The command directly uses both control-plane and data-plane APIs. |
```
alzimmermsft: How do we treat look ups for subscription and tenant IDs? If we're going to consider those `Control` most tools will be `Both`, but I don't think I'd want to do that as that results in more confusing information.

tools/Azure.Mcp.Tools.EventGrid/src/Commands/Events/EventsPublishCommand.cs
```diff
@@ -22,6 +22,7 @@ namespace Azure.Mcp.Tools.EventGrid.Commands.Events;
         Event Grid topics with schema validation and delivery guarantees for downstream subscribers. Returns publish operation 
         status. Requires topic, data, and optional schema.
         """,
+    OperationPlane = ToolOperationPlane.Both,
```
g2vinay: This is only `Both` because `publish` resolves the topic endpoint through ARM (`GetSubscription` + `FindTopic` then `topic.Data.Endpoint`) before the data-plane send. If an ARM lookup to find the endpoint counts as control-plane, nearly every data-plane tool becomes `Both`. I'd classify by primary intent. This is a data-plane publish, so I'd treat endpoint/subscription resolution as incidental plumbing and mark it `Data`.

docs/design/operation-plane-metadata.md
g2vinay: Building on Al's question about subscription/tenant lookups: suggest defining the plane by the operation the user actually asked for, explicitly excluding subscription/tenant/resource-id/endpoint resolution. Then `Both` is reserved for tools doing a real user-meaningful operation on each plane, which is rare, and `EventsPublish` becomes `Data`.

docs/design/operation-plane-metadata.md
g2vinay: If we keep the incremental path, let's land the enforcement in this PR: a test that fails on `Unspecified` for Azure service commands, seeded with an explicit allowlist of the not-yet-classified tools. Deferring to step 4 leaves no forcing function, so `unspecified` will linger.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,125 @@
...
+| Any `Unspecified` | `Unspecified` |
+
+`NotApplicable` does not change an otherwise applicable aggregate. For example, `Data` plus `NotApplicable` aggregates to `Data`.
+
+Keeping aggregation in the consumer avoids duplicating the rule across the namespace, consolidated, proxy, and single-tool loading paths for a value none of them currently expose.
+
+## Initial rollout
+
+1. Add the enum, the attribute property, and the CLI JSON serialization.
+2. Classify Event Grid commands to validate data-, control-, and mixed-plane commands.
```
g2vinay: The three EventGrid commands land as `Both`, `Control`, `Control`, so there's no pure `Data` example. The case we most need to pin down, a data-plane action behind an ARM lookup, isn't actually exercised. Worth adding one.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,125 @@
...
+    "operationPlane": "control",
+    "destructive": { "value": false, "description": "..." },
+    "readOnly": { "value": true, "description": "..." }
+  }
+}
+```
+
+The behavioral flags use a `{ value, description }` shape because `"destructive": true` needs explanation. `"operationPlane": "control"` is self-describing, so it is emitted as a plain string rather than carrying a description that is a fixed function of the value.
+
+Stable serialized values are `unspecified`, `data`, `control`, `both`, and `notApplicable`. Unrecognized values, including any added by a future version, deserialize as `Unspecified` so older binaries keep reading newer metadata.
```
g2vinay, line 115: Once these string values ship in `tools list`, doc-gen and other consumers will depend on them, so a later reclassification (e.g. publish `Both` to `Data`) is effectively a contract change even though the JSON is additive. Suggest settling the plane rule before merging the serializer.

core/Microsoft.Mcp.Core/src/Commands/ToolMetadataJsonConverter.cs
```diff
@@ -14,6 +14,28 @@ namespace Microsoft.Mcp.Core.Commands;
...
+    /// Stable serialized values for <see cref="ToolOperationPlane"/>. Unknown values read as
+    /// <see cref="ToolOperationPlane.Unspecified"/> so newer metadata remains readable.
+    /// </summary>
+    private static string ToJsonValue(ToolOperationPlane operationPlane) => operationPlane switch
+    {
+        ToolOperationPlane.Data => "data",
+        ToolOperationPlane.Control => "control",
+        ToolOperationPlane.Both => "both",
+        ToolOperationPlane.NotApplicable => "notApplicable",
+        _ => "unspecified"
```
g2vinay: Minor: on write we always know our own enum, so a future value added without updating `ToJsonValue` would silently serialize as `unspecified`. Consider throwing in the write default while keeping the read side lenient.

core/Microsoft.Mcp.Core/src/Commands/ToolOperationPlane.cs
```diff
@@ -0,0 +1,25 @@
...
+    Unspecified,
+
+    /// <summary>The tool operates against an Azure service data-plane API.</summary>
+    Data,
+
+    /// <summary>The tool operates against an Azure management or control-plane API.</summary>
+    Control,
+
+    /// <summary>The tool operates against both Azure data-plane and control-plane APIs.</summary>
+    Both,
```
g2vinay, line 46: Do we need `Both` in v1? If we exclude incidental ARM lookups, genuinely dual-plane tools are rare, so it might be simpler to drop it until one appears, or tighten the doc comment to "performs a distinct user-meaningful operation on each plane."

servers/Azure.Mcp.Server/changelog-entries/kaghiya-operation-plane-metadata.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Features Added"
+    description: "Added Azure operation-plane metadata that distinguishes data-plane, control-plane, mixed, and non-Azure tools in `azmcp tools list` output."
```
g2vinay: Slightly overstated. Today only a few EventGrid/server commands are classified and everything else emits `unspecified`, so consider softening to note the classification is rolling out incrementally.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Agreed, and done. Every command is now classified in this PR — 372 tools, none `unspecified` (221 `control`, 111 `data`, 37 `notApplicable`, 3 `both`).

The incremental rollout language is gone from the design doc, and `CommandOperationPlaneTests` now fails the build on any unclassified command, with no allowlist, so the unclassified state can't come back.

docs/design/operation-plane-metadata.md
KarishmaGhiya: You were right, and this reshaped the design.

The rule is now: classify by the **deliverable** — the call that produces what the user asked for. Calls that only address or identify the target (subscription/tenant resolution, resource lookup to read an endpoint) do not count toward the plane. Otherwise, exactly as you said, nearly everything becomes `Both` and the label stops discriminating.

After applying it across all 372 tools, `both` came out to exactly 3.

tools/Azure.Mcp.Tools.EventGrid/src/Commands/Events/EventsPublishCommand.cs
KarishmaGhiya: Fixed — now `Data`.

I confirmed your reading of `EventGridService.PublishEventAsync`: `GetSubscription` → `FindTopic` → `topic.Data.Endpoint` → `EventGridPublisherClient`. ARM is purely addressing.

Added `Command_OperationPlane_IgnoresArmLookupUsedOnlyForAddressing` to pin it, so a future reclassification back to `Both` fails a test rather than passing silently.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Adopted as the governing rule.

The doc's Classification section is rewritten around the deliverable call, with an "Addressing does not count" subsection using `publish` as the worked example, and `Both` explicitly narrowed to two genuine user-facing deliverables across planes.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Enforcement landed: `CommandOperationPlaneTests.AllCommands_DeclareAnOperationPlane` walks every registered command and fails on `Unspecified`.

One deliberate departure: **no allowlist.** It was the right call for a partial rollout, but since everything is now classified, an allowlist would only be an escape hatch for new debt. I verified the test actually fails by temporarily unsetting one command's plane rather than trusting a green run.

Two companion tests guard against an undefined enum value and against a degenerate all-one-plane classification.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Resolved — `publish` is now that example, and it is exactly the case you named: a data-plane action behind an ARM lookup. Pinned by the test linked in the `EventsPublishCommand` thread.

Repo-wide there are now 111 `data` tools.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Agreed, and this is why the full classification is in this PR rather than a follow-up.

The rule is settled, every value is assigned, and the one reclassification you predicted (`publish` `Both` → `Data`) happened before anything shipped. Nothing here changes after merge without being a visible contract change.

core/Microsoft.Mcp.Core/src/Commands/ToolMetadataJsonConverter.cs
KarishmaGhiya: Done. `ToJsonValue` now enumerates all five values explicitly and throws `JsonException` on anything undefined; the read path stays lenient for forward compatibility.

Covered by `Serialize_UndefinedOperationPlane_Throws`.

core/Microsoft.Mcp.Core/src/Commands/ToolOperationPlane.cs
KarishmaGhiya: Kept, but tightened along the lines you suggested. The XML doc now reads "two genuine user-facing deliverables that fall on different planes" and states that a control-plane lookup supporting a data-plane deliverable is `Data`, not `Both`.

It earns its place, though narrowly: 3 of 372 tools use it — `CosmosListCommand`, `MySqlListCommand`, `PostgresListCommand`. Each branches on arguments to either enumerate ARM resources or list data-plane objects, so both are real user-facing deliverables. Without `Both` these would need an arbitrary pick or a tool split. Happy to revisit if you would rather split them.

servers/Azure.Mcp.Server/changelog-entries/kaghiya-operation-plane-metadata.yaml
KarishmaGhiya: You were right at the time.

Rather than soften it, the coverage caught up — the entry now names the field and values concretely, states the deliverable rule, and the "every tool is classified" claim is enforced by a test.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/PluginTelemetryCommand.cs
```diff
@@ -33,6 +33,7 @@ Publish plugin-related telemetry events from agent hooks.
         '--plugin-name', '--plugin-version', '--skill-name', '--skill-version', '--tool-name', and '--file-reference'. 
         Use this command from agent hooks in clients like VS Code, Claude Desktop, or Copilot CLI to emit usage metrics.
         """,
+    OperationPlane = ToolOperationPlane.NotApplicable,
```
alzimmermsft, line 36: Can be done in a follow-up PR given the size of this change, we may want to introduce an `Internal` operation to denote tools which are bound to the server itself.

core/Microsoft.Mcp.Core/src/Commands/CommandMetadataAttribute.cs
```diff
@@ -31,6 +31,9 @@ public sealed class CommandMetadataAttribute : Attribute
     /// </summary>
     public required string Title { get; init; }
 
+    /// <summary>The Azure API plane used by the tool. Default: unspecified.</summary>
+    public ToolOperationPlane OperationPlane { get; init; } = ToolOperationPlane.Unspecified;
```
alzimmermsft, line 35: Another long-term change, which I'd pair with a broader PR, we shouldn't have default values in `CommandMetadata`, everything should be explicitly required. Tool authors should be required to know this when creating them and having defaults may leave things ambiguous such as the default below with `Destructive = true` where incorrectly attributing that results in massive differences during runtime.

docs/design/operation-plane-metadata.md
alzimmermsft, line 1: Another long-term follow-up, we should create a broader design document for `CommandMetadata`, which will replace this.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+
+Related issue: [microsoft/mcp#1616](https://github.com/microsoft/mcp/issues/1616)
+
+## Status
+
+Proposed.
+
+## Context
+
+Azure MCP commands can call Azure management APIs, service data APIs, or both. The existing tool metadata describes behavioral characteristics such as whether a command is read-only or destructive, but it does not identify the Azure API plane used by a command.
```
alzimmermsft: nit: This metadata goes outside of just Azure MCP, not a big deal but we need to get out of the habit of having central documentation being specific to one MCP server or another as it makes it more confusing.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+
+The strongly typed option conversion moves command identity and behavioral hints to `CommandMetadataAttribute`. The `[Option]` attributes introduced by that conversion only describe command-line parameters and therefore do not solve operation-plane classification.
```
alzimmermsft: I'm not really understanding why we need to discuss `[Option]` here. Should just keep this scoped to `CommandMetadata` and why we needed to add this new feature.

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+    LocalRequired = false)]
+```
+
+### Classification
+
+A command's plane is determined by its **deliverable call**: the request that produces the result the user asked for.
+
+| Value | Meaning |
+|---|---|
+| `Data` | The deliverable is a workload call against a service data-plane API. |
```
alzimmermsft: Don't use deliverable, call them tools as that's what they are.


```suggestion
| `Data` | The tool performs its action(s) against a service data-plane API. |
```

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+| `Control` | The deliverable is a call against Azure Resource Manager or another management-plane API. |
+| `Both` | The command has two genuine user-facing deliverables, one on each plane. |
+| `NotApplicable` | The command does not perform an Azure service-plane operation, such as `tools list`, `server start`, or a local-only utility. |
+| `Unspecified` | The command has not been classified. This is an unset marker, not a valid answer. |
```
alzimmermsft: `command` -> `tool`

docs/design/operation-plane-metadata.md
```diff
@@ -0,0 +1,157 @@
...
+
+`NotApplicable` is a deliberate classification meaning "reviewed, and no Azure plane applies". `Unspecified` means "not yet reviewed" and fails validation for Azure service commands.
+
+#### Addressing does not count
```
alzimmermsft: Let's use `Setup` as `Addressing` is really confusing wording

docs/design/operation-plane-metadata.md
KarishmaGhiya: Is this applicable to other MCP servers in our repo like Fabric mcp?

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/PluginTelemetryCommand.cs
KarishmaGhiya: Agreed, and worth doing. I looked at what it would cover: 4 of the 47 `notApplicable` tools are server-bound — `PluginTelemetryCommand`, `ServerInfoCommand`, `ServerStartCommand`, `ToolsListCommand`. The other 43 are a different thing entirely (embedded docs, best practices, locally generated content), so today `notApplicable` is conflating "operates the server" with "returns content without calling anything." Splitting those is a real improvement.

Tracked in #3541. Leaving it out of this PR since it changes the public value set of the enum.

core/Microsoft.Mcp.Core/src/Commands/CommandMetadataAttribute.cs
KarishmaGhiya: Strongly agree, and `Destructive = true` is the right example of why: a tool author who omits it inherits a runtime-visible behavior they never chose, and the default is the dangerous value.

Worth noting `OperationPlane` is deliberately the inverse of that pattern — its default is an unset marker that fails a test rather than a plausible-looking value, which is the property you are describing for the rest. If the flags become required, `OperationPlane` can become required too and `CommandOperationPlaneTests` can retire.

Tracked in #3542. It is breaking for every `[CommandMetadata]` in the repo, so it wants its own PR.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Makes sense. This doc explains one field in isolation and never explains the metadata model it lives in, which is the more useful thing to write down.

Tracked in #3543, listing this doc as content to fold in and supersede.

docs/design/operation-plane-metadata.md
KarishmaGhiya: You are right, and to answer my own question above: yes, this applies beyond Azure MCP. `ToolOperationPlane` lives in `Microsoft.Mcp.Core`, and 38 Fabric OneLake tools are already classified by it.

Fixed in 4bc5bb92b — retitled to "Operation-Plane Metadata" and the framing is server-neutral. ARM stays as the dominant example rather than the definition.

4ab5fbb60 does the same for the `ToolOperationPlane` and `CommandMetadata.OperationPlane` doc comments, which had the same Azure-only phrasing and are in shared core.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Fair — I removed it. The Context section now only covers what tool metadata describes today and what it does not.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Taken, and applied further than the one row: "deliverable" is now gone from the doc, the `ToolOperationPlane` XML docs, the enforcement test's failure message, and the changelog entry.

Your `Data` row is used verbatim and the other four rows are parallel to it. The rule itself is unchanged — classify by the API the tool acts against to produce the result the user asked for — it just no longer needs a coined noun to state it.

docs/design/operation-plane-metadata.md
KarishmaGhiya: Done, and swept the whole doc. `Command` now appears only as a type name (`CommandMetadataAttribute`, `CommandOperationPlaneTests`).

docs/design/operation-plane-metadata.md
KarishmaGhiya: Agreed, "addressing" was jargon. Renamed to "Setup does not count" throughout, including the EventGrid regression test, now `Command_OperationPlane_IgnoresArmLookupUsedOnlyForSetup`.

servers/Azure.Mcp.Server/changelog-entries/kaghiya-operation-plane-metadata.yaml
```diff
@@ -0,0 +1,3 @@
+changes:
+  - section: "Features Added"
+    description: "Added an `operationPlane` field to tool metadata in `azmcp tools list` output, identifying each tool as `data`, `control`, `both`, or `notApplicable`. A tool's plane is the call that produces the requested result; Resource Manager calls made only to locate the target do not count toward it. Every tool is classified."
```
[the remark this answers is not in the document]
KarishmaGhiya: Filled in — the description now covers what the field is, the classification rule and why setup calls are excluded, why the whole repo is classified in one pass instead of migrating incrementally, and the serialization behaviour. Linked to #1616 and to the three follow-ups (#3541, #3542, #3543).

docs/design/operation-plane-metadata.md
KarishmaGhiya: Following up, because my earlier reply answered the migration question but not the one you actually anchored to — why the value is still in the enum at all.

The migration state is gone: every registered tool is classified, the incremental-rollout language is out of the doc, and `AllCommands_DeclareAnOperationPlane` fails the build on `Unspecified` with no allowlist.

What is left is the sentinel, and it survives only because it is the zero value. `Unspecified` is first, so a tool that omits `OperationPlane` gets it by default. If it were removed, `Data` would become `0`, and an author who forgot the field would ship a tool silently labelled data-plane — a wrong answer instead of a detectable one. It is also the fallback that lets an older binary read an `operationPlane` value added by a newer one without failing.

That trade-off disappears entirely once `OperationPlane` is `required` on `CommandMetadataAttribute`, which is the change you asked for in the comment below and which I tracked as #3542. At that point the unset state is unrepresentable and the sentinel is dead weight, so I have added removing it as explicit acceptance criteria on that issue. Keeping the two changes together avoids a window where omitting the field is both possible and silent.

Happy to pull it forward into this PR instead if you would rather not carry it, but it would mean making the property `required` here, which touches every `[CommandMetadata]` in the repo and is the broader change you suggested keeping separate.

core/Microsoft.Mcp.Core/src/Commands/ToolOperationPlane.cs
```diff
@@ -0,0 +1,51 @@
...
+    Data,
+
+    /// <summary>
+    /// The tool performs its action against Azure Resource Manager or another management-plane API,
+    /// such as listing or creating resources.
+    /// </summary>
+    Control,
+
+    /// <summary>
+    /// The tool performs two distinct user-facing actions that fall on different planes, such as
```
g2vinay: `Both` is defined as performing both actions in one invocation, but Cosmos, MySQL, and PostgreSQL list commands select only one plane through mutually exclusive branches. Please redefine `Both` as “may use either plane depending on the invocation,” or split/reclassify those tools before this becomes a stable JSON contract.

tools/Azure.Mcp.Tools.Extension/src/Commands/AzqrCommand.cs
```diff
@@ -20,6 +20,7 @@ namespace Azure.Mcp.Tools.Extension.Commands;
     Name = "azqr",
     Title = "Azure Quick Review CLI Command",
     Description = "Runs Azure Quick Review CLI (azqr) to scan an Azure subscription (or resource group) for compliance issues and provide compliance recommendations. Generates a compliance and security assessment report that identifies non-compliant configurations and recommends improvements for your Azure resources. Use this whenever a user wants to scan, check, review, or assess a subscription for compliance issues or compliance recommendations, or wants recommendations to fix compliance and security problems. Requires a subscription (ID or name) and optionally a resource group. Returns the generated report file paths (XLSX and JSON). Note: azqr performs compliance and security scans and is different from Azure CLI (az), from Azure Policy assignments, and from Azure Advisor recommendations for cost, security, reliability, operational excellence, and performance improvements.",
+    OperationPlane = ToolOperationPlane.NotApplicable,
```
g2vinay: `NotApplicable` means no service-plane operation, but this command runs an Azure subscription scan and returns its results. Calling Azure indirectly through `azqr` should not erase the operation plane. Please classify this as `Control` or `Both` based on the APIs used by the scan.

g2vinay (changes requested):
Requesting changes for two contract-level issues: the documented meaning of `Both` conflicts with every current use, and `AzqrCommand` is classified as `NotApplicable` despite performing the requested Azure subscription scan.

core/Microsoft.Mcp.Core/src/Commands/ToolOperationPlane.cs
KarishmaGhiya: Agreed, and the definition was wrong. `cosmos list`, `mysql list`, and `postgres list` are the only three uses of `Both` in the repo, and all three dispatch to exactly one plane per invocation — so as written, `Both` described a case that never occurs while the actual usage was undocumented.

I went with redefining rather than splitting. Reclassifying to a single plane would lose information a consumer needs: a policy that permits only data-plane tools has to reject `cosmos list`, because it cannot know at authorization time whether the call will carry `--account`. Splitting into separate tools would fix the metadata by changing the tool surface, which does not seem appropriate for this situation.

`Both` now reads "the tool can act on either plane, so a single plane cannot be inferred from the tool alone", and the design doc spells out both qualifying shapes with `cosmos list` as the worked example. I also removed the "expected to be rare" claim, which was false. The setup-exclusion rule is unchanged.

Fixed in f335f4aad.

tools/Azure.Mcp.Tools.Extension/src/Commands/AzqrCommand.cs
KarishmaGhiya: Agreed, fixed — this is now `Control`. `AzqrCommand` runs `azqr scan --subscription-id`, and that scan reads the subscription through ARM, so the plane is real regardless of the process boundary.

I checked the rest of the toolset: `cli generate` and `cli install` call no service at all, so they remain `NotApplicable`. I've added an explicit rule to the design doc — classify by the API the work ultimately reaches, not by how the tool reaches it — so the next external-CLI wrapper doesn't repeat this.

Fixed in f335f4aad.

alzimmermsft (approved):

alzimmermsft (approved):

g2vinay (approved):

PR #3426, opened 2026-08-28
Capture telemetry-specific command failure messages

### LINKED ISSUE #3395: Capture a telemetry-specific message for tool failures

tmeschter:
## What does this PR do?
Adds an opt-in `TelemetryFailureMessage` to `CommandResponse` so individual commands can provide a sanitized, telemetry-specific message when a tool call fails.

The property is excluded from client-visible serialization. For non-2xx responses, the shared command execution path records the value as `ToolFailureMessage`; successful responses ignore it. This keeps failed command results distinct from `exception.*` telemetry, which remains reserved for thrown exceptions. The documentation recommends static messages so telemetry values are easy to group and analyze.

Added focused tests for failure capture, success suppression, blank messages, coexistence with exception telemetry, and serialization exclusion. Added Azure and Fabric changelog entries.

## GitHub issue number?
Fixes https://github.com/microsoft/mcp/issues/3395

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes: Not applicable; this changes shared command telemetry infrastructure without adding or modifying a tool.
- [ ] Extra steps for **Azure MCP Server** tool changes: Not applicable; this does not change an Azure MCP tool.

## Validation
- `dotnet test core/Microsoft.Mcp.Core/tests/Microsoft.Mcp.Core.Tests/Microsoft.Mcp.Core.Tests.csproj -- --filter-class Microsoft.Mcp.Core.Tests.Commands.CommandTelemetryTests`
- `dotnet build core/Microsoft.Mcp.Core/src/Microsoft.Mcp.Core.csproj --no-restore --nologo`
- Changed-file cspell validation
- Full local build and npm packaging, followed by smoke tests of the generated Azure, Fabric, and Template npm wrappers

## Invoking Livetests

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
+++ b/core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
@@ -118,6 +118,7 @@ public abstract class BaseCommand<[DynamicallyAccessedMembers(TrimAnnotations.Co
         }
 
         CommandResponse response = await ExecuteAsync(context, options, cancellationToken);
+        CaptureTelemetryFailureMessage(context, response);
         return response;
     }
 
@@ -135,6 +136,20 @@ public abstract class BaseCommand<[DynamicallyAccessedMembers(TrimAnnotations.Co
                 $"Command type '{GetType().FullName}' called {nameof(SetResult)} without overriding {nameof(ResultTypeInfo)}.");
 
         context.Response.Results = ResponseResult.Create(result, typeInfo);
+    }
+
+    /// <summary>
+    /// Adds a command-provided, sanitized message to telemetry for failed responses.
+    /// </summary>
+    /// <param name="context">The command execution context containing the telemetry activity.</param>
+    /// <param name="response">The command response containing the optional telemetry message.</param>
+    private static void CaptureTelemetryFailureMessage(CommandContext context, CommandResponse response)
+    {
+        var isError = response.Status < HttpStatusCode.OK || response.Status >= HttpStatusCode.Ambiguous;
+        if (isError && !string.IsNullOrWhiteSpace(response.TelemetryFailureMessage))
+        {
+            context.Activity?.SetTag(TagName.ExceptionMessage, response.TelemetryFailureMessage);
+        }
     }
 
     protected virtual void HandleException(CommandContext context, Exception ex)
--- a/core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
+++ b/core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
@@ -148,7 +148,7 @@ public abstract class BaseCommand<[DynamicallyAccessedMembers(TrimAnnotations.Co
         var isError = response.Status < HttpStatusCode.OK || response.Status >= HttpStatusCode.Ambiguous;
         if (isError && !string.IsNullOrWhiteSpace(response.TelemetryFailureMessage))
         {
-            context.Activity?.SetTag(TagName.ExceptionMessage, response.TelemetryFailureMessage);
+            context.Activity?.SetTag(TagName.ToolFailureMessage, response.TelemetryFailureMessage);
         }
     }
 
--- a/core/Microsoft.Mcp.Core/src/Commands/TelemetryConstants.cs
+++ b/core/Microsoft.Mcp.Core/src/Commands/TelemetryConstants.cs
@@ -69,6 +69,11 @@ public class TagName
     public const string ToolName = "ToolName";
 
     /// <summary>
+    /// A sanitized message describing why an MCP tool returned a failed result.
+    /// </summary>
+    public const string ToolFailureMessage = "ToolFailureMessage";
+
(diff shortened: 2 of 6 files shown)


core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
```diff
@@ -107,9 +107,24 @@ async Task<CommandResponse> IBaseCommand.ExecuteAsync(CommandContext context, Pa
...
+    /// Adds a command-provided, sanitized message to telemetry for failed responses.
+    /// </summary>
+    /// <param name="context">The command execution context containing the telemetry activity.</param>
+    /// <param name="response">The command response containing the optional telemetry message.</param>
+    private static void CaptureTelemetryFailureMessage(CommandContext context, CommandResponse response)
+    {
+        var isError = response.Status < HttpStatusCode.OK || response.Status >= HttpStatusCode.Ambiguous;
+        if (isError && !string.IsNullOrWhiteSpace(response.TelemetryFailureMessage))
+        {
+            context.Activity?.SetTag(TagName.ExceptionMessage, response.TelemetryFailureMessage);
```
alzimmermsft: Any chance that `TagName.ExceptionMessage` would have already been set when we reach this point? If so, may want to use `SetTagIfNotExists`

alzimmermsft (approved):

core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
vukelich: Alternatively, use a different tag name so that we can retain both. We can use [coalesce()](https://learn.microsoft.com/en-us/kusto/query/coalesce-function?view=azure-data-explorer) in KQL pretty easily on multiple properties. When troubleshooting issues, it's incredibly helpful to be confident of when a telemetry data point is being populated to backtrack where an error happened. Multiple places for a data point to be written allows for uncertainty and incorrect conclusions on the flow of code execution.

vukelich: Additionally, our data classification tools are often property-based. Reusing the same property name prevents granular classification and prevents isolation between properties if something needs to be nuked.

vukelich: In general, I'd also strongly recommend any properties named with `*Exception*` be strictly for thrown `Exception`s in the code. This allows for us to differentiate between "what was thrown?" from "what is the cause and outcome of a failed request?" `Exceptions` are symptoms and signals of a failure, but a failure is a broader expression of an operation's result.

tmeschter: Updated this to use a dedicated `ToolFailureMessage` tag instead of `exception.message`. Existing exception telemetry is now preserved independently, so thrown exceptions and failed command results have distinct provenance and classification. The tests now verify that both values coexist without overwriting.

core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
```diff
@@ -137,6 +138,20 @@ protected void SetResult(CommandContext context, TResult result)
...
 
+    /// <summary>
+    /// Adds a command-provided, sanitized message to telemetry for failed responses.
+    /// </summary>
+    /// <param name="context">The command execution context containing the telemetry activity.</param>
+    /// <param name="response">The command response containing the optional telemetry message.</param>
+    private static void CaptureTelemetryFailureMessage(CommandContext context, CommandResponse response)
+    {
+        var isError = response.Status < HttpStatusCode.OK || response.Status >= HttpStatusCode.Ambiguous;
+        if (isError && !string.IsNullOrWhiteSpace(response.TelemetryFailureMessage))
```
vukelich, line 149: Optional: why check the value of `response.Status`? When would `TelemetryFailureMessage` be populated that could be ignored?

My question gets into the meme of "task failed successfully." Bugs can be written that misattribute errors as success, like code that catches an `Exception` but accidentally omits rethrowing it. Always logging `TelemetryFailureMessage` could identify such a situation.

The alternative is that we better construct instances of `CommandResponse` to prevent mismatching and inconsistent data. For example:
- Make the constructor private and introduce `public static` methods of `CommandResponse.FromSuccess(...)`, `CommandResponse.FromException(....`)`, `CommandResponse.FromNonExceptionFailure(....)`, etc. This can help the code author or agent write the correct code based on well named methods.
- Add validation logic within the constructor to throw if properties contradict each other. This is probably good regardless, but without other ease-of-use APIs, then misuse will only be found at runtime.

vukelich (approved):
Optional comment.

core/Microsoft.Mcp.Core/src/Commands/BaseCommand`2.cs
tmeschter: I've captured this in #3504 as a follow-up.

AmosHersch (approved):

PR #3433, opened 2026-08-31
Add BestPrice remote MCP server

phaistonian:
## What does this PR do?

Adds BestPrice Shopping as a public remote MCP endpoint to the Azure MCP external-server registry.

- Endpoint: https://mcp.bestprice.gr/mcp
- Transport: Streamable HTTP
- Authentication: none
- Scope: read-only Greek shopping decisions, product search, offer comparison, and price history
- Validation: endpoint initializes successfully with MCP protocol version 2025-06-18 and lists four read-only tools.

The entry uses the unique bestprice_ tool prefix to avoid collisions.


phaistonian:
@microsoft-github-policy-service agree company="Phaistos Networks S.A."

phaistonian:
Thank you for flagging this. We agree a third-party remote MCP must not be active by default or create an implicit outbound dependency. Does Azure MCP support a registry-level opt-in / explicit-enable mechanism for external servers? If not, we will withdraw this default-registry entry rather than introduce that behavior.

PR #3434, opened 2026-08-31
Honor the elicitation decision field when evaluating approval

g2vinay:
HandleElicitationAsync evaluated only the transport-level envelope action and ignored the user's selection in the elicitation form's required 'decision' field. A client that submits the form returns action == 'accept' even when the user selects 'Reject', so the operation ran anyway.

Require both the envelope action and the decision content to equal 'accept', reading the value via a new TryGetElicitationDecision helper. Add regression tests for the reject-with-accept-envelope and missing-decision cases.

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
@@ -19,6 +19,10 @@ namespace Microsoft.Mcp.Core.Areas.Server.Commands.ToolLoading;
 /// <param name="logger">Logger instance for this tool loader.</param>
 public abstract class BaseToolLoader(ILogger logger) : IToolLoader
 {
+    private const string ElicitationDecisionPropertyName = "decision";
+    private const string ElicitationAcceptDecision = "accept";
+    private const string ElicitationRejectDecision = "reject";
+
     /// <summary>
     /// Logger instance for this tool loader.
     /// </summary>
@@ -271,18 +275,18 @@ public abstract class BaseToolLoader(ILogger logger) : IToolLoader
                 {
                     Properties = new Dictionary<string, ElicitRequestParams.PrimitiveSchemaDefinition>
                     {
-                        ["decision"] = new ElicitRequestParams.TitledSingleSelectEnumSchema
+                        [ElicitationDecisionPropertyName] = new ElicitRequestParams.TitledSingleSelectEnumSchema
                         {
                             Title = "Decision",
                             Description = "Approve or reject this sensitive operation.",
                             OneOf = new List<ElicitRequestParams.EnumSchemaOption>
                             {
-                                new() { Title = "Approve", Const = "accept" },
-                                new() { Title = "Reject", Const = "reject" }
+                                new() { Title = "Approve", Const = ElicitationAcceptDecision },
+                                new() { Title = "Reject", Const = ElicitationRejectDecision }
                             }
                         }
                     },
-                    Required = ["decision"]
+                    Required = [ElicitationDecisionPropertyName]
                 }
             };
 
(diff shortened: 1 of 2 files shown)


RickWinter (approved):

JasonYeMSFT (approved):

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
```diff
@@ -312,4 +324,27 @@ protected internal static McpClientOptions CreateClientOptions(McpServer server)
...
+    /// The elicitation schema declares a required "decision" field whose value is either
+    /// "accept" or "reject". This value—not the transport envelope action—represents the user's
+    /// selection when a client submits the form, so it must be inspected to honor a rejection.
+    /// </summary>
+    /// <param name="content">The content payload returned in the elicitation response.</param>
+    /// <returns>
+    /// The decision string (e.g. "accept" or "reject") when present as a string value; otherwise
+    /// <c>null</c>.
+    /// </returns>
+    private static string? TryGetElicitationDecision(IDictionary<string, JsonElement>? content)
```
alzimmermsft: This doesn't follow .NET `Try*` method patterns where they return a `bool` and `out` a value.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
```diff
@@ -288,13 +288,25 @@ protected internal static McpClientOptions CreateClientOptions(McpServer server)
...
+            // Determine approval from BOTH the transport-level envelope action and the user's
+            // selection carried in the response content. The elicitation schema declares a
+            // required "decision" field, so the envelope action alone is not sufficient: a
+            // client that submits the form returns Action == "accept" even when the user picked
+            // "Reject" (their selection lives in Content["decision"]). Require the envelope
+            // action to be "accept" AND the decision value to be "accept"; otherwise treat the
+            // operation as not approved and do not execute it.
+            bool envelopeAccepted = string.Equals(protocolResponse.Action, "accept", StringComparison.Ordinal);
+            string? decision = TryGetElicitationDecision(protocolResponse.Content);
+            bool approved = envelopeAccepted && string.Equals(decision, "accept", StringComparison.Ordinal);
```
alzimmermsft: Let's turn `accept` into a shared value somewhere in this class given we're defining it as the constant value for the elicitation request then checking against it here. This a liable to skew and become incorrect if we change one spot but not the other.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
```diff
@@ -288,13 +288,25 @@ protected internal static McpClientOptions CreateClientOptions(McpServer server)
...
 
-            if (protocolResponse.Action != "accept")
+            // Determine approval from BOTH the transport-level envelope action and the user's
+            // selection carried in the response content. The elicitation schema declares a
+            // required "decision" field, so the envelope action alone is not sufficient: a
+            // client that submits the form returns Action == "accept" even when the user picked
+            // "Reject" (their selection lives in Content["decision"]). Require the envelope
+            // action to be "accept" AND the decision value to be "accept"; otherwise treat the
+            // operation as not approved and do not execute it.
+            bool envelopeAccepted = string.Equals(protocolResponse.Action, "accept", StringComparison.Ordinal);
```
alzimmermsft: Is this `accept` value separate from the constant value we supply in the elicitation request?

There's also a property on `ElicitResult` that's more appropriate than string comparison on `Action`, which is `IsAccepted` which we should use here instead as that pushes the logic into ModelContextProtocol.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
g2vinay: Addressed in 583f21afa. `TryGetElicitationDecision` now follows the .NET `Try*` pattern by returning `bool` and providing the parsed value through an `out string?` parameter.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
g2vinay: Addressed in 583f21afa. I added shared constants for the application-defined `decision` property and its `accept`/`reject` values, and use them in both the request schema and response validation.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
g2vinay: Addressed in 583f21afa. The envelope check now uses `protocolResponse.IsAccepted`. The envelope action remains SDK-owned, while the separate application-defined decision value uses the shared local constant.

alzimmermsft (approved):

PR #3579, opened 2026-09-04
fix(authorization): return role assignments for management group scopes

### LINKED ISSUE #3285: [BUG] role_assignment_list returns empty with status 200 for management-group scopes
### Describe the bug

`role_assignment_list` returns an empty assignment array with
`"status": 200, "message": "Success"` for a management-group scope that
demonstrably has role assignments. The failure is silent — it is
indistinguishable from a genuine empty result.

### Expected behavior

The assignments present at that scope — in our case Reader (`acdd72a7-…`)
and Cost Management Reader (`72fafb9e-…`) on the calling SP, plus four user
Owner assignments.

### Actual behavior

The assignments present at that scope — in our case Reader (`acdd72a7-…`)
and Cost Management Reader (`72fafb9e-…`) on the calling SP, plus four user
Owner assignments.

### Verification that the result is wrong

Two independent paths, same credential, same session:

1. `arm` / `execute_query` against `authorizationresources` returns six
   management-group-scoped assignments, including the two on the calling SP.
2. `az role assignment list --all --assignee <appId> --include-inherited
   --include-groups` under a separate identity returns the same two.

Reader includes `Microsoft.Authorization/roleAssignments/read`, so this is
not a permissions gap — and a permissions gap should surface as an error,
not as `Success`.

### Why this matters beyond the empty array

The response echoes back no scope, so a caller cannot tell which scope was
actually queried. For an agent consuming this tool, a false empty is worse
than an error: it reads as positive evidence of absence. In our case it
produced a confident, well-formed, entirely wrong conclusion that a
production RBAC assignment was missing.

### Reproduction Steps

1. Assign Reader at `/providers/Microsoft.Management/managementGroups/<mg-id>`
   to a service principal.
2. Configure the MCP server with that SP's credentials.
3. Call `role_assignment_list` with:
   `{"scope": "/providers/Microsoft.Management/managementGroups/<mg-id>"}`

### Environment

- Azure MCP Server 2.0 via `npx -y @azure/mcp@latest server start`
- Client: Claude Code (VS Code extension), macOS
- Credential: service principal via `AZURE_TENANT_ID` / `AZURE_CLIENT_ID` /
  `AZURE_CLIENT_SECRET`, holding Reader + Cost Management Reader at the
  management group being queried

KarishmaGhiya:
## Description

`role_assignment_list` returned an empty array with a `200` success response for any management group scope, so callers could not tell "no assignments" from "not queried".

### Root cause

`BaseAzureResourceService.ExecuteResourceQueryAsync` built every Resource Graph query with only `Subscriptions` populated on `ResourceQueryContent`, never `ManagementGroups` (`core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs:115-118` before this change). Management group role assignments have IDs of the form:

```
/providers/Microsoft.Management/managementGroups/<mg>/providers/Microsoft.Authorization/roleAssignments/<guid>
```

They live outside any subscription, so Resource Graph excluded them from the result set before the scope filter was ever applied. The query itself succeeded, which is why the tool reported success with zero rows.

### Changes

1. **Management group scoping.** Added `ExecuteManagementGroupResourceQueryAsync<T>` to `BaseAzureResourceService`, which populates `ManagementGroups` instead of `Subscriptions`. `AuthorizationService` routes management group scopes to it. A private `ResolveTenantResourceAsync` obtains the `TenantResource` without needing a subscription.
2. **Anchored scope filter.** The previous filter was `id contains '<scope>'`, an unanchored substring match, so requesting `.../resourceGroups/rg1` also returned assignments from `rg10`. It is now `(properties.scope =~ '<scope>' or properties.scope startswith '<scope>/')`. Both KQL operators are case-insensitive, which matters because Resource Graph returns `resourcegroups` in lowercase.
3. **Subscription handling is scope-aware.** Management groups sit outside subscriptions, so the shared `SubscriptionCommand` now has an explicit applicability hook used by both subscription resolution and validation. Management group scopes skip ambient subscription resolution and reject an explicitly supplied `--subscription`; every other supported scope still requires one.
4. **The requested scope is echoed in the result**, so an empty list is attributable to a specific scope. This was explicitly requested in the issue.

### Semantics

Results remain **at and below** the requested scope. Assignments inherited from a parent scope are **not** included, and this PR does not add an `--include-inherited` option. This is now stated in the tool description, in `azmcp-commands.md`, and in the option help so the behavior is no longer implicit.

### Not addressed

The issue also notes that an insufficient-permissions result should surface an error rather than an empty success. Resource Graph returns zero rows for both "nothing there" and "you cannot see it", so the two are not distinguishable at this layer. Left as a separate concern.

## Testing

- `Azure.Mcp.Tools.Authorization.Tests` — 15/15 pass, including the recorded `Should_list_role_assignments` integration test.
- Six new unit tests cover: a management group scope with no subscription or ambient subscription resolution, rejection of an explicit subscription for management group scopes at both command and service layers, the scope being echoed on an empty result, `--subscription` still being required for non-management-group scopes, and a six-case theory over `ManagementGroupScope.TryParse` (casing, trailing slash, empty remainder, nested provider path).
- The existing recording was re-pushed (`Azure.Mcp.Tools.Authorization.Tests_a8cd50f9a5`) because the anchored filter changes the request body the test proxy matches on. The recorded response contains a single assignment whose scope is exactly the test resource group, so both the old and new queries select the identical row — the response payload is unchanged.
- Full `dotnet build` succeeds; `dotnet format --verify-no-changes` and `Invoke-Cspell.ps1` are clean.

**Known limitation:** there is no recorded or live test for the management group path. Recording one requires a real management group, which the tenant-level permissions in the existing `test-resources.bicep` cannot provision.

## Pre-merge checklist

- [x] Unit tests added for the new behavior
- [x] `servers/Azure.Mcp.Server/docs/azmcp-commands.md` updated
- [x] `servers/Azure.Mcp.Server/docs/e2eTestPrompts.md` updated with two new prompts
- [x] Changelog entry added under `servers/Azure.Mcp.Server/changelog-entries/`
- [x] Spell check clean
- [x] `dotnet format` clean
- [x] Recordings updated and pushed; `assets.json` `Tag` bumped
- [n/a] `consolidated-tools.json` — no tool is added, renamed, or removed
- [x] **ToolDescriptionEvaluator score** — the tool description changed, so a score is required. 

Evaluator results ( text-embedding-3-large ) — new description beats the original on every prompt that routes:

┌───────────────────────────────────────────────┬────────────┬────────────┐
│ Prompt                                        │ Original   │ New        │
├───────────────────────────────────────────────┼────────────┼────────────┤
│ list role assignments for my management group │ #2, 0.5477 │ #1, 0.6406 │
├───────────────────────────────────────────────┼────────────┼────────────┤
│ show rbac role assignments at this scope      │ #1, 0.5920 │ #1, 0.6068 │
└───────────────────────────────────────────────┴────────────┴────────────┘

Both rank #1, comfortably above the 0.4 threshold. Checkbox ticked in the PR body with the numbers.

One incidental finding: `Who has access to resource group X?` does not surface `role_assignment_list` in the top 10 — for the old description and the new one alike, so it is a pre-existing discovery gap rather than something this PR caused.

**Update following review feedback.** `role_assignment_list` exposes only `--scope`, `--subscription` and `--tenant`, so the two prompts originally added to `e2eTestPrompts.md` wrongly implied `management-group` and `resource-group` parameters. Both now name `--scope` and give the value as an explicit ARM scope path:

- `List the role assignments at scope /providers/Microsoft.Management/managementGroups/<management-group>`
- `List the role assignments at scope /subscriptions/<subscription>/resourceGroups/<resource-group>`

The tool description has also had the inaccurate `, returned as a JSON array` clause removed. That is a pure deletion, so the text measured in the table above still stands and no re-score was required.

A later review also led to an explicit subscription-applicability hook in `SubscriptionCommand`, clearer `<null>`/`<empty>` scope logging, and a `[NotNull]` single-parameter validation helper that removes the null-forgiveness operator. Azure Resource Graph supports tenant-wide queries by omitting both subscription and management-group scopes, but `/` is not added in this focused management-group fix because tenant-wide semantics, permissions, and testing need separate design work.

Fixes #3285

## Invoking Livetests

### DIFF
--- a/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
+++ b/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
@@ -14,11 +14,17 @@ public abstract class SubscriptionCommand<
 {
     private readonly ISubscriptionResolver _subscriptionResolver = subscriptionResolver;
 
+    /// <summary>
+    /// Determines whether subscription resolution and validation apply to the bound options.
+    /// </summary>
+    /// <remarks>This method is evaluated during post-binding and must only inspect already-bound option values.</remarks>
+    protected virtual bool IsSubscriptionApplicable(TOptions options) => true;
+
     public override void ValidateOptions(TOptions options, ValidationResult validationResult)
     {
         base.ValidateOptions(options, validationResult);
 
-        if (string.IsNullOrEmpty(options.Subscription))
+        if (IsSubscriptionApplicable(options) && string.IsNullOrEmpty(options.Subscription))
         {
             validationResult.Errors.Add("Missing Required options: --subscription");
         }
@@ -26,6 +32,13 @@ public abstract class SubscriptionCommand<
 
     public override void PostBindOptions(TOptions options)
     {
+        base.PostBindOptions(options);
+
+        if (!IsSubscriptionApplicable(options))
+        {
+            return;
+        }
+
         // Always post-process subscription via resolver (env var / CLI profile fallback)
         options.Subscription = _subscriptionResolver.ResolveSubscription(options.Subscription);
         if (!string.IsNullOrEmpty(options.Subscription))
--- a/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
+++ b/core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
@@ -14,17 +14,11 @@ public abstract class SubscriptionCommand<
 {
     private readonly ISubscriptionResolver _subscriptionResolver = subscriptionResolver;
 
-    /// <summary>
-    /// Determines whether subscription resolution and validation apply to the bound options.
-    /// </summary>
-    /// <remarks>This method is evaluated during post-binding and must only inspect already-bound option values.</remarks>
-    protected virtual bool IsSubscriptionApplicable(TOptions options) => true;
-
     public override void ValidateOptions(TOptions options, ValidationResult validationResult)
     {
         base.ValidateOptions(options, validationResult);
 
-        if (IsSubscriptionApplicable(options) && string.IsNullOrEmpty(options.Subscription))
+        if (string.IsNullOrEmpty(options.Subscription))
         {
             validationResult.Errors.Add("Missing Required options: --subscription");
         }
@@ -32,13 +26,6 @@ public abstract class SubscriptionCommand<
 
     public override void PostBindOptions(TOptions options)
     {
-        base.PostBindOptions(options);
-
-        if (!IsSubscriptionApplicable(options))
-        {
-            return;
-        }
-
         // Always post-process subscription via resolver (env var / CLI profile fallback)
         options.Subscription = _subscriptionResolver.ResolveSubscription(options.Subscription);
         if (!string.IsNullOrEmpty(options.Subscription))
--- a/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
@@ -88,29 +88,12 @@ public abstract class BaseAzureResourceService(IAzureService azureService)
     {
         ValidateRequiredParameters((nameof(resourceType), resourceType), (nameof(subscription), subscription));
         ArgumentNullException.ThrowIfNull(converter);
-
-        if (!string.IsNullOrEmpty(additionalFilter) && additionalFilter.Contains('|'))
-        {
-            throw new ArgumentException(
-                "additionalFilter must not contain the pipe operator '|' to prevent KQL injection.",
-                nameof(additionalFilter));
-        }
-
-        var results = new List<T>();
+        ValidateAdditionalFilter(additionalFilter);
 
         var subscriptionResource = await AzureService.GetSubscription(subscription, tenant, cancellationToken);
         var tenantResource = await GetTenantResourceAsync(subscriptionResource!.Data.TenantId, cancellationToken);
 
-        var queryFilter = $"{tableName} | where type =~ '{EscapeKqlString(resourceType)}'";
-        if (!string.IsNullOrEmpty(resourceGroup))
-        {
-            queryFilter += $" and resourceGroup =~ '{EscapeKqlString(resourceGroup)}'";
-        }
-        if (!string.IsNullOrEmpty(additionalFilter))
-        {
-            queryFilter += $" and {additionalFilter}";
-        }
-        queryFilter += $" | limit {limit}";
+        var queryFilter = BuildResourceQuery(tableName, resourceType, resourceGroup, additionalFilter, limit);
 
         var queryContent = new ResourceQueryContent(queryFilter)
         {
@@ -118,19 +101,9 @@ public abstract class BaseAzureResourceService(IAzureService azureService)
         };
 
         ResourceQueryResult result = await tenantResource.GetResourcesAsync(queryContent, cancellationToken);
-        if (result != null && result.Count > 0)
-        {
-            using var jsonDocument = JsonDocument.Parse(result.Data);
-            var dataArray = jsonDocument.RootElement;
-            if (dataArray.ValueKind == JsonValueKind.Array)
-            {
-                foreach (var item in dataArray.EnumerateArray())
-                {
-                    results.Add(converter(item));
-                }
-            }
-        }
-        else if (!string.IsNullOrEmpty(resourceGroup))
+        var results = MaterializeResults(result, converter);
+
+        if ((result == null || result.Count == 0) && !string.IsNullOrEmpty(resourceGroup))
         {
             // If the query returned no results and a resource group filter was applied, validate that the resource group exists to provide better error handling
             if (!await ValidateResourceGroupExistsAsync(subscriptionResource, resourceGroup, cancellationToken))
@@ -140,6 +113,131 @@ public abstract class BaseAzureResourceService(IAzureService azureService)
         }
 
         return new ResourceQueryResults<T>(results, result?.ResultTruncated == ResultTruncated.True);
+    }
+
+    /// <summary>
+    /// Executes a Resource Graph query scoped to a management group and returns a list of resources of the specified type.
+    /// </summary>
+    /// <remarks>
+    /// Resource Graph evaluates a query only within the scopes it is given. A subscription-scoped query therefore
+    /// cannot see resources whose IDs live outside any subscription, such as role assignments made directly on a
+    /// management group. Use this overload when the caller asked about a management group scope.
+    /// </remarks>
+    /// <typeparam name="T">The type to convert each resource to</typeparam>
+    /// <param name="resourceType">The Azure resource type to query for</param>
+    /// <param name="managementGroup">The management group ID to scope the query to</param>
+    /// <param name="converter">Function to convert JsonElement to the target type</param>
+    /// <param name="tableName">Optional table name to query (default: "resources")</param>
+    /// <param name="additionalFilter">Optional additional KQL filter condition</param>
+    /// <param name="limit">Maximum number of results to return (default: 50)</param>
+    /// <param name="tenant">Optional tenant to use for the query</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>List of resources converted to the specified type</returns>
+    protected async Task<ResourceQueryResults<T>> ExecuteManagementGroupResourceQueryAsync<T>(
+        string resourceType,
+        string managementGroup,
+        Func<JsonElement, T> converter,
+        string? tableName = "resources",
+        string? additionalFilter = null,
+        int limit = 50,
+        string? tenant = null,
+        CancellationToken cancellationToken = default)
+    {
+        ValidateRequiredParameters((nameof(resourceType), resourceType), (nameof(managementGroup), managementGroup));
+        ArgumentNullException.ThrowIfNull(converter);
+        ValidateAdditionalFilter(additionalFilter);
+
+        var tenantResource = await ResolveTenantResourceAsync(tenant, cancellationToken);
+
+        var queryFilter = BuildResourceQuery(tableName, resourceType, resourceGroup: null, additionalFilter, limit);
+
+        var queryContent = new ResourceQueryContent(queryFilter)
+        {
+            ManagementGroups = { managementGroup }
+        };
+
+        ResourceQueryResult result = await tenantResource.GetResourcesAsync(queryContent, cancellationToken);
+        var results = MaterializeResults(result, converter);
+
+        return new ResourceQueryResults<T>(results, result?.ResultTruncated == ResultTruncated.True);
+    }
+
+    private static void ValidateAdditionalFilter(string? additionalFilter)
+    {
+        if (!string.IsNullOrEmpty(additionalFilter) && additionalFilter.Contains('|'))
+        {
+            throw new ArgumentException(
+                "additionalFilter must not contain the pipe operator '|' to prevent KQL injection.",
+                nameof(additionalFilter));
+        }
+    }
+
+    private static string BuildResourceQuery(
+        string? tableName,
+        string resourceType,
+        string? resourceGroup,
+        string? additionalFilter,
+        int limit)
+    {
+        var queryFilter = $"{tableName} | where type =~ '{EscapeKqlString(resourceType)}'";
+        if (!string.IsNullOrEmpty(resourceGroup))
+        {
+            queryFilter += $" and resourceGroup =~ '{EscapeKqlString(resourceGroup)}'";
+        }
+        if (!string.IsNullOrEmpty(additionalFilter))
+        {
+            queryFilter += $" and {additionalFilter}";
+        }
+
+        return queryFilter + $" | limit {limit}";
+    }
+
+    private static List<T> MaterializeResults<T>(ResourceQueryResult? result, Func<JsonElement, T> converter)
+    {
+        var results = new List<T>();
+        if (result == null || result.Count == 0)
+        {
+            return results;
+        }
+
+        using var jsonDocument = JsonDocument.Parse(result.Data);
+        var dataArray = jsonDocument.RootElement;
+        if (dataArray.ValueKind == JsonValueKind.Array)
+        {
+            foreach (var item in dataArray.EnumerateArray())
+            {
+                results.Add(converter(item));
+            }
+        }
+
+        return results;
+    }
+
+    /// <summary>
+    /// Resolves the tenant to run a query against when no subscription is available to derive it from.
+    /// </summary>
+    private async Task<TenantResource> ResolveTenantResourceAsync(string? tenant, CancellationToken cancellationToken)
+    {
+        var tenantId = await AzureService.ResolveTenantIdAsync(tenant, cancellationToken);
+        if (!string.IsNullOrEmpty(tenantId))
+        {
+            return await GetTenantResourceAsync(Guid.Parse(tenantId), cancellationToken);
+        }
+
+        var tenants = await AzureService.GetTenants(cancellationToken);
+        if (tenants.Count == 1)
+        {
+            return tenants[0];
+        }
+
+        if (tenants.Count == 0)
+        {
+            throw new InvalidOperationException("No accessible Azure tenants were found for the current credential.");
+        }
+
+        throw new ArgumentException(
+            "Multiple tenants are accessible, so the tenant to query cannot be inferred. Specify the tenant explicitly.",
+            nameof(tenant));
     }
 
     /// <summary>
--- a/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
 using Azure.Core;
 using Azure.Mcp.Core.Services.Azure.Helpers;
 using Azure.ResourceManager;
@@ -111,6 +112,15 @@ public abstract class BaseAzureService(IAzureService azureService)
         ArmClientOptions? armClientOptions = null,
         CancellationToken cancellationToken = default) =>
         await AzureHelper.CreateArmClientAsync(AzureService, tenantIdOrName, armClientOptions: armClientOptions, cancellationToken: cancellationToken);
+
+    /// <summary>
+    /// Validates that the provided parameter is not null or empty.
+    /// </summary>
+    /// <param name="name">The parameter name.</param>
+    /// <param name="value">The parameter value.</param>
+    /// <exception cref="ArgumentException">Thrown when the parameter is null or empty.</exception>
+    public static void ValidateRequiredParameter(string name, [NotNull] string? value) =>
+        AzureHelper.ValidateRequiredParameter(name, value);
 
     /// <summary>
     /// Validates that the provided named parameters are not null or empty
--- a/core/Azure.Mcp.Core/src/Services/Azure/Helpers/AzureHelper.cs
+++ b/core/Azure.Mcp.Core/src/Services/Azure/Helpers/AzureHelper.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Runtime.Versioning;
 using Azure.Core;
@@ -44,6 +45,20 @@ public static class AzureHelper
         // Initialize the default user agent policy without transport type
         s_defaultUserAgent = $"azmcp/{s_version} ({s_framework}; {s_platform})";
         s_sharedUserAgentPolicy = new UserAgentPolicy(s_defaultUserAgent);
+    }
+
+    /// <summary>
+    /// Validates that the provided parameter is not null or empty.
+    /// </summary>
+    /// <param name="name">The parameter name.</param>
+    /// <param name="value">The parameter value.</param>
+    /// <exception cref="ArgumentException">Thrown when the parameter is null or empty.</exception>
+    internal static void ValidateRequiredParameter(string name, [NotNull] string? value)
+    {
+        if (string.IsNullOrEmpty(value))
+        {
+            throw new ArgumentException($"Required parameter is null or empty: {name}");
+        }
     }
 
     /// <summary>
--- a/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
+++ b/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
@@ -1024,8 +1024,8 @@ The `Interaction` column describes whether a prompt can invoke its tool immediat
 |:----------|:------------|:------------|
 | role_assignment_list | List all available role assignments in my subscription | none |
 | role_assignment_list | Show me the available role assignments in my subscription | none |
-| role_assignment_list | List the role assignments on management group <management_group_name> | none |
-| role_assignment_list | List the role assignments on resource group <resource_group_name> | none |
+| role_assignment_list | List the role assignments at scope /providers/Microsoft.Management/managementGroups/<management-group> | none |
+| role_assignment_list | List the role assignments at scope <scope> | none |
 
 ## Azure Redis
 
--- a/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
+++ b/servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
@@ -1025,7 +1025,7 @@ The `Interaction` column describes whether a prompt can invoke its tool immediat
 | role_assignment_list | List all available role assignments in my subscription | none |
 | role_assignment_list | Show me the available role assignments in my subscription | none |
 | role_assignment_list | List the role assignments at scope /providers/Microsoft.Management/managementGroups/<management-group> | none |
-| role_assignment_list | List the role assignments at scope <scope> | none |
+| role_assignment_list | List the role assignments at scope /subscriptions/<subscription>/resourceGroups/<resource-group> | none |
 
 ## Azure Redis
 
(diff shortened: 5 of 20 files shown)


tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -32,26 +35,38 @@ public sealed class RoleAssignmentListCommand(ILogger<RoleAssignmentListCommand>
     private readonly ILogger<RoleAssignmentListCommand> _logger = logger;
     private readonly IAuthorizationService _authorizationService = authorizationService;
 
+    public override void ValidateOptions(RoleAssignmentListOptions options, ValidationResult validationResult)
+    {
+        // A management group scope sits outside every subscription, so the inherited --subscription
+        // requirement would reject a valid request. Skipping the base call skips only that check.
```
vukelich: This comment concerns me. I can’t research well on my phone, but it seems aggressive to skip what all of the base type does to bypass one check. Please understand what the base type does, if derived types are expected to always use the validation from their inherited types, and either (a) fix the comment to be clear that no base type validation is required OR (b) change the code to allow for derived types to have some form of explicit opt-out of specific pieces of inherited type’s validation.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -32,26 +35,38 @@ public sealed class RoleAssignmentListCommand(ILogger<RoleAssignmentListCommand>
...
                 options.Tenant,
                 cancellationToken);
 
-            context.Response.Results = ResponseResult.Create(new(assignments?.Results ?? [], assignments?.AreResultsTruncated ?? false), AuthorizationJsonContext.Default.RoleAssignmentListCommandResult);
+            context.Response.Results = ResponseResult.Create(new(options.Scope, assignments?.Results ?? [], assignments?.AreResultsTruncated ?? false), AuthorizationJsonContext.Default.RoleAssignmentListCommandResult);
         }
         catch (Exception ex)
         {
-            _logger.LogError(ex, "An exception occurred listing role assignments.");
+            _logger.LogError(ex, "An exception occurred listing role assignments for scope {Scope}.", options.Scope);
```
vukelich: Is this is good exception message if the scope is null or empty? Please add quotes to deliniate the value from the rest of the sentence, and please use some clear messaging when the value is null rather than an  empty string. Is there more that would be helpful to include in this, too?

tools/Azure.Mcp.Tools.Authorization/src/Services/AuthorizationService.cs
```diff
@@ -12,22 +11,45 @@ namespace Azure.Mcp.Tools.Authorization.Services;
...
+        }
+
+        ValidateRequiredParameters((nameof(subscription), subscription));
+
         return await ExecuteResourceQueryAsync(
-            "Microsoft.Authorization/roleAssignments",
+            RoleAssignmentResourceType,
             null, // all resource groups
-            subscription,
+            subscription!,
```
vukelich: I believe we can avoid adding the null-forgiveness operator (!) here if we update ValidateRequiredParameter by using NotNull on the second method parameter. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull and double check if appropriate. I can’t see the code on mobile to know if this is not suitable.

tools/Azure.Mcp.Tools.Authorization/src/Services/IAuthorizationService.cs
```diff
@@ -9,15 +9,18 @@ namespace Azure.Mcp.Tools.Authorization.Services;
...
+    /// <param name="scope">The scope that the role assignments apply against.</param>
     /// <param name="tenantId">Optional tenant ID for cross-tenant operations.</param>
     /// <param name="cancellationToken">Optional cancellation token for the operation.</param>
-    /// <returns>List of role assignments in the format "Role Definition ID: Principal ID"</returns>
+    /// <returns>The role assignments found at or below the scope.</returns>
     Task<ResourceQueryResults<RoleAssignment>> ListRoleAssignmentsAsync(
-        string subscription,
+        string? subscription,
         string scope,
         string? tenantId = null,
```
vukelich, line 25: I’m getting a sense of AI cooking up some speghetti code over time. These parameters are being used in ways that sometimes overlap and are sometimes incompatible. 

Scope is the real valuable parameter. It dictates the query is within (a) subscription, (b) a resource group, (c) some resource, (d) a management group, or (e) the current tenant (is just ‘/’ supported by our code and ARG? I assume it should be but could be wrong.)

Subscription is used in the base type to both 

(a) get to an instance if TenantResource. I’d wager we could replace this with using the optional tenant parameter while more likely just pulling the tenantId from the current auth context.

(b) add the REST API argument for ARG’s endpoint to have it narrow down the results to a specific sub. This code change shows this is not a universally appropriate parameter and should be required only by flows actually needing that closer to the tool call rather than deep in BaseAzureResourceService, potentially.

servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
```diff
@@ -980,6 +980,8 @@ The `Interaction` column describes whether a prompt can invoke its tool immediat
 |:----------|:------------|:------------|
 | role_assignment_list | List all available role assignments in my subscription | none |
 | role_assignment_list | Show me the available role assignments in my subscription | none |
+| role_assignment_list | List the role assignments on management group <management_group_name> | none |
+| role_assignment_list | List the role assignments on resource group <resource_group_name> | none |
```
alzimmermsft: The tool doesn't have a `resource-group` parameter

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -88,19 +88,97 @@ protected async Task<ResourceQueryResults<T>> ExecuteResourceQueryAsync<T>(
     {
         ValidateRequiredParameters((nameof(resourceType), resourceType), (nameof(subscription), subscription));
         ArgumentNullException.ThrowIfNull(converter);
+        ValidateAdditionalFilter(additionalFilter);
```
vukelich, line 91: just a disclaimer: I’m not reviewing this file because it’s way too hard to review on mobile. I left a good comment on the interface, though.

vukelich (commented):
Some design comments. Also, please check if we and ARG support tenant-level queries. This is the granularity between "a discrete set of subscriptions" and "as a management group."

tools/Azure.Mcp.Tools.Authorization/src/Services/IAuthorizationService.cs
```diff
@@ -9,15 +9,18 @@ namespace Azure.Mcp.Tools.Authorization.Services;
...
     /// <summary>
-    /// Lists all role assignments in the subscription.
+    /// Lists the role assignments at or below the requested scope.
     /// </summary>
-    /// <param name="subscription">The subscription ID to query.</param>
-    /// <param name="scope">The scope that the resource will apply against.</param>
+    /// <param name="subscription">
+    /// The subscription ID to query. Not required when <paramref name="scope"/> is a management group scope,
+    /// because those assignments live outside any subscription.
+    /// </param>
```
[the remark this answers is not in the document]
KarishmaGhiya: Fixed in 766315e - now reads `The subscription ID or name to query`, matching the repo convention that `subscription` accepts either.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -17,8 +17,11 @@ namespace Azure.Mcp.Tools.Authorization.Commands;
     Name = "list",
     Title = "List Role Assignments",
     Description = """
-        List role assignments. This command retrieves and displays all Azure RBAC role assignments
-        in the specified scope. Results include role definition IDs and principal IDs, returned as a JSON array.
+        List role assignments. This command retrieves and displays the Azure RBAC role assignments
+        at the specified scope and at any scope nested beneath it. Assignments inherited from a parent
+        scope are not included. The scope may be a subscription, resource group, resource, or management
+        group; a subscription is not required when the scope is a management group. Results include role
+        definition IDs and principal IDs, returned as a JSON array.
```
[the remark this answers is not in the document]
KarishmaGhiya: Correct, and I introduced this. The result is `RoleAssignmentListCommandResult(string Scope, List<RoleAssignment> Assignments, bool AreResultsTruncated)` - an object, not an array. Dropped the trailing `, returned as a JSON array` clause in 98f99a8.

servers/Azure.Mcp.Server/docs/e2eTestPrompts.md
KarishmaGhiya: Good catch - fixed in 98f99a8.

`role_assignment_list` only exposes `--scope`, `--subscription` and `--tenant`, so both prompts I added here were implying parameters that don't exist. I've reworded them to name `--scope` explicitly and spell the value out as an ARM scope path, so the placeholders are segments of the scope rather than standalone parameters:

- `List the role assignments at scope /providers/Microsoft.Management/managementGroups/<management-group>`
- `List the role assignments at scope /subscriptions/<subscription>/resourceGroups/<resource-group>`

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -17,8 +17,11 @@ namespace Azure.Mcp.Tools.Authorization.Commands;
     Name = "list",
     Title = "List Role Assignments",
     Description = """
-        List role assignments. This command retrieves and displays all Azure RBAC role assignments
-        in the specified scope. Results include role definition IDs and principal IDs, returned as a JSON array.
+        List role assignments. This command retrieves and displays the Azure RBAC role assignments
```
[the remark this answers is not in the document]
KarishmaGhiya: Score is in the PR description. Against `text-embedding-3-large`:

| Prompt | Original | New |
| --- | --- | --- |
| list role assignments for my management group | #2, 0.5477 | #1, 0.6406 |
| show rbac role assignments at this scope | #1, 0.5920 | #1, 0.6068 |

Both rank #1, above the 0.4 threshold. The only later edit to the description **removed** the inaccurate `returned as a JSON array` clause, so the scored text still stands.

tools/Azure.Mcp.Tools.Authorization/src/Services/AuthorizationService.cs
```diff
@@ -12,22 +11,45 @@ namespace Azure.Mcp.Tools.Authorization.Services;
...
     {
         ValidateRequiredParameters((nameof(scope), scope));
 
-        var scopeId = new ResourceIdentifier(scope!);
+        // Match the scope itself, plus anything nested beneath it. The trailing separator keeps a scope
+        // ending in "rg1" from also matching "rg10".
+        var escapedScope = EscapeKqlString(scope.TrimEnd('/'));
+        var scopeFilter = $"(properties.scope =~ '{escapedScope}' or properties.scope startswith '{escapedScope}/')";
+
+        if (ManagementGroupScope.TryParse(scope, out var managementGroup))
```
JasonYeMSFT, line 30: This method checks if the provided scope is a managementGroupScope and executes the management group scoped resource graph query. Otherwise, it executes the subscription scoped resource graph query.

1. What's the expected behavior when only a scope is provided but it's a resource group scope? One of the example e2e prompt is "/subscriptions/<subscription>/resourceGroups/<resource-group>" and it doesn't have the management group scope prefix used in ManagementGroupScope.cs. It seems the query will list role assignments for all the resource groups instead.

2. What's the expected behavior when both management group scope and subscription are provided? It seems the subscription parameter will be silently dropped. We can either document this behavior or throw an error.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
KarishmaGhiya: Fixed in 9ba0216. `SubscriptionCommand<TOptions, TResult>` now has an explicit `IsSubscriptionApplicable(options)` extension point used by both post-bind subscription resolution and required-option validation. `RoleAssignmentListCommand` overrides that hook for management-group scopes and still calls `base.ValidateOptions`, so it no longer bypasses inherited validation. I also added an XML remark documenting that the hook only reads already-bound values.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
KarishmaGhiya: Fixed in 9ba0216. The structured log now quotes the scope and formats invalid binder values explicitly as `<null>` or `<empty>`, rather than allowing null to render like an empty value.

tools/Azure.Mcp.Tools.Authorization/src/Services/AuthorizationService.cs
KarishmaGhiya: Fixed in 9ba0216. I added the singular `ValidateRequiredParameter(string name, [NotNull] string? value)` helper and use it for `subscription`, so nullable flow analysis proves the value is non-null and the null-forgiveness operator is removed.

tools/Azure.Mcp.Tools.Authorization/src/Services/IAuthorizationService.cs
KarishmaGhiya: Addressed the subscription/base-type concern in 9ba0216 by making subscription applicability explicit in `SubscriptionCommand`. Management-group scopes skip ambient subscription resolution, reject an explicitly supplied subscription, and use `ManagementGroups` in the ARG request; subscription/resource-group/resource scopes continue using the subscription to select the tenant and populate the ARG subscription scope, with `properties.scope` providing the exact requested-scope filter. I also checked tenant-level ARG behavior: ARG supports it by omitting both `subscriptions` and `managementGroups` from the request. The current tool does not treat `/` as a supported scope, and I have not added tenant-wide querying to this focused management-group fix because it introduces separate semantics, permissions, and testing requirements.

tools/Azure.Mcp.Tools.Authorization/src/Services/AuthorizationService.cs
KarishmaGhiya: Addressed in 9ba0216. For a resource-group ARM scope, the ARG request is limited to the supplied subscription and the KQL `properties.scope` predicate limits results to that exact resource-group scope and descendants, so it does not return assignments from all resource groups. For a management-group scope, an explicitly supplied subscription is now rejected by command validation and service validation rather than silently ignored; ambient/default subscription resolution is skipped for that scope.

KarishmaGhiya: 1. Resource-group scope without a separate subscription argument: it is rejected during command validation. Although the subscription appears inside the ARM scope, the tool still requires  --subscription  for subscription, resource-group, and resource scopes. When supplied, Resource Graph is limited to that subscription, while this predicate restricts assignments to the requested resource group and descendants:

properties.scope =~ '<resource-group-scope>'
or properties.scope startswith '<resource-group-scope>/'

Therefore, it does not list assignments from every resource group. The E2E prompt should include the tool’s separate subscription input to be directly invocable:

| role_assignment_list | List the role assignments at scope /subscriptions/<subscription>/resourceGroups/<resource-group> in subscription <subscription> | none |

2. Management-group scope with a subscription: it is no longer silently ignored. Commit  9ba0216  rejects the combination in both command validation and  AuthorizationService , with guidance to omit  --subscription . Ambient subscription resolution is also skipped for management-group scopes because management groups exist outside subscriptions.

JasonYeMSFT (approved):

core/Azure.Mcp.Core/src/Services/Azure/Helpers/AzureHelper.cs
```diff
@@ -46,6 +47,20 @@ static AzureHelper()
...
+    /// <param name="value">The parameter value.</param>
+    /// <exception cref="ArgumentException">Thrown when the parameter is null or empty.</exception>
+    internal static void ValidateRequiredParameter(string name, [NotNull] string? value)
+    {
+        if (string.IsNullOrEmpty(value))
+        {
+            throw new ArgumentException($"Required parameter is null or empty: {name}");
+        }
+    }
+
```
alzimmermsft, line 63: Option binding already validates that a required parameter is non-null, non-whitespace, do we actually need this?

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
```diff
@@ -112,6 +113,15 @@ protected async Task<ArmClient> CreateArmClientAsync(
...
 
+    /// <summary>
+    /// Validates that the provided parameter is not null or empty.
+    /// </summary>
+    /// <param name="name">The parameter name.</param>
+    /// <param name="value">The parameter value.</param>
+    /// <exception cref="ArgumentException">Thrown when the parameter is null or empty.</exception>
+    public static void ValidateRequiredParameter(string name, [NotNull] string? value) =>
+        AzureHelper.ValidateRequiredParameter(name, value);
+
```
alzimmermsft, line 124: If the response to the other question about this is that it isn't needed this can be removed as well.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -33,26 +36,50 @@ public sealed class RoleAssignmentListCommand(ILogger<RoleAssignmentListCommand>
...
-            context.Response.Results = ResponseResult.Create(new(assignments?.Results ?? [], assignments?.AreResultsTruncated ?? false), AuthorizationJsonContext.Default.RoleAssignmentListCommandResult);
+            context.Response.Results = ResponseResult.Create(new(options.Scope, assignments?.Results ?? [], assignments?.AreResultsTruncated ?? false), AuthorizationJsonContext.Default.RoleAssignmentListCommandResult);
         }
         catch (Exception ex)
         {
-            _logger.LogError(ex, "An exception occurred listing role assignments.");
+            _logger.LogError(
+                ex,
+                "An exception occurred listing role assignments for scope '{Scope}'.",
+                FormatScopeForLogging(options.Scope));
```
alzimmermsft: Shouldn't this error logging if / else based on what type of listing we performed? If we're listing on a subscription and it fails, we now log an unhelpful message.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -33,26 +36,50 @@ public sealed class RoleAssignmentListCommand(ILogger<RoleAssignmentListCommand>
...
         return context.Response;
     }
 
-    public sealed record RoleAssignmentListCommandResult(List<RoleAssignment> Assignments, bool AreResultsTruncated);
+    private static string FormatScopeForLogging(string? scope) => scope switch
+    {
+        null => "<null>",
+        "" => "<empty>",
+        _ => scope
+    };
```
alzimmermsft: Pending the answer on logging, this may be removed.

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
```diff
@@ -14,18 +14,31 @@ public abstract class SubscriptionCommand<
 {
     private readonly ISubscriptionResolver _subscriptionResolver = subscriptionResolver;
 
+    /// <summary>
+    /// Determines whether subscription resolution and validation apply to the bound options.
+    /// </summary>
+    /// <remarks>This method is evaluated during post-binding and must only inspect already-bound option values.</remarks>
+    protected virtual bool IsSubscriptionApplicable(TOptions options) => true;
```
alzimmermsft: Instead of adding this, do what other commands are doing when `subscription` is optional, have them extend from `AuthenticatedCommand` directly and manually bind `subscription` with an `ISubscriptionResolver`. Here's an example:

https://github.com/microsoft/mcp/blob/main/tools/Azure.Mcp.Tools.EventGrid/src/Commands/Subscription/SubscriptionListCommand.cs#L27

Then we can remove this API as it's a bit confusing. And remove the other changes to this class.

core/Azure.Mcp.Core/src/Services/Azure/Helpers/AzureHelper.cs
KarishmaGhiya: Left this one as-is. The option binder only validates at the CLI boundary — `AuthorizationService`  is a public method that gets called directly in tests (and potentially by other future callers), so I'd rather keep an explicit guard at the service layer than assume every caller goes through command validation first. The `[NotNull]`  overload is also what lets subscription flow through without a ` !`  at the  `ExecuteResourceQueryAsync`  call site, which was your suggestion earlier in review. If you feel strongly it should go, I'm fine either way

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureService.cs
KarishmaGhiya: Tied to the answer above — since I'm keeping the single-param overload, this wrapper stays too (it's the same pattern as the existing plural  `ValidateRequiredParameters`  wrapper right next to it). If we end up dropping #1, this comes out with it.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
KarishmaGhiya: Good catch, you're right that it was too generic. Fixed — the catch block now branches on scope type: management-group failures just log the scope (no subscription to report), everything else logs scope + subscription so we're not losing context on a subscription-scoped failure anymore.

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
KarishmaGhiya: Kept it, just renamed and reused —  `FormatScopeForLogging`  →  `FormatForLogging` , now handles both the scope and subscription values for the two log messages above.

core/Azure.Mcp.Core/src/Commands/Subscription/SubscriptionCommand`2.cs
KarishmaGhiya: Agreed, that's cleaner. Reverted  `SubscriptionCommand<TOptions,TResult>`  back to how it was — no more  `IsSubscriptionApplicable`  hook.  `RoleAssignmentListCommand`  now extends  `AuthenticatedCommand`  directly and handles its own subscription binding/validation, same as  `EventGrid.SubscriptionListCommand` . Double-checked nothing else was relying on the hook. All the Authorization tests plus a full solution build are still green

core/Azure.Mcp.Core/src/Services/Azure/Helpers/AzureHelper.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
```
g2vinay: .

tools/Azure.Mcp.Tools.Authorization/tests/Azure.Mcp.Tools.Authorization.Tests/AuthorizationServiceTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using Xunit;
+
+namespace Azure.Mcp.Tools.Authorization.Tests;
+
+public sealed class AuthorizationServiceTests
+{
+    [Fact]
+    public async Task ListRoleAssignmentsAsync_RejectsSubscription_ForManagementGroupScope()
+    {
+        var service = new AuthorizationService(Substitute.For<IAzureService>());
```
g2vinay, line 16: Could we add a transport-level service test that exercises the successful management-group path and asserts the ARG request uses `managementGroups`, not `subscriptions`, then verifies response mapping? The command test mocks `IAuthorizationService`, and this test exits on validation, so the core fix currently has no regression coverage. A real management-group live test can remain an acknowledged limitation.

g2vinay (approved):

tools/Azure.Mcp.Tools.Authorization/tests/Azure.Mcp.Tools.Authorization.Tests/AuthorizationServiceTests.cs
KarishmaGhiya: Good call — you're right that the current coverage has a gap there. I added a transport-level test (`AuthorizationServiceManagementGroupTransportTests`, commit `378486b`) that wires a real `ArmClient` to a fake `HttpMessageHandler` instead of mocking `IAuthorizationService`. It:

- Lists tenants and issues the Resource Graph query through the real `AuthorizationService.ListRoleAssignmentsAsync` path
- Asserts the outgoing request body scopes to `managementGroups` and never populates `subscriptions`
- Asserts the response JSON is correctly mapped back into a `RoleAssignment` (id, scope, principalId, principalType, roleDefinitionId)

So the core fix now has regression coverage if someone accidentally reverts it to a subscription-scoped query. Agreed that a real live management-group test remains a reasonable acknowledged limitation for now.

KarishmaGhiya:
@alzimmermsft please re-review

tools/Azure.Mcp.Tools.Authorization/src/Commands/RoleAssignmentListCommand.cs
```diff
@@ -28,31 +30,95 @@ List role assignments. This command retrieves and displays all Azure RBAC role a
...
+        {
+            return;
+        }
+
+        // Always post-process subscription via resolver (env var / CLI profile fallback)
+        options.Subscription = _subscriptionResolver.ResolveSubscription(options.Subscription);
+        if (!string.IsNullOrEmpty(options.Subscription))
+        {
+            // Trim any surrounding quotes that may have been included in the input
+            options.Subscription = options.Subscription.Trim('"', '\'');
```
alzimmermsft: nit: `SubscriptionResolver.ResolveSubscription` already removes these surrounding quotes.

alzimmermsft (approved):

g2vinay (approved):

PR #3630, opened 2026-09-11
Return concise errors for invalid command names

### LINKED ISSUE #2624: [BUG] Incorrect command name yields large amounts of help text
### Describe the bug

If the LM provides the wrong command name to a tool the response can be a large amount of help information, including every command, the command's description, and the schema of the parameters. With this much information to dig through the LM often doesn't find the correct command.

### Expected behavior

The text returned by Azure MCP should explain that the given command isn't valid for the tool, and list just the available command names. Possibly whittled down to the closest matches.

The LM is then able to identify the correct command name and successfully call the tool.

### Actual behavior

For the `azure-appservice` tool the help content is 21.7 KB. Copilot CLI ends up writing the content to a file because it is so long, and provides the LM the file path and a "preview" of the first 500 characters. The LM may not read the entire file (or perhaps any of it) and if the correct command name isn't in the preview then it may never identify it. 

### Reproduction Steps

While using Azure MCP in the default namespace mode ask Copilot CLI to "Get app service details for <app service ID>".

### Environment

_No response_

RickWinter:
## What does this PR do?

Return an error with available command names instead of full help when namespace or consolidated routing cannot resolve a command, including external-server routing. Preserve one supported sampling correction attempt and leave explicit learning unchanged.

## GitHub issue number?

Fixes #2624.

## Invoking Livetests

### DIFF
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
@@ -92,6 +92,30 @@ public abstract class BaseToolLoader(ILogger logger) : IToolLoader
         return parametersElem.EnumerateObject().ToDictionary(prop => prop.Name, prop => (object?)prop.Value);
     }
 
+    internal static CallToolResult CreateUnknownCommandResult(string toolName, string commandName, IEnumerable<string> availableCommandNames)
+    {
+        var names = availableCommandNames.Distinct(StringComparer.Ordinal).Order(StringComparer.Ordinal).ToArray();
+        var availableCommands = names.Length == 0
+            ? "No commands are available for this tool in the current server configuration."
+            : $"Available commands: {string.Join(", ", names)}";
+
+        return new CallToolResult
+        {
+            Content =
+            [
+                new TextContentBlock
+                {
+                    Text = $"""
+                        The command '{commandName}' is not available for the '{toolName}' tool.
+                        {availableCommands}
+                        Use "learn=true" to get command descriptions and parameter schemas.
+                        """
+                }
+            ],
+            IsError = true
+        };
+    }
+
     /// <summary>
     /// The name of the option used to pass raw MCP tool input directly to a command.
     /// </summary>
--- a/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/ServerToolLoader.cs
+++ b/core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/ServerToolLoader.cs
@@ -244,40 +244,31 @@ public sealed class ServerToolLoader(
         {
             Activity.Current?.SetTag(TagName.ToolSource, "external." + client.ServerInfo.Name);
             var availableTools = await GetChildToolListAsync(request, tool, cancellationToken);
+            var resolvedTool = availableTools.FirstOrDefault(t => string.Equals(t.Name, command, StringComparison.OrdinalIgnoreCase));
 
-            // When the specified command is not available, we try to learn about the tool's capabilities
-            // and infer the command and parameters from the users intent.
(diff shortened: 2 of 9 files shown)


core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
```diff
@@ -93,6 +92,30 @@ protected static JsonElement GetParametersJsonElement(RequestContext<CallToolReq
         return parametersElem.EnumerateObject().ToDictionary(prop => prop.Name, prop => (object?)prop.Value);
     }
 
+    internal static CallToolResult CreateUnknownCommandResult(string toolName, string commandName, IEnumerable<string> availableCommandNames)
+    {
+        var names = availableCommandNames.Distinct(StringComparer.Ordinal).Order(StringComparer.Ordinal).ToArray();
+        var availableCommands = names.Length == 0
+            ? "No commands are available for this tool in the current server configuration."
```
alzimmermsft, line 99: Are we able to reach this spot? I believe all the hierarchical tool loaders should fail and return root learning when the tool doesn't exist.

But if we do hit this point somehow we probably should return a different error message that what is happening now. Something like `The tool '{toolName}' wasn't found. Use "learn=true" to get the tools available.`

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
```diff
@@ -93,6 +92,30 @@ protected static JsonElement GetParametersJsonElement(RequestContext<CallToolReq
...
+        return new CallToolResult
+        {
+            Content =
+            [
+                new TextContentBlock
+                {
+                    Text = $"""
+                        The command '{commandName}' is not available for the '{toolName}' tool.
+                        {availableCommands}
+                        Use "learn=true" to get command descriptions and parameter schemas.
```
alzimmermsft, line 111: I'd expand this slightly to include something like `Select the command which best fits the current intent and use "learn=true" to get the description and parameters for the command.`

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/ServerToolLoader.cs
```diff
@@ -561,10 +552,15 @@ await request.Server.NotifyProgressAsync(progressToken.Value,
                     parameters = parametersElem.EnumerateObject().ToDictionary(prop => prop.Name, prop => (object?)prop.Value.Clone()) ?? [];
                 }
             }
-            if (commandName != null && commandName != "Unknown")
+            var resolvedTool = !string.IsNullOrWhiteSpace(commandName) && commandName != "Unknown"
+                ? availableTools.FirstOrDefault(t => string.Equals(t.Name, commandName, StringComparison.OrdinalIgnoreCase))
+                : null;
```
alzimmermsft, line 557: Should we pull this logic up into `BaseToolLoader` so both `NamespaceToolLoader` and `ServerToolLoader` don't need to implement this same logic?

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
RickWinter: This is reachable when the tool exists but read-only or HTTP filtering removes all of its commands. The tool was found, so saying it wasn’t found would be misleading. I’ve kept the empty-catalog message and the regression coverage for that case.

core/Microsoft.Mcp.Core/src/Areas/Server/Commands/ToolLoading/BaseToolLoader.cs
RickWinter: Stronger recovery guidance makes sense, but `learn=true` currently returns the catalog rather than one command’s schema, and a nonblank intent can trigger execution through sampling. We don't want that to happen.

alzimmermsft (approved):

PR #3675, opened 2026-09-15
fix(http): support hostname, domain suffix, and CIDR in NO_PROXY bypass list (#3338)

### LINKED ISSUE #3338: [BUG] NO_PROXY host, suffix, and CIDR entries do not bypass the configured proxy
### Describe the bug

Azure MCP honors `HTTP_PROXY` / `HTTPS_PROXY`, but its custom `NO_PROXY` handling is incompatible with standard hostname, leading-dot suffix, and CIDR entries.

`AddHttpClientServices` reads `ALL_PROXY`, `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`. `HttpClientFactoryConfigurator.CreateProxy` then creates a `WebProxy`, converts each `NO_PROXY` entry with `ConvertGlobToRegex`, and assigns the result to `WebProxy.BypassList`:

- https://github.com/microsoft/mcp/blob/main/core/Microsoft.Mcp.Core/src/Extensions/HttpClientServiceCollectionExtensions.cs
- https://github.com/microsoft/mcp/blob/main/core/Microsoft.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs

The conversion anchors the supplied entry itself. For example:

```text
api.loganalytics.io -> ^api\.loganalytics\.io$
.ods.opinsights.azure.com -> ^\.ods\.opinsights\.azure\.com$
10.0.0.0/8 -> ^10\.0\.0\.0/8$
```

However, .NET `WebProxy.BypassList` contains regular expressions evaluated against a formatted URL such as `https://api.loganalytics.io` or `https://workspace.ods.opinsights.azure.com`, not only `Uri.Host`. Therefore the exact-host and standard leading-dot suffix expressions above do not match. CIDRs are treated as literal strings and also do not match addresses in their range.

A raw glob such as `*.ods.opinsights.azure.com` may appear to work accidentally because its leading `.*` can consume the URL scheme and hostname prefix, but standard `NO_PROXY` implementations use `.ods.opinsights.azure.com` for suffix matching. Callers also cannot work around the mismatch by supplying a `WebProxy` URL regex because `ConvertGlobToRegex` escapes regex metacharacters before assignment.

This custom parser also bypasses the platform `HttpEnvironmentProxy` implementation. .NET 11 added native IPv4/IPv6 CIDR support for `NO_PROXY` in https://github.com/dotnet/runtime/pull/131397, but current Azure MCP still performs its own conversion. Azure MCP reads only uppercase variables as well, while .NET checks lowercase names first on case-sensitive platforms.

The implementation is present in the `Azure.Mcp.Server-2.0.1` tag and remains on current `main`.

### Expected behavior

Standard `NO_PROXY` values should bypass the proxy:

- `api.loganalytics.io` for that exact hostname
- `.ods.opinsights.azure.com` for the domain and its subdomains
- `10.0.0.0/8` for IP destinations within that network when supported by the target runtime
- lowercase and uppercase environment names according to platform/.NET precedence

Preferably Azure MCP should delegate environment-proxy parsing to .NET instead of maintaining a divergent parser. If a custom proxy remains necessary, it should implement the same hostname, suffix, CIDR, casing, and precedence semantics explicitly.

### Actual behavior

With `HTTPS_PROXY` configured, requests covered by exact-host, standard leading-dot suffix, or CIDR `NO_PROXY` entries still go through the proxy.

Observed examples include:

```text
api.loganalytics.io
api.loganalytics.azure.com
api.applicationinsights.io
.ods.opinsights.azure.com
10.0.0.0/8
```

### Reproduction Steps

1. Start Azure MCP on Linux with a proxy that logs requests:

   ```bash
   HTTP_PROXY=http://proxy.example:3128 \
   HTTPS_PROXY=http://proxy.example:3128 \
   NO_PROXY=api.loganalytics.io,.ods.opinsights.azure.com,10.0.0.0/8 \
   npx -y @azure/mcp@2.0.1 server start --namespace monitor --mode all --read-only
   ```

2. Invoke Monitor tools that call Log Analytics or Application Insights, or otherwise exercise a destination covered by the configured entries.
3. Observe those requests reaching the proxy despite the matching `NO_PROXY` entries.
4. The behavior follows directly from testing the generated regexes against the URL strings used by `WebProxy.BypassList`.

### Environment

- Hosting platform: Linux container / Azure sandbox
- Azure MCP: `@azure/mcp@2.0.1`
- Azure MCP target framework at that tag: .NET 10
- Source review confirms current `main` retains the same custom parser
- Relevant runtime CIDR change: dotnet/runtime#131397, targeting .NET 11

teddiesloco:
Fixes #3338

## Problem
In `HttpClientFactoryConfigurator.CreateProxy`, `NO_PROXY` entries were converted using `ConvertGlobToRegex` by escaping and anchoring the pattern directly (e.g. `api.loganalytics.io` -> `^api\.loganalytics\.io$`).

However, .NET `WebProxy.BypassList` evaluates regular expressions against full URIs (such as `https://api.loganalytics.io/v1/workspaces`) rather than just the host. Consequently:
- Exact hostnames (e.g. `api.loganalytics.io`) never matched because the URI starts with the scheme (`https://`).
- Leading-dot domain suffixes (e.g. `.ods.opinsights.azure.com`) never matched.
- IPv4 CIDRs (e.g. `10.0.0.0/8`, `172.16.0.0/12`) were treated as literal strings and failed to match destinations.
- Lowercase proxy environment variables (`all_proxy`, `http_proxy`, `https_proxy`, `no_proxy`) were ignored on Unix/Linux platforms.

## Solution
1. **URI-aware regex conversion**: `ConvertGlobToRegex` now generates regexes matching URI schemes, optional ports, and path suffixes:
   - Exact hosts: `^[^:]+://host(:\d+)?(/.*)?$`
   - Domain suffixes (`.example.com`, `*.example.com`): `^[^:]+://([^/:]+\.)?example\.com(:\d+)?(/.*)?$`
   - Host with port (`localhost:5000`): `^[^:]+://localhost:5000(/.*)?$`
   - Wildcards: `*` and `*.*` map to `.*`
2. **IPv4 CIDR support**: Converts standard CIDRs (such as `/8`, `/12`, `/16`, `/24`, `/32`) to matching IP range regex patterns.
3. **Environment variable casing fallback**: Added fallback to lowercase `all_proxy`, `http_proxy`, `https_proxy`, `no_proxy` in `HttpClientServiceCollectionExtensions.cs`.
4. **Unit Tests**: Added comprehensive unit test coverage in `HttpClientFactoryConfiguratorTests.cs` covering exact hosts, subdomains, CIDRs, ports, and wildcards.


KarishmaGhiya:
Found a couple of edge-case gaps worth flagging (not blockers, but should be called out — see inline comments):
1. CIDR prefixes outside {8,16,24,32}∪[8,30] (e.g. `/7`, `/31`) silently never bypass.
2. IPv6 CIDRs and bare IPv6 host literals (`::1`, `fe80::1`) silently never bypass, since real request URIs bracket IPv6 hosts (`https://[::1]/`) but the generated regex doesn't include brackets, and the CIDR branch only accepts `AddressFamily.InterNetwork`.

Neither of these is a regression — they were already unsupported before this PR — but since the issue explicitly calls out .NET 11's native IPv4 *and* IPv6 CIDR support as the preferred direction, it'd be good to either add IPv6 support here or explicitly document it as an acknowledged follow-up/limitation in the PR description so reviewers don't assume full parity.

core/Microsoft.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -149,7 +149,119 @@ private static HttpMessageHandler CreateHttpMessageHandler(IServiceProvider serv
...
+                }
+                if (mask == 24)
+                {
+                    return $@"^[^:]+://{bytes[0]}\.{bytes[1]}\.{bytes[2]}\.\d{{1,3}}(:\d+)?(/.*)?$";
+                }
+                if (mask == 32)
+                {
+                    return $@"^[^:]+://{bytes[0]}\.{bytes[1]}\.{bytes[2]}\.{bytes[3]}(:\d+)?(/.*)?$";
+                }
+                if (mask is >= 8 and <= 30)
```
KarishmaGhiya: Prefixes outside `{8,16,24,32} ∪ [8,30]` — e.g. `/7` or `/31` — fall through this whole CIDR block and get treated as a literal string containing `/`, so they silently never bypass the proxy (confirmed: `10.0.0.0/7` never matches `10.5.5.5`). Worth either widening the range to cover `/31`-`/32`-style host CIDRs or explicitly rejecting/logging unsupported masks instead of silently no-op'ing.

core/Microsoft.Mcp.Core/src/Services/Http/HttpClientFactoryConfigurator.cs
```diff
@@ -149,7 +149,119 @@ private static HttpMessageHandler CreateHttpMessageHandler(IServiceProvider serv
...
+        if (pattern == "*" || pattern == "*.*")
+        {
+            return ".*";
+        }
+
+        // IPv4 CIDR notation (e.g. 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 127.0.0.1/32)
+        if (pattern.Contains('/'))
+        {
+            var parts = pattern.Split('/');
+            if (parts.Length == 2 && int.TryParse(parts[1], out var mask) && IPAddress.TryParse(parts[0], out var ip) && ip.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
```
KarishmaGhiya: This only handles IPv4 CIDRs — IPv6 CIDRs (e.g. `2001:db8::/32`) fail the `AddressFamily.InterNetwork` check and fall through to literal-string handling, so they never bypass. Separately, bare IPv6 host literals like `::1` or `fe80::1` also never match in practice, since the generated regex omits the `[...]` brackets that real request URIs use for IPv6 hosts (confirmed `https://[::1]:8080/` doesn't match). Since the linked issue calls out .NET 11's native IPv4/IPv6 CIDR support as the preferred direction, consider at least documenting this as a known limitation if IPv6 support isn't in scope for this PR.

KarishmaGhiya (changes requested):

KarishmaGhiya:
@teddiesloco There are some pipeline failures

teddiesloco:
@KarishmaGhiya Fixed the pipeline issues in the latest commit (wrapped CIDR regex ranges with non-capturing groups to preserve precedence, and updated test domain to satisfy spellcheck). Thanks for catching that!

KarishmaGhiya (approved):

PR #3694, opened 2026-09-16
Add additional EndpointValidator usages

alzimmermsft:
## What does this PR do?

Added namespace-aware endpoint validation for Event Grid, IoT Hub, Key Vault, Managed HSM, Managed Lustre, Monitor web test, and MySQL endpoints.

## GitHub issue number?

## Pre-merge Checklist
- [x] Required for All PRs
- [ ] For MCP tool changes:


core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.AllowLists.cs
```diff
@@ -124,12 +124,48 @@ public string[] GetSuffixes(ArmEnvironment environment) =>
             UsGov: [".communication.azure.us"],
             Germany: [".communication.azure.de"],
             UseLegacyCheck: true), // INITIAL SEEDED UseLegacyCheck. NEEDS VERIFICATION.
+        ["eventgrid"] = new AllowedSuffixManager(
+            Public: [".eventgrid.azure.net"],
+            China: [".eventgrid.azure.cn"],
+            UsGov: [".eventgrid.azure.us"],
+            Germany: [".eventgrid.azure.de"],
+            UseLegacyCheck: false),
+        ["eventhubs"] = new AllowedSuffixManager(
```
g2vinay: Nothing in this PR uses `eventhubs`, and its values are the same as `servicebus`. Event Hubs also isn't mentioned in the PR description. Can we remove this for now and add it back when a service actually needs it?

tools/Azure.Mcp.Tools.ManagedLustre/src/Services/ManagedLustreService.cs
```diff
@@ -453,6 +458,18 @@ public async Task<LustreFileSystem> CreateFileSystemAsync(
         return Map(fileSystemResource);
     }
 
+    internal static Uri CreateValidatedKeyUri(string keyUrl, ArmEnvironment armEnvironment)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(keyUrl);
+        var keyUri = new Uri(keyUrl, UriKind.Absolute);
+        EndpointValidator.ValidateAzureServiceEndpoint(
+            endpoint: keyUri.AbsoluteUri,
+            serviceType: "keyvault",
```
g2vinay, line 467: This only allows Key Vault key URLs. If Managed Lustre also lets you use keys from Managed HSM (`*.managedhsm.azure.net`), those would get rejected here. Does it support HSM keys? If yes, we should allow both.

core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.AllowLists.cs
```diff
@@ -124,12 +124,48 @@ public string[] GetSuffixes(ArmEnvironment environment) =>
...
+            China: [".servicebus.chinacloudapi.cn"],
+            UsGov: [".servicebus.usgovcloudapi.net"],
+            Germany: [".servicebus.cloudapi.de"],
+            UseLegacyCheck: false),
         ["foundry"] = new AllowedSuffixManager(
             Public: [".services.ai.azure.com"],
             China: [".services.ai.azure.cn"],
             UsGov: [".services.ai.azure.us"],
             Germany: [".services.ai.azure.de"],
             UseLegacyCheck: true), // INITIAL SEEDED UseLegacyCheck. NEEDS VERIFICATION.
```
vukelich: Foundry is in your list to flip this over to `false` if we're good to go on it. See the instructions higher up on what `UseLegacyCheck` means.

core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.AllowLists.cs
vukelich: > Can we remove this for now and add it back when a service actually needs it?

Agreed.

> its values are the same as servicebus.

I don't outright agree or disagree with this statement. 

For example, Azure Storage's blob and table URIs are sibling child subdomains. I'd argue they should have separate entries in the allow list. We should not include them into a single `"storage"` key nor should their separate lists allow matches to the other Storage feature. We do not want a blob request going to a table URL nor the reverse.

I don't know enough about Event Hub to know if this is an appropriate analogy or not. This is the responsibility of the author to determine and present evidence of preferably in-code for later reference and intention.

tools/Azure.Mcp.Tools.IoTHub/src/Services/IoTHubDeviceService.cs
```diff
@@ -81,6 +82,23 @@ private static HttpPipeline BuildDataPlanePipeline(HttpClient httpClient)
...
 
+    internal static Uri CreateValidatedDataPlaneUri(
+        string hostname,
+        string pathAndQuery,
+        ArmEnvironment armEnvironment)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(hostname);
+        ArgumentException.ThrowIfNullOrWhiteSpace(pathAndQuery);
+
+        var endpoint = new Uri($"https://{hostname}{pathAndQuery}", UriKind.Absolute);
```
vukelich, line 93: A fear I have with this code path is that the `endpoint` value is not what ends up getting used elsewhere. This method ensures _this_ URL is fine, but it's not testing what is actually used.

Could you please tell me more about this code? I'm reviewing quickly on GitHub's website, so navigating the files is difficult.

tools/Azure.Mcp.Tools.ManagedLustre/src/Services/ManagedLustreService.cs
alzimmermsft: They do not support Managed HSM keys

https://learn.microsoft.com/en-us/azure/security/fundamentals/encryption-customer-managed-keys-support#storage

vukelich (commented):
Some initial comments.

tools/Azure.Mcp.Tools.IoTHub/src/Services/IoTHubDeviceService.cs
alzimmermsft: Everywhere calling this does use the `return endpoint` URI. Not sure I understand how this isn't testing what is actually used.

But also, isn't that a general concern always as we can have these validators that get called but the return is never used?

vukelich: Chatting offline over DMs about leveraging `HttpPipelinePolicy`.

vukelich: I'm sorry. I misread the code, and I did not take your correction to heart. You're correct, and my comment is wrong.

vukelich: Adding more context from chat:

> I believe both contexts [in-product URL "testing" and unit/integration "testing"] are important:
> 1. We must validate a URI before we use it. CreateValidatedDataPlaneUri standalone does that.
> 2. We must test our product through automation. This is done via testing the internal static CreateValidatedDataPlaneUri
>
> What's missing is that we're not testing that CreateValidatedDataPlaneUri is used by the other methods in the type. It could be dead product code at any point in time, and the automated tests wouldn't catch that regression.
>
> My comment was stating that by having this hole poked through to the unit tests to get CreateValidatedDataPlaneUri, we leave ourselves open to regressions in the future.
> 1. First can be broken by simply no longer calling CreateValidatedDataPlaneUri or using its results by other methods in the class. We would no longer have our SSRF protections.
> 2. Second will fail to catch anything because it'll see that CreateValidatedDataPlaneUri is still working correctly but won't see that the rest of the type isn't using that method anymore.

tools/Azure.Mcp.Tools.KeyVault/src/Services/KeyVaultService.cs
```diff
@@ -313,13 +321,37 @@ public async Task<GetSettingsResult> GetVaultSettings(
...
+        ArgumentNullException.ThrowIfNull(endpoint);
+        EndpointValidator.ValidateAzureServiceEndpoint(
+            endpoint: endpoint.AbsoluteUri,
+            serviceType: "managedhsm",
+            armEnvironment: armEnvironment,
+            executingToolNamespaceName: "keyvault");
+        return endpoint;
+    }
+
     private KeyVaultSettingsClient CreateSettingsClient(Uri hsmUri, Azure.Core.TokenCredential credential)
```
vukelich: I like that `CreateCertificateClient` does the validation and `HttpClient` creation in the same method. Can we do that here, too?

tools/Azure.Mcp.Tools.KeyVault/src/Services/KeyVaultService.cs
```diff
@@ -297,7 +303,9 @@ private SecretClient CreateSecretClient(string vaultName, Azure.Core.TokenCreden
 
     private CertificateClient CreateCertificateClient(string vaultName, Azure.Core.TokenCredential credential)
     {
-        var vaultUri = new Uri(BuildVaultUri(vaultName));
+        var vaultUri = ValidateVaultEndpoint(
+            new Uri(BuildVaultUri(vaultName)),
+            AzureService.CloudConfiguration.ArmEnvironment);
         var httpClient = AzureService.GetClient();
         httpClient.BaseAddress = vaultUri;
```
vukelich: Warning: `HttpClient.BaseAddress` is ignored if the `HttpClient` method used for sending some request has a different `Uri`. Just because we set `BaseAddress`, it might not mean that's the value actually used. What happens if a caller gives the full URI to a key in a different key vault than the one by the vault name argument given? What does the Key Vault SDK do in that case?

I'd recommend adding test automation to prove what happens for cross-vault URIs. Ideally, the Key Vault SDK throws upon seeing a mismatch without any outbound call, or if our usage of the Key Vault SDK means that we won't end up in the code paths for the KV SDK to handle URIs of keys/certs/secrets.

tools/Azure.Mcp.Tools.ManagedLustre/src/Services/ManagedLustreService.cs
vukelich: This is a great detail to add in the code for other maintainers. Share the knowledge around. Put a record in the code. Add a link to that source in the code. As we're going through this work, we should be making it easier for someone else to come by after for another round of changes.

tools/Azure.Mcp.Tools.Monitor/src/Services/MonitorWebTestService.cs
```diff
@@ -430,6 +440,37 @@ public async Task<WebTestDetailedInfo> UpdateWebTest(
         };
     }
 
+    internal static Uri CreateValidatedRequestUri(string requestUrl, ILogger? logger)
+    {
+        EndpointValidator.ValidatePublicTargetUrl(
+            url: requestUrl,
+            logger: logger,
+            executingToolNamespaceName: "monitor");
+        return new Uri(requestUrl, UriKind.Absolute);
```
vukelich: Minor suggestion: construct the `Uri` object first, then pass a `string` around from it.

```csharp
Uri uri = new(requestUrl, UriKind.Absolute);
EndpointValidator.ValidatePublicTargetUrl(
    url: uri.AbsoluteUri,
    logger: logger,
    executingToolNamespaceName: "monitor");
return uri;
```

tools/Azure.Mcp.Tools.MySql/src/Services/MySqlService.cs
```diff
@@ -47,18 +49,11 @@ private string GetOpenSourceRDBMSScope()
...
-        ".mysql.database.usgovcloudapi.net",
-        ".mysql.database.chinacloudapi.cn",
-    ];
-
     private string NormalizeServerName(string server)
     {
-        if (!server.Contains('.'))
-        {
-            return AzureService.CloudConfiguration.CloudType switch
+        ArgumentException.ThrowIfNullOrWhiteSpace(server);
```
vukelich: My changes to `PostgresService` were initially similar, unsurprisingly. I think our agentic coding tools came up with the same steps needed to confirm these DB endpoints. However, I went through extensive iterating to make sure the code and comments are tight and well detailed for others. I didn't understand why some of these steps were added until I told the model to explain itself.

See: https://github.com/microsoft/mcp/pull/3740/changes/BASE..be6f5a9d18efb03b8626a790b244d54457fe042a#diff-97524150fa7b4e6fbb758739a23ed16b15197f0ef5b6196243825bafeb48affeR62

(If above link is dead from a force-push, see #3740 and `PostgresService.cs` and its `CreateAndValidateServerHostname` method.)

PR #3718, opened 2026-09-18
fix(authorization): avoid tenant enumeration for Resource Graph queries

### LINKED ISSUE #2141: [BUG] `role` tool fails with "No accessible tenant found" while other tools (appservice) work fine
### Describe the bug

The `role` tool (e.g. `role_assignment_list`) returns `InvalidOperationException: No accessible tenant found for tenant ID '<id>'` even when:

- `az login` is authenticated to the correct tenant
- Other MCP tools (e.g. `appservice`) work against the same subscription/tenant
- The `tenant` parameter is passed explicitly to the tool
- The equivalent CLI command `az role assignment list --scope <resource-id>` works correctly

### Expected behavior

The `role_assignment_list` tool should return role assignments for the specified scope, the same way `az role assignment list --scope <resource-id>` does via the CLI.

### Actual behavior

The tool returns HTTP 500 with:

```
InvalidOperationException: No accessible tenant found for tenant ID <GUID>
```

**Root cause in code:**

In [`BaseAzureResourceService.cs`](https://github.com/microsoft/mcp/blob/main/core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs#L42), the `GetTenantResourceAsync` method calls `TenantService.GetTenants()` without passing the tenant parameter:

```csharp
var allTenants = await TenantService.GetTenants(cancellationToken);
var tenantResource = allTenants.FirstOrDefault(t => t.Data.TenantId == tenantId.Value);
```

And in `ExecuteResourceQueryAsync`, the explicitly passed `tenant` parameter is not forwarded:

```csharp
var subscriptionResource = await _subscriptionService.GetSubscription(subscription, tenant, retryPolicy, cancellationToken);
var tenantResource = await GetTenantResourceAsync(subscriptionResource!.Data.TenantId, cancellationToken);
// ^ 'tenant' parameter from caller is not used to scope the credential
```

The [`TenantService.GetTenants()`](https://github.com/microsoft/mcp/blob/main/core/Azure.Mcp.Core/src/Services/Azure/Tenant/TenantService.cs#L39) method enumerates all tenants via `ArmClient.GetTenants()` without using a credential scoped to the requested tenant. If the default credential cannot enumerate the target tenant, it fails.

### Reproduction Steps

1. Run `az login --tenant <tenant-id>`
2. Confirm login: `az account show` returns the correct tenant
3. Use the `role_assignment_list` tool with a valid scope, passing `tenant` and `subscription` explicitly
4. Observe: `InvalidOperationException: No accessible tenant found for tenant ID '<id>'`
5. Run `az role assignment list --scope <same-scope> --include-inherited` — works correctly

### Environment

- OS: Windows 11
- IDE: VS Code (latest)
- Azure MCP Server: latest
- Shell: Git Bash

KarishmaGhiya:
## Description

Fixes #2141.

Resource Graph queries resolved subscriptions with the requested tenant, but then enumerated all accessible tenants with an unscoped credential to obtain a `TenantResource`. In cross-tenant scenarios that enumeration could not see the requested tenant and failed with `No accessible tenant found`.

### Changes

- Send Resource Graph requests through the configured Azure HTTP pipeline using an ARM token scoped to the requested tenant.
- Preserve Azure SDK retry and user-agent behavior while supporting sovereign cloud ARM endpoints.
- Keep tenant enumeration only as the fallback for management-group queries where no tenant was supplied.
- Add transport-level regression coverage proving subscription-scoped role assignment queries do not enumerate tenants.
- Add an Azure MCP changelog entry.

## Testing

- `dotnet test --project tools\Azure.Mcp.Tools.Authorization\tests\Azure.Mcp.Tools.Authorization.Tests\Azure.Mcp.Tools.Authorization.Tests.csproj` — 17 passed.
- `dotnet build tools\Azure.Mcp.Tools.Authorization\src\Azure.Mcp.Tools.Authorization.csproj --no-restore` — succeeded with 0 warnings and 0 errors.
- `dotnet format ... --verify-no-changes` — clean.
- `eng\common\spelling\Invoke-Cspell.ps1` — clean.

## Invoking Livetests


core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -91,7 +69,10 @@ protected async Task<ResourceQueryResults<T>> ExecuteResourceQueryAsync<T>(
         ValidateAdditionalFilter(additionalFilter);
 
         var subscriptionResource = await AzureService.GetSubscription(subscription, tenant, cancellationToken);
-        var tenantResource = await GetTenantResourceAsync(subscriptionResource!.Data.TenantId, cancellationToken);
+        var tenantId = await AzureService.ResolveTenantIdAsync(tenant, cancellationToken)
```
JasonYeMSFT: If in most cases the subscription object contains the tenantId, we should make that the first option. I think it would be more efficient.

msalaman (approved):

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -192,42 +169,75 @@ private static string BuildResourceQuery(
         return queryFilter + $" | limit {limit}";
     }
 
-    private static List<T> MaterializeResults<T>(ResourceQueryResult? result, Func<JsonElement, T> converter)
+    private async Task<ResourceQueryResults<T>> ExecuteResourceGraphQueryAsync<T>(
+        ResourceQueryContent queryContent,
+        string tenant,
```
JasonYeMSFT: Nit: name this parameter `tenantId` for clarity

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
KarishmaGhiya: Updated in 86e749bf2. The subscription tenant ID is now preferred, with option resolution used only as a fallback. The regression test also verifies tenant resolution occurs only once during token acquisition.

core/Azure.Mcp.Core/src/Services/Azure/BaseAzureResourceService.cs
KarishmaGhiya: Updated in 86e749bf2. Renamed the internal parameter to `tenantId` and updated its use.

msalaman (approved):

PR #3731, opened 2026-09-18
Add endpoint validation for Confidential Ledger, Azure Terraform, and Azure Migrate tools

g2vinay:
## What does this PR do?

Adds outbound endpoint/URL validation to three toolsets so that constructed or data-derived request targets are checked against the expected hosts before a request is made. No tool names, command signatures, or public behavior change — these are internal validation additions.

- **Confidential Ledger**: The data-plane ledger endpoint is built from the caller-supplied ledger name and the target cloud. `ConfidentialLedgerService` now validates the constructed endpoint against the expected `*.confidential-ledger.azure.{com,cn,us}` domain for the active cloud (new `confidential-ledger` allow-list entry). `GetLedgerUri` was refactored into a testable `CreateValidatedLedgerEndpoint`.
- **Azure Terraform**: The AVM / AzureRM / AzAPI documentation services fetch content from GitHub. Outbound URLs are now validated against the expected hosts (`raw.githubusercontent.com`, `api.github.com`), including the releases URL that is derived from the fetched module-index CSV.
- **Azure Migrate**: The platform landing zone download URL is parsed from the generation response and fetched without authentication. `PlatformLandingZoneService` now validates that URL before the download.

Each change includes unit tests. Control-plane/ARM SDK calls and fixed-host requests were reviewed and left unchanged.

## GitHub issue number?

N/A

## Pre-merge Checklist
- [x] Required for All PRs
    - [x] Commit history is clean with descriptive messages (one commit per toolset, each including code, tests, and changelog entry)
    - [x] Created a changelog entry (one per toolset under `Other Changes`)
- [ ] For MCP tool changes:
    - [ ] N/A — no MCP tools are added, renamed, or modified; command signatures and descriptions are unchanged. Changes are internal validation only, so `consolidated-tools.json`, `azmcp-commands.md`, `e2eTestPrompts.md`, and `ToolDescriptionEvaluator` updates do not apply.

## Notes for reviewers

- Although this spans three toolsets, it does not add or modify any MCP tool, so the "one tool per PR" guidance is not applicable. Happy to split into three PRs if preferred.
- Unit test results: Confidential Ledger 35 passed / 3 skipped (live-only); Azure Terraform 168 passed; Azure Migrate 47 passed.

## Invoking Livetests


tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
```diff
@@ -129,7 +130,8 @@ public async Task<string> DownloadAsync(PlatformLandingZoneContext context, stri
         if (string.IsNullOrEmpty(downloadUrl))
             throw new InvalidOperationException("Download URL not yet available. The landing zone may still be generating. Please try again in 1-2 minutes.");
 
-        var bytes = await httpHelper.DownloadBytesAsync(downloadUrl, cancellationToken);
+        var validatedDownloadUrl = ValidateDownloadUrl(downloadUrl, logger);
+        var bytes = await httpHelper.DownloadBytesAsync(validatedDownloadUrl, cancellationToken);
```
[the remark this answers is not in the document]
g2vinay: Correct that this checks the initial URL only and the download client follows redirects. Keeping it as an allow-list/public-target check on purpose: the downloadUrl comes from a trusted ARM response, and we can't disable redirects globally because other callers (the guidance service) intentionally follow aka.ms redirects. I reworded the comment so it no longer claims full SSRF protection. Open to a scoped no-redirect handler later if we want to fully close the redirect path.

core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.AllowLists.cs
```diff
@@ -124,6 +124,12 @@ public string[] GetSuffixes(ArmEnvironment environment) =>
             UsGov: [".communication.azure.us"],
             Germany: [".communication.azure.de"],
             UseLegacyCheck: true), // INITIAL SEEDED UseLegacyCheck. NEEDS VERIFICATION.
+        ["confidential-ledger"] = new AllowedSuffixManager(
+            Public: [".confidential-ledger.azure.com"],
+            China: [".confidential-ledger.azure.cn"],
+            UsGov: [".confidential-ledger.azure.us"],
+            Germany: [".confidential-ledger.azure.com"], // Confidential Ledger is not offered in the Germany cloud; fall back to the public suffix, matching the service's default host.
```
alzimmermsft: Would it be better to fail out on this rather than an implicit success?

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AvmDocsService.cs
```diff
@@ -78,11 +86,13 @@ public async Task<string> GetDocumentationAsync(string moduleName, string module
         using var client = httpClientFactory.CreateClient();
         ConfigureGitHubHeaders(client);
 
+        EndpointValidator.ValidateExternalUrl(readmeUrl, s_allowedHosts);
```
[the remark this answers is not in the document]
g2vinay: Correct that this validates the initial URL only and the factory client follows redirects. We're keeping it as a host allow-list rather than adding redirect-aware handling here: these fetches target raw.githubusercontent.com and api.github.com, which don't return 30x redirects to internal hosts, and the module URLs come from the Microsoft-owned Azure-Verified-Modules CSV. A blanket AllowAutoRedirect=false isn't safe either, since other callers (Azure Migrate's guidance service) rely on aka.ms redirects. I reworded the comments to describe this as host allow-list validation rather than full SSRF protection.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AzApiExamplesService.cs
```diff
@@ -73,6 +77,7 @@ public async Task<List<AzApiExample>> GetExamplesAsync(
 
         try
         {
+            EndpointValidator.ValidateExternalUrl(remarksUrl, s_allowedHosts);
```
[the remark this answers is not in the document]
g2vinay: Same as the AvmDocsService thread: the check is on the initial URL and the factory client follows redirects. Keeping it as a host allow-list here. The fetches target raw.githubusercontent.com, which doesn't redirect to internal hosts, and disabling redirects globally would break other callers (Azure Migrate's guidance service relies on aka.ms redirects). Wording is being treated as allow-list validation rather than full SSRF protection.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AzureRMDocsService.cs
```diff
@@ -36,6 +40,7 @@ public async Task<AzureRMDocsResult> GetDocumentationAsync(
         string? markdownContent = null;
         bool isDataSourceUrl = isDataSource;
 
+        EndpointValidator.ValidateExternalUrl(docUrl, s_allowedHosts);
```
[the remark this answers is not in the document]
g2vinay: Same as the other AzureTerraform threads: the check is on the initial URL and the factory client follows redirects. Keeping it as a host allow-list. The fetches target raw.githubusercontent.com, which doesn't redirect to internal hosts, and a global AllowAutoRedirect=false would break other callers (Azure Migrate's guidance service relies on aka.ms redirects). Treating this as allow-list validation rather than full SSRF protection.

core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.AllowLists.cs
g2vinay: Good call. Changed the Germany suffix to an empty list so validation fails (SecurityException) instead of implicitly accepting the public `.com` host, since Confidential Ledger isn't offered in the Germany cloud. Added a test asserting the Germany cloud throws. Note this branch is currently unreachable anyway (AzureCloudConfiguration.AzureCloud has no Germany value and ParseCloudValue throws on anything but public/china/usgov), so this is defensive correctness rather than a live path.

tools/Azure.Mcp.Tools.ConfidentialLedger/src/Services/ConfidentialLedgerService.cs
```diff
@@ -126,6 +132,21 @@ private string GetLedgerUri(string ledgerName)
         };
     }
 
+    /// <summary>
+    /// Validates a Confidential Ledger data-plane endpoint against the endpoint allow-list for the target
+    /// cloud, guarding against endpoint injection before the endpoint is used to create a client.
+    /// </summary>
+    internal static Uri ValidateLedgerEndpoint(Uri endpoint, ArmEnvironment armEnvironment)
```
vukelich, line 139: This method is only `internal` for unit testing. It's not needed for the product, and the separation allows for code paths that use `ValidateLedgerEndpoint` but not `GetLedgerUri`. This is an artifact of updating code using tooling, and I spotted it in the previous `EndpointValidator` change. The separation of "make" and "validate" is a pitfall for both human and AI, but realistically, the latter is touching the code more.

I'd recommend either of the following:

1. Add a clear comment for AI to never use this method outside of testing. This helps more with AI touches because I expect AI to at least read method doc comments. I could be wrong, though. It's what I opted for in my rush [earlier this month](https://github.com/microsoft/mcp/blob/f8f9263d8895332727751362dd3f3b48e3f072ea/core/Microsoft.Mcp.Core/src/Helpers/EndpointValidator.cs#L85-L88)
1. Update `GetLedgerUri` to itself be also doing the call to `EndpointValidator.ValidateAzureServiceEndpoint`. This is the safer approach because it functional joins creation and validation. It might mean turning `GetLedgerUri` into `internal` as an instanced method, but not much can be done about that :/

vukelich (commented):
Early comment

tools/Azure.Mcp.Tools.ConfidentialLedger/src/Services/ConfidentialLedgerService.cs
g2vinay: Good catch, went with option 2. Merged creation and validation into a single `internal Uri GetValidatedLedgerUri(string ledgerName)` instance method that builds the host from CloudType and validates before returning, so there's no longer a separate "make" method a caller could use without validating. Removed the standalone `ValidateLedgerEndpoint`. The method is `internal` only for testing (with a doc comment saying so), and the test now constructs the service with a mocked `IAzureCloudConfiguration` (same pattern as the MySql tests). Build clean, 32 passed / 3 skipped.

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
```diff
@@ -165,6 +166,24 @@ Cached parameters (updated {p.CachedAt:u}):
     private string BuildUrl(PlatformLandingZoneContext ctx, string action) =>
         $"{AzureService.CloudConfiguration.ArmEnvironment.Endpoint}subscriptions/{ctx.SubscriptionId}/resourceGroups/{ctx.ResourceGroupName}/providers/Microsoft.Migrate/MigrateProjects/{ctx.MigrateProjectName}/{action}?api-version={PlatformLandingZoneConstants.ApiVersion}";
```
vukelich, line 167: We should be escaping the values before just throwing them into a formatted string.

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
```diff
@@ -165,6 +166,24 @@ Cached parameters (updated {p.CachedAt:u}):
...
 
+    /// <summary>
+    /// Parses the download URL from a landing-zone generation response and validates it as a public,
+    /// non-reserved target before returning. Parsing and validation are intentionally joined so that no
+    /// caller can obtain an unvalidated download URL. Returns <see langword="null"/> when the response does
+    /// not yet contain a download URL. This method is <see langword="internal"/> only to enable unit testing.
+    /// </summary>
+    internal static string? TryGetValidatedDownloadUrl(string response, ILogger? logger)
```
vukelich: Similar feedback I brought up with Alan's PR: having a new `internal static` for URI validation product code exposed this way just for unit testing means a pit-of-failure possibility for regressions. This method could go silently unused since it's not core to the logic that builds and uses the validated URI.

My recommendation is to combine this logic into `BuildUrl`. If you want to unit test for the validation logic, then you could open `BuildUrl` up to be an `internal static` that takes some of the instance arguments (e.g., `AzureService.CloudConfiguration.ArmEnvironment.Endpoint`). I'm fine with that route here because `BuildUrl` looks to be more likely to be called by human or AI authors in this file rather than recreating the logic themselves. My preference would be on endpoint validation testing the `public` methods, but it's your call.

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
vukelich: [`Uri.EscapeDataString`](https://learn.microsoft.com/en-us/dotnet/api/system.uri.escapedatastring?view=net-10.0)

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
vukelich, line 175: Please follow full doc comment syntax, such as using `<returns>` and `<param>`.

vukelich: Also, please add a `<remarks>` doc comment that this is only `internal static` for unit testing. Use within the class is good; use by anything else is bad.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AvmDocsService.cs
vukelich, line 85: Add `Uri.EscapeDataString` here.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AvmDocsService.cs
```diff
@@ -78,11 +87,13 @@ public async Task<string> GetDocumentationAsync(string moduleName, string module
...
         var response = await client.GetAsync(new Uri(readmeUrl), cancellationToken).ConfigureAwait(false);
 
         if (!response.IsSuccessStatusCode)
         {
             readmeUrl = $"https://raw.githubusercontent.com/{repoPath}/{cleanVersion}/README.md";
```
vukelich, line 95: Add `Uri.EscapeDataString` here.

tools/Azure.Mcp.Tools.ConfidentialLedger/src/Services/ConfidentialLedgerService.cs
```diff
@@ -109,11 +110,17 @@ public async Task<LedgerEntryGetResult> GetLedgerEntryAsync(string ledgerName, s
...
     }
 
-    private string GetLedgerUri(string ledgerName)
+    /// <summary>
+    /// Builds the Confidential Ledger data-plane endpoint for the caller-supplied ledger name and the target
+    /// cloud, and validates it against the endpoint allow-list before returning. Creation and validation are
+    /// intentionally joined in a single method so that no caller can obtain an unvalidated endpoint. This
+    /// method is <see langword="internal"/> only to enable unit testing; do not expose or bypass it.
+    /// </summary>
+    internal Uri GetValidatedLedgerUri(string ledgerName)
```
vukelich, line 119: Please add a `<remarks>` doc comment that this is only `internal static` for unit testing. Use within the class is good; use by anything else is bad.

vukelich (changes requested):

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
g2vinay: Done. Wrapped the subscription, resource group, and project name in `Uri.EscapeDataString` before building the URL.

tools/Azure.Mcp.Tools.AzureMigrate/src/Services/PlatformLandingZoneService.cs
g2vinay: Done. Added `<param>` and `<returns>` tags, plus a `<remarks>` note saying it is `internal static` for unit testing only and should not be called from anything else.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AvmDocsService.cs
g2vinay: Done. The version now goes through `Uri.EscapeDataString` before it is put in the URL. The repo path is left as-is because it is an `owner/repo` value with a slash, and escaping the whole thing would break the URL.

tools/Azure.Mcp.Tools.AzureTerraform/src/Services/AvmDocsService.cs
g2vinay: Done. Same fix as the other README URL: the version is escaped where it is set, so both URLs use the escaped value.

tools/Azure.Mcp.Tools.ConfidentialLedger/src/Services/ConfidentialLedgerService.cs
g2vinay: Done. Added a `<remarks>` note that this method is `internal` only for unit testing, along with `<param>` and `<returns>` tags.

alzimmermsft (approved):

